PR #1, opened 2025-01-30
feat: prototype of a Solidity-to-EVM compiler statically linked with solc

hedgar2017:
Pushes the initial codebase derived from https://github.com/matter-labs/era-compiler-solidity

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -49,7 +49,7 @@ jobs:
           - name: "MacOS arm64"
             runner: [self-hosted, macOS, ARM64]
           - name: "Linux x86 gnu"
-            runner: matterlabs-ci-runner-high-performance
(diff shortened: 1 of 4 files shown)


solx-solc/build.rs
```diff
@@ -0,0 +1,36 @@
+//!
+//! The default build script for `solx`.
+//!
+
+///
+/// Links `libsolc` statically.
+///
+fn main() {
+    // Tell Cargo where to find static libraries.
+    println!("cargo:rustc-link-search=native=/Users/hedgarmac/src/era-solidity/build/libsolc/");
```
popzxc: Nit: probably we will need a machine-independent solution 😅

hedgar2017: Haha, but not today sir 🫡  !

PR #10, opened 2025-02-25
chore: set the license to GPL-3.0

hedgar2017:
Sets the license to GPL-3.0.

@popzxc I've copied it from the Solidity repo as you suggested. Please verify!

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ resolver = "2"
 
 [workspace.package]
 authors = [
-    "Oleksandr Zarudnyi <a.zarudnyy@matterlabs.dev>",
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -10,6 +10,6 @@ resolver = "2"
 authors = [
     "Oleksandr Zarudnyi <a.zarudnyy@matterlabs.dev>",
 ]
```
popzxc: Please change the authors to `"The Matter Labs Team <hello@matterlabs.dev>"` similarly to how we do it [in other crates](https://github.com/matter-labs/zksync-era/blob/main/core/Cargo.toml)

PR #35, opened 2025-04-09
fix: reduce Yul object identifier to their last segment

hedgar2017:
# What ❔

Removes all segments but the last one in Yul object identifiers such as `D_25.D_25_deployed`, so only `D_25_deployed` is used from now on.

## Why ❔

It creates a discrepancy between Yul object names themselves and references used in `dataoffset` and `datasize`, so the names do not match during linkage.

## Checklist


hedgar2017:
Hey @antonbaliasnikov - here we'll need the `--checkout` flag that I've described in the ticket.

antonbaliasnikov (approved):

PR #64, opened 2025-05-20
fix: Yul object naming assumptions and keywords

hedgar2017:
1. Removes the requirement for Yul objects to be naming strictly as pairs with a `_deployed` suffix (e.g. `Contract`/`Contract_deployed`). Now runtime code is detected solely by its position in solc Yul output.
2. Removes `object` and `code` from reserved keywords similarly to how `data` has been handled. Now all 3 are treated as identifiers according to the Yul spec.

Resolves https://github.com/matter-labs/solx/issues/61

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1552,6 +1552,7 @@ name = "solx-yul"
 version = "0.1.0-alpha.2"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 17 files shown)


PavelKopyl (approved):
LGTM

PR #69, opened 2025-05-31
feat: add remaining output data CLI parameters

hedgar2017:
Resolves https://github.com/matter-labs/solx/issues/56.

Additionally, improves compilation times by optimizing out copying of large data chunks such as Yul and EVM assembly.

## Checklist


docs/src/02-command-line-interface.md
```diff
@@ -77,85 +95,274 @@ main:
...
+**solx** supports multiple input files. The following command compiles two Solidity files and prints the bytecode:
+
+```bash
+solx 'Simple.sol' 'Complex.sol' --bin
+```
+
+[Solidity import remappings](https://docs.soliditylang.org/en/latest/path-resolution.html#import-remapping) are passed in the way as input files, but they are distinguished by a `=` symbol between source and destination. The following command compiles a Solidity file with a remapping and prints the bytecode:
+
+```bash
+solx 'Simple.sol' 'github.com/ethereum/dapp-bin/=/usr/local/lib/dapp-bin/' --bin
```
antonbaliasnikov, line 336: As `=` is the allowed file name symbol, there is a possibility that some file names will be recognized as import remappings?

docs/src/02-command-line-interface.md
```diff
@@ -249,7 +456,7 @@ Sets the optimization level of the LLVM optimizer. Available values are:
 
 | Level | Meaning                      | Hints                                            |
 |:------|:-----------------------------|:-------------------------------------------------|
-| 0     | No optimization              | Currently not supported
+| 0     | No optimization              | For lightning-last compilation during development
```
antonbaliasnikov:
```suggestion
| 0     | No optimization              | For lightning-fast compilation during development
```

Typo here, but maybe we can just use `For faster compilation during development` ?

docs/src/02-command-line-interface.md
hedgar2017: `=` is not allowed in file names. IIRC, it is true for all platforms we support.

docs/src/02-command-line-interface.md
hedgar2017: Hehe, sure!

docs/src/02-command-line-interface.md
hedgar2017: It is not my invention though. It just mimics the Solidity counterpart feature.

Btw it is shown as added in the current patch just because I moved things around. It was implemented long ago.

PR #77, opened 2025-06-07
feat: make stack never too deep

hedgar2017:
Delivers the comprehensive solution to the stack-too-deep problem.

## Checklist


solx/tests/data/contracts/solidity/StackTooDeep.sol
2xic, line 1: Can you add a similar test, but for emitting events ? (event with 16 + fields?)

2xic (approved):
Github is so nice that it allows me to review this so I just wanted to give it my approval as it fixed via-ir on some more complex that I had

solx/tests/data/contracts/solidity/StackTooDeep.sol
2xic: Also normal solc is also able to compile both of these test with via-ir.

Shouldn't there be a even more complex cases for this given "never too deep again" ?

hedgar2017: Hey @2xic, thank you for the review!
Feel free to send the tests here or just commit it into the branch, I'll make sure they're run properly. We'll also greatly appreciate adding relevant stack-too-deep cases to our [test collection](https://github.com/matter-labs/era-compiler-tests/tree/main/solidity).
Complex cases are covered by our [integration test suite](https://github.com/matter-labs/era-compiler-tester/pull/167), whereas this test is only for CLI and second-pass functionality.
Actually it is added specifically for the old solc codegen that is still used by default under the hood. With this patch, we want to elimiate all possibilities of stack-too-deep in:

1. Old solc codegen (EVM assembly)
2. New solc codegen (Yul)
3. Our stackification algorithm in LLVM

All three areas are already fixed, and now we're merely cleaning things up across the compilation pipeline. I'll probably add a couple of tests for various cases of stack-too-deep soon.

PR #85, opened 2025-06-25
ci: add compilation time table

antonbaliasnikov:


antonbaliasnikov:
Pushed changes directly to the [az-stack-never-too-deep](https://github.com/matter-labs/solx/tree/az-stack-never-too-deep) branch.

PR #116, opened 2025-07-23
feat: add an LLVM attribute for EVM entry functions

hedgar2017:
# What ❔

Adds an LLVM string attribute `evm-entry-function` for EVM entry functions.

## Why ❔

It is needed for LLVM to always put the entry function at the beginning of the bytecode without fragile explicit checks for function names.

## Checklist


PavelKopyl (approved):
LGTM

PR #144, opened 2025-09-29
fix: improve the handling of empty input

hedgar2017:
Fixes https://github.com/matter-labs/solx/issues/143

Additional fixes:

1. Input paths (and `-`) are validated before any output selection checks.
2. The hyphen ( `-`) can now be used with `--standard-json` as an stdin identifier and not treated as an invalid path.
3. The `--help` message does not start with `Error: ` anymore (it's a `clap` library issue).

## Checklist


akiramenai (approved):
LGTM.

I started discussing it with @antonbaliasnikov. It would be great to add `llvm-lit` or `compiletest` (rust version of `lit`) into solx infra to make tests on diagnostic less verbose. And to unify the approach we write tests in different components of the compiler.

solx/src/solx/arguments.rs
```diff
@@ -203,10 +203,34 @@ impl Arguments {
     pub fn validate(&self) -> Arc<Mutex<Vec<solx_standard_json::OutputError>>> {
         let mut messages = vec![];
 
-        if self.version && std::env::args().count() > 2 {
+        if self.recursive_process {
+            if std::env::args().count() > 3 {
+                messages.push(solx_standard_json::OutputError::new_error(
```
akiramenai, line 208: It looks like `solx_standard_json::OutputError::new_error` is frequently called with `None` as the 1st, 3rd, and 4th parameter. Making a wrapper for it would reduce duplication.

hedgar2017: Makes sense, let me address it.

PR #149, opened 2025-10-02
feat: emit div as intrinsic

vladimirradosavljevic:
# What ❔
Instead of emitting CFG that handles zero case, emit div intrinsic.

## Why ❔
To check the performance numbers.

## Checklist

- [ ] PR title corresponds to the body of PR.
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] Code has been formatted via `cargo fmt` and checked with `cargo clippy`.


hedgar2017 (changes requested):
Thanks for dropping this here! A couple of questions:

1. Does it require an LLVM counterpart? A few sentences in the description would be nice.
2. Do we have to do something similar for `rem`?

vladimirradosavljevic:
> Thanks for dropping this here! A couple of questions:
> 
> 1. Does it require an LLVM counterpart? A few sentences in the description would be nice.
> 2. Do we have to do something similar for `rem`?

This was just an experiment for now, and most likely we will postpone this work.

PR #163, opened 2025-10-30
fix: clone the LLVM IR module before emitting assembly

hedgar2017:
# What ❔

Clone the module before calling `LLVMWriteToMemoryBuffer` for `Assembly`.

## Why ❔

`LLVMWriteToMemoryBuffer` preserves the module's state, so the optimizer is called twice: once before getting the assembly, and then before the codegen.

## Checklist


PavelKopyl (approved):
LGTM

PR #171, opened 2025-11-10
ci: replace ML runners with Nomic runners

alcuadrado:
This PR replaces both the `matterlabs-ci-runner-high-performance` and `matterlabs-ci-runner-highdisk`, with the `self-hosted` one.

I used the same, as it 2 X 8 TB NVMe + 2 X 960 GB NVMe in Raid 0. So I didn't see the need for a highdisk one.

I also found that there's a `matterlabs-ci-runner-arm`. What are the requirements for that one?

There are still some references to the matterlabs repo in the workflows, which probably need to be updated.

### DIFF
--- a/solx-benchmark-converter/src/output/xlsx/mod.rs
+++ b/solx-benchmark-converter/src/output/xlsx/mod.rs
@@ -212,8 +212,10 @@ impl TryFrom<(Benchmark, Source)> for Xlsx {
             .set_totals(xlsx.toolchain_ids.len())?;
 
         let comparison_mapping = match source {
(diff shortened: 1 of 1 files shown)


alcuadrado:
@antonbaliasnikov and what about these?

  - `windows-2022-github-hosted-16core` 
  - `windows-2022-github-hosted-64core`
  - `macos-latest-large`
  - `macos-latest-xlarge`


Are they they https://docs.github.com/en/actions/concepts/runners/larger-runners?

To add a bit of context, my goal right now is to get the CI running again asap. Then we should analyze the workflows to see if there are more ML->Nomic transitions to be done. e.g. maybe `ghcr.io/matter-labs/zksync-llvm-runner:latest`

PR #239, opened 2026-02-24
feat: add MLIR passthrough mode

hedgar2017:
- When both `solc` and `mlir` features are active, solc's internal MLIR pipeline produces final EVM bytecode — solx becomes a thin wrapper
- Decouples `mlir` feature into a marker (`mlir = []`) to avoid inkwell feature poisoning; actual MLIR codegen gated behind `solx-mlir`
- Adds `Build::from_solc_output()` to extract bytecode/abi/metadata from solc standard JSON output
- Splits `standard_output_evm()` and `standard_json_evm()` with cfg guards for passthrough vs normal paths
- Guards CLI flags that would panic in passthrough mode
- Wires CLI `-O` flag to solc's `optimizer.mode` via `middle_end_as_char()`
- Strips `optimizer.size_fallback` (solc doesn't recognize it); keeps `optimizer.mode` (solc-mlir uses it)
- Adds missing MLIR link libraries, fixes cmake path resolution, adds `build-mlir`/`test-mlir` cargo aliases
- Updates `solx-solidity` and `solx-llvm` submodules for MLIR support

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -36,3 +36,5 @@ CARGO_BIN_EXE_SOLX = { value = "./target/debug/solx", relative = true, force = f
 
 [alias]
 test-slang = "test -p solx-slang -p solx-mlir -p solx --no-default-features --features slang"
+build-mlir = "build --features mlir"
+test-mlir = "test -p solx --features mlir"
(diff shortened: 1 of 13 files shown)


hedgar2017:
Superseded by LIT tests and other test infrastructure.

PR #250, opened 2026-03-09
ci: add Socket Firewall (SFW) composite action and test workflow

nebasuke:
Add setup-sfw composite action that installs SFW with platform detection (skips Windows and container environments) and a label-gated (ci:sfw) test workflow to validate SFW across runners, containers, and Windows.


.github/workflows/test.yaml
```diff
@@ -63,6 +63,13 @@ jobs:
           git config --global --add safe.directory '*'
           git submodule update --force --depth=1 --recursive --checkout
 
+      - name: Set SFW prefix
+        shell: bash
+        run: |
+          if command -v sfw &>/dev/null; then
+            echo "SFW_PREFIX=sfw" >> "$GITHUB_ENV"
+          fi
```
hedgar2017: **[Bug] `cargo-checks` container job: "Set SFW prefix" step is a no-op**

The `cargo-checks` job runs in a container (`solx-ci-runner` image) which does not have `sfw` pre-installed. The "Set SFW prefix" step checks `command -v sfw` which will fail, so `SFW_PREFIX` is never set. All the `${SFW_PREFIX:-} cargo …` calls in `cargo-check/action.yml` then silently fall back to plain `cargo`.

Compare with the `build-and-test` job (line ~146) which correctly adds:
```yaml
- name: Install SFW (container)
  if: matrix.image != ''
  run: npm install -g sfw
```

The `cargo-checks` job needs an equivalent `npm install -g sfw` step inserted before "Set SFW prefix". Otherwise remove the prefix step from this job to make the intent explicit.

.github/actions/setup-sfw/action.yml
```diff
@@ -0,0 +1,38 @@
...
+          echo "::notice::SFW skipped: running inside a container (install via npm instead)"
+          echo "available=false" >> "$GITHUB_OUTPUT"
+          exit 0
+        fi
+
+        echo "available=true" >> "$GITHUB_OUTPUT"
+
+    - name: Install SFW
+      if: steps.detect.outputs.available == 'true'
+      uses: socketdev/action@4337a545deecc20f19a909e52db7a2f6ba292f42 # v1
```
hedgar2017: **Security (SHA pinning):** The SHA pin is the correct approach for third-party actions. Verify that `4337a545deecc20f19a909e52db7a2f6ba292f42` corresponds to the actual `v1` tag on the `socketdev/action` repo before merging. Consider adding a verification date in the comment (e.g., `# v1 — verified 2026-03-09`).

.github/actions/setup-sfw/action.yml
```diff
@@ -0,0 +1,38 @@
...
+      run: |
+        # Skip inside containers (GitHub sets 'container' env var for container jobs;
+        # /.dockerenv is a fallback heuristic)
+        if [[ -n "${container:-}" ]] || [[ -f /.dockerenv ]]; then
+          echo "::notice::SFW skipped: running inside a container (install via npm instead)"
+          echo "available=false" >> "$GITHUB_OUTPUT"
+          exit 0
+        fi
+
+        echo "available=true" >> "$GITHUB_OUTPUT"
```
hedgar2017: **Windows:** This action does **not** skip Windows (only containers). That's fine — socketdev/action presumably handles Windows. But worth noting in the description since the PR summary says "skips Windows" which could be misleading.

.github/actions/rust-unit-tests/action.yml
```diff
@@ -74,8 +74,8 @@ runs:
         if [[ ${RUNNER_OS} = Windows ]] || [[ ${RUNNER_OS} = Linux ]]; then
           export RUSTFLAGS="${RUSTFLAGS} -C link-arg=-fuse-ld=lld"
         fi
-        if [ "${TEST_COMMAND}" = "test" ]; then cargo build --frozen; fi
-        cargo ${TEST_COMMAND} --frozen --release ${{ steps.build-target.outputs.target }} -- -Z unstable-options \
+        if [ "${TEST_COMMAND}" = "test" ]; then ${SFW_PREFIX:-} cargo build --frozen; fi
+        ${SFW_PREFIX:-} cargo ${TEST_COMMAND} --frozen --release ${{ steps.build-target.outputs.target }} -- -Z unstable-options \
           --format json | tee -a release-results.json | grep 'failed'
         if [ $? -eq 0 ]; then
           cargo2junit < release-results.json > "release-${{ inputs.results-xml }}"
```
hedgar2017: **[Correctness] `cargo2junit` correctly left unprefixed**

`cargo2junit < release-results.json` is a local binary reading stdin — no network calls, no package resolution. Leaving it unprefixed is correct. Same applies to the identical usage on the debug path below. ✓

.github/workflows/sfw-test.yaml
```diff
@@ -0,0 +1,99 @@
...
+          sfw --version
+
+      - name: Test sfw npm install
+        if: steps.sfw.outputs.available == 'true'
+        run: sfw npm install -g semver
+
+      - name: Test sfw cargo fetch (expected SSL failure — tracking)
+        if: steps.sfw.outputs.available == 'true'
+        continue-on-error: true
+        run: sfw cargo fetch --locked
```
hedgar2017: **Known tech debt — needs a tracking issue.** The step name acknowledges an SSL failure, but there's no linked issue. If `sfw cargo fetch` doesn't work reliably, the entire `${SFW_PREFIX:-}` pattern across all composite actions is essentially a no-op today (SFW_PREFIX would never be set to `sfw` if it breaks cargo).

Please link a tracking issue here so this doesn't get lost. Also consider: if SFW can't wrap cargo due to SSL interception issues, is the current approach of prefixing all cargo commands premature? An alternative would be to only wrap `cargo fetch`/`cargo install` (network-facing) and defer the rest until the SSL issue is resolved.

.github/workflows/sfw-test.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Install SFW via npm
+        run: npm install -g sfw
+
+      - name: Verify SFW binary
+        run: |
+          command -v sfw
+          sfw --version
+
+      - name: Test sfw npm install
+        run: sfw npm install -g semver
```
hedgar2017: **Missing `sfw cargo fetch` test inside containers.** The container job tests `sfw npm install` but not `sfw cargo fetch`. Since the primary goal is wrapping cargo in CI, this gap means the container-based cargo wrapping (which is what `test.yaml` actually uses for Linux jobs) is untested.

Consider adding:
```yaml
      - name: Test sfw cargo fetch (container)
        continue-on-error: true
        run: sfw cargo fetch --locked
```

.github/actions/cargo-check/action.yml
```diff
@@ -17,23 +17,23 @@ runs:
...
 
     - name: Cargo check
       shell: ${{ runner.os == 'Windows' && 'msys2 {0}' || 'bash' }}
-      run: cargo check --frozen --verbose
+      run: ${SFW_PREFIX:-} cargo check --frozen --verbose
 
     - name: Cargo format
       shell: ${{ runner.os == 'Windows' && 'msys2 {0}' || 'bash' }}
-      run: cargo fmt --all -- --check --verbose
+      run: ${SFW_PREFIX:-} cargo fmt --all -- --check --verbose
```
hedgar2017: **Nit:** `cargo fmt` and `cargo clippy --frozen` don't perform network operations — they work on local sources only. Prefixing them with SFW is harmless but unnecessary. Not a blocker, just noting for clarity on what SFW is actually monitoring.

.github/workflows/integration-tests.yaml
hedgar2017, line 1: **Coverage gap: 7 unprefixed cargo commands in this workflow.**

Lines 92, 100, 108, 123, 129, 135, 171 all have bare `cargo fetch`/`cargo build` without `${SFW_PREFIX:-}`. This workflow runs in a container and has no SFW setup, so these won't be monitored.

If the intent is to eventually wrap all cargo commands with SFW, these should be prefixed now (like the composite actions) so they'll automatically pick up SFW when it's added to this workflow. If integration-tests is intentionally excluded from SFW, that's fine but worth a comment explaining why.

.github/workflows/coverage.yaml
hedgar2017, line 1: **Unprefixed cargo commands.** These two direct cargo invocations (lines 121, 134) are not wrapped with `${SFW_PREFIX:-}`, and this workflow has no SFW setup. Same situation as `integration-tests.yaml` — consider prefixing for forward compatibility or adding a comment explaining the exclusion.

hedgar2017 (changes requested):
## Review: SFW CI Integration

Overall the approach is sound — SHA-pinning `socketdev/action`, the `${SFW_PREFIX:-}` pattern for graceful degradation, and the label-gated test workflow are all good design choices. Two bugs need fixing before this is ready.

---

### 🔴 Bugs (blocking)

**1. `setup-sfw` does not skip Windows (`.github/actions/setup-sfw/action.yml:26`)**
The PR description and action description both say Windows is skipped, but the detect step only checks for containers. On a Windows runner, `available=true` is set and `socketdev/action` is invoked unconditionally. If `socketdev/action` doesn't support Windows, the workflow fails instead of skipping gracefully. Either add an explicit Windows skip or correct the documentation to say Windows is intentionally included.

**2. `cargo-checks` container job: "Set SFW prefix" step is a no-op (`.github/workflows/test.yaml:66–71`)**
The `cargo-checks` job runs in the CI container which doesn't have `sfw` installed. `command -v sfw` always fails, so `SFW_PREFIX` is never set and all the prefixed cargo calls in `cargo-check/action.yml` fall back to plain `cargo`. The `build-and-test` matrix job correctly does `npm install -g sfw` before the prefix check; `cargo-checks` is missing the equivalent step.

---

### 🟡 Medium concerns

**3. `npm install -g sfw` is unpinned**
Both `sfw-test.yaml:90` and `test.yaml:148` install SFW without a version — ironic for a supply-chain security tool. Pin to a specific version (e.g. `sfw@1.2.3`).

**4. Known SFW+cargo SSL failure in production path**
The `continue-on-error: true` on the test workflow's `sfw cargo fetch` step documents a known SSL incompatibility. The production workflows (`test.yaml`) also prefix cargo commands with `${SFW_PREFIX:-}` and have no such safety net. Confirm the SSL issue is resolved before promoting this beyond the `ci:sfw` label gate, or link a tracking issue in the comment.

**5. `sanitizer` job has no SFW coverage**
The `sanitizer` job in `test.yaml` has neither `npm install -g sfw` nor "Set SFW prefix". This may be intentional (it's a specialized label-gated job), but if so it's worth a comment explaining the omission.

---

### 🟢 Things that look correct

- `socketdev/action` is pinned to a full 40-char SHA — correct supply chain hygiene. ✓
- `socketdev` / socket.dev is a legitimate and well-known supply chain security vendor. ✓
- `${SFW_PREFIX:-}` expansion is correct POSIX bash and works correctly in both `bash` and `msys2 {0}` shells on Windows (both inherit `$GITHUB_ENV` and `$GITHUB_PATH`). ✓
- `/.dockerenv` + `$container` heuristic is reliable for GitHub Actions Docker containers. ✓
- `cargo2junit` left unprefixed is correct — it's a local binary reading stdin, not a network call. ✓
- The label-check → dependent-jobs pattern for `ci:sfw` gating is correct. ✓
- Node.js 24 via nodesource is appropriate; the `curl | bash` risk exists only at image-build time and is mitigated by GPG package signing. ✓
- Image SHA updated consistently across all six affected workflow files. ✓

.github/docker/Dockerfile
```diff
@@ -111,6 +111,12 @@ ENV PATH="/root/.cargo/bin:/usr/lib/llvm-21/bin:${PATH}" \
 # ── Cargo tools ────────────────────────────────────────────────────────────
 RUN cargo install cargo2junit cargo-llvm-cov cargo-udeps --locked
 
+# ── Node.js (for SFW and npm-based CI tools) ─────────────────────────────
+ARG NODE_MAJOR=24
+RUN curl -fsSL https://deb.nodesource.com/setup_${NODE_MAJOR}.x | bash - \
```
hedgar2017: **Node.js 24 is appropriate** — current LTS as of March 2026. The `--no-install-recommends` and apt cache cleanup are good practice.

**Minor concern:** `curl | bash` from nodesource is itself a supply chain vector. For a CI image focused on supply chain security (the whole point of SFW), consider pinning to a specific nodesource setup script version or using the official Node.js Docker image as a multi-stage source. Not a blocker.

hedgar2017 (commented):
One more suggestion:

.github/workflows/coverage.yaml
nebasuke: Fixed!

.github/workflows/test.yaml
nebasuke: Fixed. Refactored the install + prefix pattern into the `setup-sfw` composite action, which now handles both container and non-container environments. All workflows use `uses: ./.github/actions/setup-sfw` — no more manual steps to forget.

.github/actions/setup-sfw/action.yml
nebasuke: Fixed. The detect step now explicitly skips non-Linux runners (macOS and Windows) since the MITM proxy is incompatible with their TLS stacks. Updated the action description to match.

.github/workflows/sfw-test.yaml
nebasuke: The `sfw-test.yaml` workflow has been removed. The SSL issue was specific to the MITM proxy on bare-metal runners — inside containers (where most Linux CI jobs run), `sfw` wraps cargo directly without a proxy, so there's no SSL issue. The `setup-sfw` action now handles both paths cleanly.

.github/workflows/sfw-test.yaml
nebasuke: `sfw-test.yaml` has been removed. SFW is now tested implicitly by the main CI workflows themselves — if `sfw cargo fetch` fails, the build fails. Separate test workflow was redundant.

.github/actions/cargo-check/action.yml
nebasuke: Acknowledged. Kept the prefix on all cargo commands for consistency — it's a no-op for non-network commands and avoids having to reason about which commands do/don't hit the network.

.github/workflows/integration-tests.yaml
nebasuke: Fixed. All cargo commands in `integration-tests.yaml` are now prefixed with `${SFW_PREFIX:-}`, and the workflow uses the `setup-sfw` action.

.github/workflows/coverage.yaml
nebasuke: Fixed. All cargo commands prefixed and workflow uses the `setup-sfw` action.

.github/docker/Dockerfile
nebasuke: Addressed. Switched from `curl | bash` to proper GPG-signed apt repo setup (download key, add signed source, `apt-get install`), following the same pattern as the LLVM apt repo in the Dockerfile. Also pinned `sfw` to `sfw@2.0.4` in all npm install calls.

.github/workflows/integration-tests.yaml
```diff
@@ -113,40 +116,67 @@ jobs:
...
       - name: Build solx-dev and solx-tester (main)
         working-directory: temp-solx-main
         continue-on-error: true
         env:
           BOOST_PREFIX: ${{ github.workspace }}/temp-solx-main/solx-solidity/boost/lib
           SOLC_PREFIX: ${{ github.workspace }}/temp-solx-main/solx-solidity/build
-        run: cargo build --frozen --release --bin solx-dev --bin solx-tester
+        run: ${SFW_PREFIX:-} cargo build --frozen --release --bin solx-dev --bin solx-tester
+
+      - name: Build LLVM (PR)
```
hedgar2017: **Duplicate PR build block.** Lines 154-179 (`Build LLVM (PR)`, `Build solc (PR)`, `Fetch dependencies (PR)`, `Build solx (PR)`, `Build solx-dev and solx-tester (PR)`) duplicate the work already done at lines 80-131. Looks like a merge artifact. These should be removed — the LLVM/solc builds are cached so they won't recompile, but the cargo steps and step overhead are unnecessary.

.github/workflows/sanitizer.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
     needs: [label-check, cooldown-check]
     runs-on: solx-linux-amd64
     container:
-      image: ghcr.io/nomicfoundation/solx-ci-runner@sha256:c0866d146261cd0a51dc7d9077444b8ac3dde12c53d2151137834e6be149dbc7
+      image: ghcr.io/nomicfoundation/solx-ci-runner@sha256:c2b03000a1074d2cc3e6cf25a1c4fdc6eb0a61d23e63bef59205050249fa1d6e
```
hedgar2017, line 42: This workflow uses `build-llvm`, `build-solc`, and `rust-unit-tests` composite actions that now have `${SFW_PREFIX:-}` cargo prefixes, but there's no `Setup SFW` step in this job. SFW monitoring is silently absent. If this is intentional (specialized label-gated job), a comment explaining the exclusion would help. Otherwise, add `uses: ./.github/actions/setup-sfw` after the submodule checkout.

.github/actions/setup-sfw/action.yml
```diff
@@ -0,0 +1,124 @@
...
+          echo "::notice::SFW skipped: MITM proxy not compatible with $RUNNER_OS TLS stack"
+          echo "method=none" >> "$GITHUB_OUTPUT"
+          exit 0
+        fi
+
+        # Inside containers, the socketdev/action MITM proxy doesn't work,
+        # but the sfw CLI wrapper can be installed directly via npm.
+        # (GitHub sets 'container' env var for container jobs;
+        # /.dockerenv is a fallback heuristic)
+        if [[ -n "${container:-}" ]] || [[ -f /.dockerenv ]]; then
```
hedgar2017: The comment says "GitHub sets 'container' env var for container jobs" but GitHub Actions exposes the container configuration through the `job.container` expression context, not as a shell environment variable. `${container:-}` will always be empty in bash. The detection works correctly via the `/.dockerenv` fallback, but the first condition is dead code and the comment is misleading.

Consider simplifying to just `[[ -f /.dockerenv ]]` or passing `${{ job.container.id }}` as an input to the action.

.github/actions/setup-sfw/action.yml
```diff
@@ -0,0 +1,124 @@
...
+
+        # Export for cargo, git, and rustup (SSL_CERT_FILE is needed by rustls-native-certs)
+        echo "CARGO_HTTP_CAINFO=$BUNDLE" >> "$GITHUB_ENV"
+        echo "GIT_SSL_CAINFO=$BUNDLE" >> "$GITHUB_ENV"
+        echo "SSL_CERT_FILE=$BUNDLE" >> "$GITHUB_ENV"
+        echo "CA bundle created at $BUNDLE ($(wc -l < "$BUNDLE") lines)"
+
+    - name: Set SFW prefix
+      id: set-prefix
+      if: steps.detect.outputs.method != 'none'
```
hedgar2017: When `method=none` (macOS/Windows), this step is skipped entirely, leaving `outputs.available` as an empty string rather than `"false"`. The output description on line 16 says `true/false`, so any consumer checking `== 'false'` would get wrong results.

Either add a separate step for the `none` branch that writes `available=false`, or document that callers must check `!= 'true'`.

.github/actions/setup-sfw/action.yml
```diff
@@ -0,0 +1,124 @@
...
+        fs.writeFileSync(process.env.RUNNER_TEMP + "/sfw-cert-path.txt", p);
+        SCRIPT
+        sfw node "${RUNNER_TEMP}/sfw-get-cert.js" > /dev/null 2>&1 || true
+        SFW_CA=""
+        [ -f "${RUNNER_TEMP}/sfw-cert-path.txt" ] && SFW_CA=$(cat "${RUNNER_TEMP}/sfw-cert-path.txt")
+        rm -f "${RUNNER_TEMP}/sfw-get-cert.js" "${RUNNER_TEMP}/sfw-cert-path.txt"
+
+        if [ -z "$SFW_CA" ] || [ ! -f "$SFW_CA" ]; then
+          echo "::warning::SFW CA cert not found — cargo may fail with SSL errors under sfw"
+          exit 0
```
hedgar2017, line 80: When the SFW CA cert isn't found, this warns and exits 0, but the `set-prefix` step at line 113 still runs and sets `SFW_PREFIX=sfw`. Subsequent `${SFW_PREFIX:-} cargo` commands will route through SFW's MITM proxy without proper CA trust, producing opaque SSL errors.

Consider either skipping the prefix setup on cert failure (e.g. set an output that the `set-prefix` step checks) or failing the step.

.github/actions/cargo-check/action.yml
```diff
@@ -36,6 +36,6 @@ runs:
     - name: Cargo udeps
       shell: ${{ runner.os == 'Windows' && 'msys2 {0}' || 'bash' }}
       run: |
-        cargo install cargo-udeps@0.1.60 --locked
+        ${SFW_PREFIX:-} cargo install cargo-udeps@0.1.60 --locked
         cargo +nightly udeps --frozen --all-targets
```
hedgar2017: `cargo install cargo-udeps` is prefixed (line 39) but `cargo +nightly udeps --frozen` on the next line is not. The `--frozen` flag means it doesn't hit the network, so this is functionally fine, but it's inconsistent with the approach in `rust-unit-tests` which prefixes all cargo commands including `--frozen` ones.

.github/workflows/slang-tests.yaml
```diff
@@ -49,11 +49,11 @@ jobs:
             runner: macos-15
           - name: "Linux x86 gnu"
             runner: ubuntu-24.04
-            image: ghcr.io/nomicfoundation/solx-ci-runner@sha256:c0866d146261cd0a51dc7d9077444b8ac3dde12c53d2151137834e6be149dbc7
+            image: ghcr.io/nomicfoundation/solx-ci-runner@sha256:c2b03000a1074d2cc3e6cf25a1c4fdc6eb0a61d23e63bef59205050249fa1d6e
```
hedgar2017, line 53: `slang-tests.yaml`, `cache-warmup.yaml`, and `release.yaml` all use composite actions that have `${SFW_PREFIX:-}` cargo prefixes but never call `setup-sfw`. The `${SFW_PREFIX:-}` graceful degradation means builds still work, but SFW monitoring is silently absent. `release.yaml` is especially notable since it produces published binaries — the highest-value target for supply chain attacks.

hedgar2017 (changes requested):

.github/workflows/integration-tests.yaml
nebasuke: Fixed. Removed the duplicate block (lines 154-179).

.github/workflows/sanitizer.yaml
nebasuke: Fixed. Added `setup-sfw` step to the sanitizer job.

.github/actions/setup-sfw/action.yml
nebasuke: Fixed. Updated comment: `/.dockerenv` is now documented as the primary heuristic, `${container:-}` as a defensive fallback (not set by GitHub Actions). Check order flipped to match.

.github/actions/setup-sfw/action.yml
nebasuke: Fixed. The `set-prefix` step now always runs and explicitly outputs `available=false` when `method=none`.

.github/actions/setup-sfw/action.yml
nebasuke: Fixed. The CA cert step now outputs `ca-failed=true` on failure, and `set-prefix` checks this — if CA cert is missing, SFW prefix is not set to prevent opaque SSL errors.

.github/actions/cargo-check/action.yml
nebasuke: Fixed. Added `${SFW_PREFIX:-}` prefix to `cargo +nightly udeps`.

.github/workflows/slang-tests.yaml
nebasuke: Fixed. Added `setup-sfw` step to `slang-tests.yaml`, `cache-warmup.yaml` (all 5 jobs), and `release.yaml` (build job).

hedgar2017 (approved):

PR #316, opened 2026-04-03
feat(slang): checked arithmetic, narrow types, and MLIR alignment

hedgar2017:
- **slang_solidity updated** to latest commit; Rust toolchain bumped to 1.94.0
- **Windows path fix**: backslash → forward slash in compilation config
- **`catch_unwind` removal**: all 3 guards removed (latest slang no longer panics)
- **Type-directed expression emission**: `SemanticAnalysis` threaded through emitter chain; expressions emit in declared Solidity types
- **`Option<Value>` for void calls**: `emit()` returns `Option<Value>` — void calls return `None`
- **MLIR module attributes**: `llvm.data_layout` and `llvm.target_triple` for EVM target
- **Narrow typed signatures, `!sol.address`, unified `TypeConversion`**: `uint8` → `ui8`, `int128` → `si128`, `bool` → `i1`, `address` → `!sol.address`; all casts (explicit and implicit) routed through `TypeConversion::from_target_type().emit()`; non-integer elementary types (`bytes`, `string`, `fixed`) marked `unimplemented!` pending sol dialect types
- **ABI-based function naming**: `@"first()"` / `@"main(uint8,uint8)"` (dropped `solx.fn.` prefix)
- **Alloca ordering**: initializer emitted before alloca (matches solc)
- **Redundant op elimination**: `emit_is_nonzero` short-circuits on `i1`; `emit_sol_cast` skips no-op casts; comparisons return `i1` directly
- **Pass pipeline aligned with solc**: canonicalizer + modifier lowering + Sol-to-Standard + canonicalizer + SCF-to-CF + Func-to-LLVM + Arith-to-LLVM + CF-to-LLVM + reconcile casts
- **Checked arithmetic**: `sol.cadd`/`sol.csub`/`sol.cmul` for Solidity 0.8+ default, with unchecked mode for `unchecked {}` blocks and for-loop step expressions; dead yields in `sol.if` regions match solc's pattern
- **Logical AND/OR rewrite**: `sol.if` + `i1` alloca pattern (was `scf.if` + `ui256`)
- **Constructor emission**: default constructor with `FunctionKind::Constructor` attribute (new `solxCreateFunctionKindAttr` FFI stub) for contracts without one
- **BuiltIn member resolution**: match on slang's resolved BuiltIn variants instead of string-matching member names; address-returning intrinsics (`sol.caller`, `sol.origin`, `sol.coinbase`) use `!sol.address` return type
- **solx-dev in CI**: use it from `./target` even if `./target-slang` is the working target directory
- **solx-dev MLIR path**: add canonicalization as solc CMake expects absolute paths (even if unused for now, may be useful later for testing)
- **cargo clippy** alias for the Slang build.

## Integration tests (`tests/solidity/simple`)

**75.6%** passing (1,511 / 1,954 at M3B3)

## MLIR differences from solc

| Difference | solc | solx | Status |
|---|---|---|---|
| Boolean constants | `arith.constant false` | `sol.constant false` | TODO — functionally equivalent |
| Function naming | `@first_13` (name + AST node ID) | `@"first()"` (name + ABI signature) | By design — Slang node IDs are memory addresses, not stable |
| Contract naming | `@Test_26` | `@Test` | Same reason |
| `id` attribute | `id = 13 : i64` on functions | Not emitted | No stable IDs from Slang |

## Known issues

- Canonicalizer pass causes infinite loop on complex loop tests (e.g. `loop/complex/1.sol`) at `-Oz` optimization level

## Other TODOs

| TODO | Location | Context |
|---|---|---|
| Encode revert reason from second `require()` argument | `call/mod.rs` | Only the condition is checked; message string is ignored |
| Use declared element type once state variable types are tracked | `arithmetic.rs` | Compound assignment on state vars falls back to loaded value type |
| Implement checked exponentiation (`sol.cexp`) | `expression/mod.rs` | `**` operator always emits unchecked `sol.exp` |
| Support multi-return functions (tuple deconstruction) | `call/mod.rs` | Only single-return calls supported |
| Map `ReturnData` and `Code` to `DataLocation` | `data_location.rs` | Sol dialect has no equivalent yet; hits `unimplemented!` |


solx-mlir/src/context/builder.rs
```diff
@@ -773,25 +865,34 @@ impl<'context> Builder<'context> {
...
+    ///
+    /// Returns a `!sol.ptr<{element_type}, Stack>` pointer. Use this when
+    /// the declared Solidity type is known (e.g. `uint64` → `ui64`).
     ///
     /// # Panics
     ///
     /// Panics if the MLIR type or operation cannot be constructed, indicating
     /// a bug in the builder.
-    pub fn emit_sol_alloca<'block, B>(&self, block: &B) -> Value<'context, 'block>
+    pub fn emit_sol_alloca_typed<'block, B>(
```
hedgar2017: Do we still need old non-typed methods?

solx-slang/src/ast/contract/function/expression/arithmetic.rs
```diff
@@ -48,16 +63,19 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
 
     /// Emits prefix operators: `!`, `-`, `~`, `++`, `--`.
+    ///
+    /// When `target_type` is `Some`, unary operations use that type (matching
+    /// solc's typed MLIR). When `None`, falls back to ui256 semantics.
     pub fn emit_prefix(
         &self,
         operator: &str,
         operand: &Expression,
+        target_type: Option<Type<'context>>,
```
hedgar2017, line 85: When is `target_type` unavailable?

solx-slang/src/ast/contract/function/expression/arithmetic.rs
```diff
@@ -106,6 +138,26 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
+    ///
+    /// For types ≤ 64 bits, uses `emit_sol_constant_typed(-1, ty)` which
+    /// sign-extends correctly. For wider types, uses hex string literals.
+    fn emit_all_ones_constant(
+        &self,
+        ty: Type<'context>,
+        block: &BlockRef<'context, 'block>,
+    ) -> anyhow::Result<Value<'context, 'block>> {
+        let bit_width = Self::integer_bit_width(ty);
+        if bit_width <= 64 {
```
hedgar2017: 64 bits have no semantic meaning in EVM architecture.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
```diff
@@ -130,27 +181,42 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
                     .get(&state_variable.node_id())
                     .ok_or_else(|| anyhow::anyhow!("unregistered state variable: {name}"))?;
                 let old = self.emit_storage_load(slot, block)?;
+                // TODO: use declared element type once state variable types are tracked.
```
hedgar2017, line 176: Why not fix it at once?

solx-slang/src/ast/contract/function/expression/call.rs
```diff
@@ -45,45 +47,36 @@ impl<'emitter, 'state, 'context, 'block> CallEmitter<'emitter, 'state, 'context,
...
+                Some("bool") => {
+                    // bool(x) → x != 0, producing i1.
+                    let zero = builder.emit_sol_constant_typed(0, value.r#type(), &block);
+                    let result = builder.emit_sol_cmp(value, zero, CmpPredicate::Ne, &block);
+                    Ok((Some(result), block))
                 }
-                "uint8" => {
+                Some("uint8") => {
                     let mask = builder.emit_sol_constant(Self::UINT8_MASK, &block);
                     let ui256 = builder.get_type(solx_mlir::Builder::UI256);
```
hedgar2017: Why is UI256 hardcoded here?

solx-slang/src/ast/contract/function/expression/mod.rs
```diff
@@ -23,52 +28,85 @@ use solx_mlir::Environment;
...
         block: BlockRef<'context, 'block>,
-    ) -> anyhow::Result<(Value<'context, 'block>, BlockRef<'context, 'block>)> {
+    ) -> anyhow::Result<(Option<Value<'context, 'block>>, BlockRef<'context, 'block>)> {
         match expression {
             Expression::DecimalNumberExpression(decimal) => {
                 let literal = decimal.literal();
                 let text = literal.text.as_str();
-                let value = if let Ok(value) = text.parse::<i64>() {
-                    self.state.builder.emit_sol_constant(value, &block)
+                let min_type = if let Ok(unsigned_value) = text.parse::<u128>() {
```
hedgar2017: 128 bit have no semantic meaning on EVM architecture.

solx-slang/src/ast/contract/function/expression/mod.rs
```diff
@@ -77,22 +115,37 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
                     .strip_prefix("0x")
                     .or(text.strip_prefix("0X"))
                     .unwrap_or(text);
-                let value = if let Ok(parsed) = i64::from_str_radix(stripped, 16) {
-                    self.state.builder.emit_sol_constant(parsed, &block)
+                let min_type = if let Ok(unsigned_value) = u128::from_str_radix(stripped, 16) {
```
hedgar2017: Use constants from solx-utils

solx-slang/src/ast/contract/function/expression/mod.rs
```diff
@@ -205,39 +298,54 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
             Expression::ExponentiationExpression(expression) => {
+                let target_type = self.resolve_expression_type(expression.node_id());
                 let left = expression.left_operand();
                 let right = expression.right_operand();
-                let (lhs, block) = self.emit(&left, block)?;
-                let (rhs, block) = self.emit(&right, block)?;
+                let (lhs, block) = self.emit_value(&left, block)?;
+                let (rhs, block) = self.emit_value(&right, block)?;
+                let result_type = target_type
+                    .unwrap_or_else(|| self.state.builder.get_type(solx_mlir::Builder::UI256));
```
hedgar2017, line 281: Does this default have semantic meaning in Solidity?

solx-slang/src/ast/contract/function/expression/mod.rs
```diff
@@ -205,39 +298,54 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
+                    result_type,
                     &block,
                 )?;
-                Ok((result, block))
+                Ok((Some(result), block))
             }
             Expression::ConditionalExpression(conditional) => {
+                let result_type = self
+                    .resolve_expression_type(conditional.node_id())
+                    .unwrap_or_else(|| self.state.builder.get_type(solx_mlir::Builder::UI256));
```
hedgar2017, line 304: Does this default have semantic meaning in Solidity?

solx-slang/src/ast/contract/function/expression/mod.rs
```diff
@@ -246,16 +354,89 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
+    }
+
+    /// Returns the bit width of an MLIR integer type, or 256 for unknown types.
+    fn integer_bit_width(ty: Type<'_>) -> u32 {
+        IntegerType::try_from(ty).map_or(256, |integer_type| integer_type.width())
+    }
+
+    /// Returns the narrowest unsigned integer type that can hold `value`.
+    fn min_uint_type_for<'ctx>(value: u128, state: &Context<'ctx>) -> Type<'ctx> {
+        let bit_width = if value <= 0xff {
```
hedgar2017: 1. Rewrite using a formula.
2. Explain why non-power-of-2 bitwidths are skipped.

solx-slang/src/ast/contract/function/expression/mod.rs
```diff
@@ -246,16 +354,89 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
+        }
+        if Self::integer_bit_width(lhs_type) >= Self::integer_bit_width(rhs_type) {
+            (lhs, self.state.builder.emit_sol_cast(rhs, lhs_type, block))
+        } else {
+            (self.state.builder.emit_sol_cast(lhs, rhs_type, block), rhs)
+        }
+    }
+
+    /// Returns the bit width of an MLIR integer type, or 256 for unknown types.
+    fn integer_bit_width(ty: Type<'_>) -> u32 {
```
hedgar2017: Move private functions to the bottom.

solx-slang/src/ast/contract/function/statement/mod.rs
```diff
@@ -33,21 +38,32 @@ pub struct StatementEmitter<'state, 'context, 'block> {
     region_pointer: *const Region<'context>,
     /// State variable node ID to storage slot mapping.
     pub storage_layout: &'state HashMap<NodeId, u64>,
+    /// The function's declared return types, for `emit_return` to cast to.
```
hedgar2017, line 42: Inconsistent public/private fields.

solx-slang/src/ast/contract/function/statement/mod.rs
```diff
@@ -197,9 +255,25 @@ impl<'state, 'context, 'block> StatementEmitter<'state, 'context, 'block> {
         block: BlockRef<'context, 'block>,
     ) -> anyhow::Result<Option<BlockRef<'context, 'block>>> {
         if let Some(ref expression) = return_statement.expression() {
-            let emitter = ExpressionEmitter::new(self.state, self.environment, self.storage_layout);
-            let (value, block) = emitter.emit(expression, block)?;
-            self.state.builder.emit_sol_return(&[value], &block);
+            anyhow::ensure!(
```
hedgar2017, line 258: Add a TODO as well.

solx-slang/src/ast/contract/function/mod.rs
```diff
@@ -22,8 +26,60 @@ use solx_mlir::StateMutability;
 
 use self::statement::StatementEmitter;
 
+/// Maps a Solidity `TypeName` to an MLIR type.
```
hedgar2017: Free functions not allowed.

solx-slang/src/ast/contract/function/mod.rs
```diff
@@ -22,8 +26,60 @@ use solx_mlir::StateMutability;
...
+        ElementaryType::AddressType(_)
+        | ElementaryType::ByteKeyword
+        | ElementaryType::StringKeyword
+        | ElementaryType::BytesKeyword(_)
+        | ElementaryType::FixedKeyword(_)
+        | ElementaryType::UfixedKeyword(_) => builder.get_type(solx_mlir::Builder::UI256),
+    }
+}
+
+/// Parses the bit width from a Solidity integer type name (e.g. "uint64" → 64).
```
hedgar2017: Avoid garbage free functions.

solx-slang/src/ast/contract/function/mod.rs
```diff
@@ -22,8 +26,60 @@ use solx_mlir::StateMutability;
...
+fn parse_integer_bits(text: &str, prefix: &str) -> u32 {
+    text.strip_prefix(prefix)
+        .and_then(|suffix| suffix.parse::<u32>().ok())
+        .unwrap_or(256)
```
hedgar2017: solx-utils

solx-slang/src/ast/contract/function/expression/call.rs
```diff
@@ -122,14 +115,41 @@ impl<'emitter, 'state, 'context, 'block> CallEmitter<'emitter, 'state, 'context,
...
+                        // sol.cast to i1 would produce llvm.trunc which gives wrong
+                        // results: bool(2) → false because bit 0 is zero.
+                        if target_type == builder.get_type(solx_mlir::Builder::I1) {
+                            let zero = builder.emit_sol_constant_typed(0, value.r#type(), &block);
+                            let result =
+                                builder.emit_sol_cmp(value, zero, CmpPredicate::Ne, &block);
+                            Ok((Some(result), block))
+                        } else {
+                            let result = builder.emit_sol_cast(value, target_type, &block);
+                            Ok((Some(result), block))
```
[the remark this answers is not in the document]
hedgar2017: Fixed. Rewrote the entire type conversion block — removed all per-type named arms and replaced with a generic TypeConversion enum that handles all elementary types uniformly. AddressType now always gets 160-bit truncation regardless of callee form.

solx-mlir/src/context/builder.rs
hedgar2017: Removed. Deleted emit_sol_constant, emit_sol_constant_from_decimal_str, and emit_sol_constant_from_hex_str (all untyped, hardcoded ui256). Literal parsing now uses string-based typed variants exclusively.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
hedgar2017: When Slang semantic analysis lacks type info for the expression node (e.g. terminal nodes, some unsupported constructs). The fallback selects the wider operand type by bit width.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
hedgar2017: Fixed. The 64-bit branch was a Rust i64 limit leaking into EVM logic. Removed the branching — emit_all_ones_constant now uses hex string emission for all bit widths uniformly.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
hedgar2017: Requires extending storage_layout to track declared element types alongside slot offsets — separate scope from this PR. The TODO documents the gap.

solx-slang/src/ast/contract/function/expression/call.rs
hedgar2017: Fixed. Removed the uint8-specific arm entirely. All type conversions now go through a generic handler that resolves the correct MLIR type for every elementary type.

solx-slang/src/ast/contract/function/expression/mod.rs
hedgar2017: Fixed. Removed parse::<u128>() and parse::<i64>() entirely. Literal parsing now uses string-based emission exclusively — no Rust numeric type limits in the path.

solx-slang/src/ast/contract/function/expression/mod.rs
hedgar2017: Fixed. Now uses solx_utils::BIT_LENGTH_FIELD and solx_utils::BIT_LENGTH_BYTE throughout.

solx-slang/src/ast/contract/function/expression/mod.rs
hedgar2017: Yes — Solidity exponentiation defaults to uint256 when the type cannot be inferred from context. This matches solc behavior.

solx-slang/src/ast/contract/function/expression/mod.rs
hedgar2017: Yes — Solidity ternary expressions default to uint256 when the result type cannot be inferred.

solx-slang/src/ast/contract/function/expression/mod.rs
hedgar2017: Fixed. Rewrote with formula: decimal uses ceil(digits * 3322 / 1000) approximation, hex uses significant_digits * 4. Both round up via round_up_to_solidity_width which uses 8-bit increments (Solidity supports uint8, uint16, uint24, ..., uint256 — all multiples of 8, not just powers of 2).

solx-slang/src/ast/contract/function/expression/mod.rs
hedgar2017: Fixed. Private functions moved to bottom of impl block.

solx-slang/src/ast/contract/function/statement/mod.rs
hedgar2017: Fixed. All fields are now private.

solx-slang/src/ast/contract/function/statement/mod.rs
hedgar2017: Added: // TODO: support multi-return functions (tuple deconstruction).

solx-slang/src/ast/contract/function/mod.rs
hedgar2017: Fixed. Moved type_name_to_mlir_type, elementary_to_mlir_type, and parse_integer_bits into FunctionEmitter as associated methods.

solx-slang/src/ast/contract/function/mod.rs
hedgar2017: Fixed. All moved into FunctionEmitter.

solx-slang/src/ast/contract/function/mod.rs
hedgar2017: Fixed. Default bit width now uses solx_utils::BIT_LENGTH_FIELD.

solx-slang/src/ast/contract/function/expression/call.rs
```diff
@@ -244,11 +191,119 @@ impl<'emitter, 'state, 'context, 'block> CallEmitter<'emitter, 'state, 'context,
...
+                )
+            }
+            TypeConversion::Cast(target_type) => {
+                Ok(builder.emit_sol_cast(value, target_type, block))
+            }
+        }
+    }
+}
+
+/// Classification of elementary type conversions.
```
hedgar2017: Not allowed.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
```diff
@@ -106,6 +138,19 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
     }
 
+    /// Emits an all-ones constant for the given integer type.
+    fn emit_all_ones_constant(
+        &self,
+        ty: Type<'context>,
+        block: &BlockRef<'context, 'block>,
+    ) -> anyhow::Result<Value<'context, 'block>> {
+        let bit_width = Self::integer_bit_width(ty);
+        let hex = "f".repeat(bit_width as usize / 4);
```
hedgar2017: Slop.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
```diff
@@ -66,38 +84,52 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
+                let zero = self
+                    .state
+                    .builder
+                    .emit_sol_constant_typed(0, op_type, &block);
                 let result = self.state.builder.emit_binary_operation(
                     solx_mlir::Builder::SOL_SUB,
                     zero,
                     value,
-                    self.state.builder.get_type(solx_mlir::Builder::UI256),
+                    op_type,
```
hedgar2017: No contractions.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
```diff
@@ -106,6 +138,19 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
         }
     }
 
+    /// Emits an all-ones constant for the given integer type.
+    fn emit_all_ones_constant(
+        &self,
+        ty: Type<'context>,
```
hedgar2017: No contractions.

solx-slang/src/ast/contract/function/expression/logical.rs
```diff
@@ -48,34 +58,45 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
-        let rhs_normalized = self
+        let i1_type = self.state.builder.get_type(solx_mlir::Builder::I1);
+        let result_ptr = self.state.builder.emit_sol_alloca_typed(i1_type, &block);
+        // TODO: solc uses `arith.constant false` here; consider switching to
+        // arith dialect for i1 constants to match solc output exactly.
+        let false_val = self
             .state
             .builder
-            .emit_sol_cast_to_ui256(rhs_bool, &then_end);
+            .emit_sol_constant_typed(0, i1_type, &block);
```
hedgar2017: If you've removed non-typed methods, remove the `_typed` suffix everywhere.

solx-slang/src/ast/contract/function/statement/control_flow.rs
```diff
@@ -177,8 +196,14 @@ impl<'state, 'context, 'block> StatementEmitter<'state, 'context, 'block> {
...
-        let emitter = ExpressionEmitter::new(self.state, self.environment, self.storage_layout);
-        let (condition_value, cond_end) = emitter.emit(&condition_expression, cond_block)?;
+        let emitter = ExpressionEmitter::new(
+            &self.semantic,
+            self.state,
+            self.environment,
+            self.storage_layout,
+            self.checked,
+        );
+        let (condition_value, cond_end) = emitter.emit_value(&condition_expression, cond_block)?;
```
hedgar2017: No contractions.

solx-slang/src/ast/contract/function/statement/mod.rs
```diff
@@ -151,24 +177,56 @@ impl<'state, 'context, 'block> StatementEmitter<'state, 'context, 'block> {
...
 
-        let block = if let Some(ref initializer_expression) = declaration.value() {
-            let (initial_value, block) = emitter.emit(initializer_expression, block)?;
-            emitter
-                .state
-                .builder
-                .emit_sol_store(initial_value, pointer, &block);
-            block
+        // Evaluate initializer before alloca to match solc's emission order
+        // (constant → cast → alloca → store).
```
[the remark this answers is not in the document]
hedgar2017: Fixed. Updated the comment to accurately describe both branches.

solx-mlir/src/context/builder.rs
```diff
@@ -329,52 +357,44 @@ impl<'context> Builder<'context> {
...
-            .ok_or_else(|| anyhow::anyhow!("invalid ui256 decimal literal: {value}"))?;
-        self.emit_constant_operation(
-            Self::SOL_CONSTANT,
-            attribute,
-            self.get_type(Self::UI256),
-            block,
-        )
+        let attribute = Attribute::parse(self.context, &format!("{value} : {ty}"))
+            .ok_or_else(|| anyhow::anyhow!("invalid {ty} decimal literal: {value}"))?;
+        self.emit_constant_operation(Self::SOL_CONSTANT, attribute, ty, block)
```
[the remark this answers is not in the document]
hedgar2017: Acknowledged. Deferred — no correctness impact, performance optimization can follow in a separate PR.

solx-slang/src/ast/contract/function/expression/call/mod.rs
```diff
@@ -244,11 +191,119 @@ impl<'emitter, 'state, 'context, 'block> CallEmitter<'emitter, 'state, 'context,
...
+                    ui256,
+                    block,
+                )?;
+                builder.emit_binary_operation(
+                    solx_mlir::Builder::SOL_AND,
+                    value,
+                    mask,
+                    ui256,
+                    block,
+                )
```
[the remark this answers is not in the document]
hedgar2017: Fixed. Added `emit_sol_cast(value, ui256, block)` before the mask computation.

solx-slang/src/ast/contract/function/expression/call.rs
hedgar2017: Fixed. Moved TypeConversion to a child module: `call/type_conversion.rs`.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
hedgar2017: Fixed.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
hedgar2017: Fixed. `op_type` → `operand_type`.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
hedgar2017: Fixed. `ty` → `integer_type`.

solx-slang/src/ast/contract/function/expression/logical.rs
hedgar2017: Fixed. Removed `_typed` suffix from all methods and deleted the remaining untyped `emit_sol_load` variant.

solx-slang/src/ast/contract/function/statement/control_flow.rs
hedgar2017: Fixed. `cond_block` → `condition_block`, `cond_end` → `condition_end`.

PavelKopyl (approved):
LGTM.

PavelKopyl (changes requested):
Can we split this into several commits to simplify the review? For instance, checked arithmetic, type conversions and so on.

solx-mlir/src/context/builder.rs
```diff
@@ -283,12 +319,14 @@ impl<'context> Builder<'context> {
                 )
                 .into(),
             ));
-            attributes.push((
-                Identifier::new(self.context, "orig_fn_type"),
-                TypeAttribute::new(function_type.into()).into(),
-            ));
         }
 
+        // Always emit orig_fn_type so the lowering pass knows the original signature.
```
abinavpp: nit: orig_fn_type is only used by the lowering pass for ABI decoding at dispatch entry points. so we only need it for things like external functions, constructors, getters. setting it unconditionally on every sol.func is harmless though

solx-slang/src/ast/contract/function/expression/call/mod.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// # Errors
+    ///
+    /// Returns an error if the callee is unsupported, arguments contain
+    /// unsupported constructs, or the function is undefined.
+    pub fn emit_function_call(
+        &self,
+        callee: &Expression,
+        arguments: &ArgumentsDeclaration,
+        is_type_conversion: bool,
+        call_node_id: NodeId,
```
abinavpp: nit: do you think we can pass the ast node here instead?

solx-slang/src/ast/contract/function/expression/call/mod.rs
```diff
@@ -0,0 +1,261 @@
...
+    fn emit_type_conversion(
+        &self,
+        value: Value<'context, 'block>,
+        conversion: TypeConversion<'context>,
+        builder: &solx_mlir::Builder<'context>,
+        block: &BlockRef<'context, 'block>,
+    ) -> anyhow::Result<Value<'context, 'block>> {
+        match conversion {
+            TypeConversion::Bool => {
+                // bool(x) → x != 0. sol.cast to i1 would produce llvm.trunc
```
abinavpp: thanks for finding this!

solx-slang/src/ast/contract/function/expression/call/mod.rs
```diff
@@ -0,0 +1,261 @@
...
+                    one,
+                    ui256,
+                    block,
+                )?;
+                builder.emit_binary_operation(
+                    solx_mlir::Builder::SOL_AND,
+                    value,
+                    mask,
+                    ui256,
+                    block,
```
abinavpp: we should use sol.cast to narrow to ui160, then sol.address_cast. this requires adding !sol.address to the type system. can we have that in this pr? otherwise i feel it's better to leave this unimplemented.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
```diff
@@ -15,27 +17,38 @@ use crate::ast::contract::function::expression::ExpressionEmitter;
...
+        let result_type = target_type.unwrap_or_else(|| {
+            let lhs_width = solx_mlir::Builder::integer_bit_width(lhs.r#type());
+            let rhs_width = solx_mlir::Builder::integer_bit_width(rhs.r#type());
+            if lhs_width >= rhs_width {
+                lhs.r#type()
+            } else {
+                rhs.r#type()
+            }
+        });
+        let lhs = self.state.builder.emit_sol_cast(lhs, result_type, &block);
```
abinavpp: here and in other places, can we unify this cast codegen to run through the full TypeConversion rules?

solx-slang/src/ast/contract/function/statement/control_flow.rs
```diff
@@ -55,12 +61,17 @@ impl<'state, 'context, 'block> StatementEmitter<'state, 'context, 'block> {
                 // Both branches terminated (e.g. both return). The parent
                 // block needs a terminator after sol.if since sol.if is not
                 // a terminator itself.
+                // TODO: solc emits dead-code (alloca + zero + sol.return) here
+                // and ^bb1 dead blocks with sol.yield after each sol.return
+                // inside the if/else; we emit llvm.unreachable instead.
```
abinavpp: we should not generate an unreachable here. a front end should not worry about generating unreachable code, but instead just let the lowering somewhere up in the call-chain (like function body lowering) to generate the terminator, even if it's dead code. cleaning up in reachable code should be done in a separate pass.

(side note: writing "^bb1" here is not clear here without an example ir)

solx-slang/src/ast/contract/function/statement/control_flow.rs
```diff
@@ -55,12 +61,17 @@ impl<'state, 'context, 'block> StatementEmitter<'state, 'context, 'block> {
...
                 block.append_operation(melior::dialect::llvm::unreachable(
                     self.state.builder.unknown_location,
                 ));
                 return Ok(None);
             }
         } else {
+            // TODO(sol-dialect): solc emits an empty else region here; dropping
```
abinavpp: this is expected. the client should generate the yield ops.

solx-slang/src/ast/contract/function/mod.rs
```diff
@@ -240,4 +284,54 @@ impl<'state, 'context> FunctionEmitter<'state, 'context> {
...
+            ElementaryType::IntKeyword(terminal) => {
+                let bits = Self::parse_integer_bits(&terminal.text, "int");
+                Type::from(IntegerType::signed(context, bits))
+            }
+            ElementaryType::AddressType(_)
+            | ElementaryType::ByteKeyword
+            | ElementaryType::StringKeyword
+            | ElementaryType::BytesKeyword(_)
+            | ElementaryType::FixedKeyword(_)
+            | ElementaryType::UfixedKeyword(_) => builder.get_type(solx_mlir::Builder::UI256),
```
abinavpp: there are sol dialect types for this. we should leave this unimplemented if we're not supporting those types in this pr

solx-slang/src/ast/contract/function/mod.rs
```diff
@@ -240,4 +284,54 @@ impl<'state, 'context> FunctionEmitter<'state, 'context> {
...
+        match type_name {
+            TypeName::ElementaryType(elementary) => {
+                Self::elementary_to_mlir_type(elementary, context, builder)
+            }
+            _ => builder.get_type(solx_mlir::Builder::UI256),
+        }
+    }
+
+    /// Maps an elementary Solidity type to an MLIR type.
+    pub fn elementary_to_mlir_type(
```
abinavpp: does the slang ast have a more refined data-structures for types? so that we can avoid things like the parse_integer_bits below.

solx-slang/src/ast/contract/function/statement/control_flow.rs
```diff
@@ -147,10 +163,16 @@ impl<'state, 'context, 'block> StatementEmitter<'state, 'context, 'block> {
...
+        // Step region — always unchecked (matches solc: loop step i++ uses sol.add).
         if let Some(ref iterator_expression) = for_statement.iterator() {
-            let emitter = ExpressionEmitter::new(self.state, self.environment, self.storage_layout);
-            let (_value, step_end) = emitter.emit(iterator_expression, step_block)?;
+            let emitter = ExpressionEmitter::new(
+                &self.semantic,
+                self.state,
+                self.environment,
+                self.storage_layout,
+                false,
```
abinavpp: nit: comment "unchecked=false"?

abinavpp (changes requested):
thank you! this is looking better. iiuc, there are some divergences in the codegen that imo would be better left unimplemented than implemented differently. but thanks for the divergence that exposed bugs in the solc-mlir integration!

solx-slang/src/ast/contract/function/statement/control_flow.rs
hedgar2017: By a separate pass, do you mean LLVM optimizer DCE?

PavelKopyl: This clearing is already done by a MLIR optimization pass.

PavelKopyl:
Please, remove "Co-Authored-By: Claude Opus" form the commit messages)

abinavpp:
just some high level questions:

do we have source location tracking in slang ast? we could start forwarding them in the lowering. i remember discussing with Gus that it was byte-offset based, which would be fine imo. we could track that instead of line+column.

are we sure that this string based mlir generation (i.e. using literal names of ops, attributes etc.) is the best we can do? i was discussing with claude, and it mentioned about `melior_macro::dialect!`? ideally we should work directly with data-structures from tablegen (https://github.com/matter-labs/solx-llvm/blob/main/mlir/include/mlir/Dialect/Sol/SolOps.td, SolBase.td etc.). i will also try to find out, i'm still learning rust :sweat_smile:

hedgar2017:
> Please, remove "Co-Authored-By: Claude Opus" form the commit messages)

@PavelKopyl removed 😿

hedgar2017:
> do we have source location tracking in slang ast? we could start forwarding them in the lowering. i remember discussing with Gus that it was byte-offset based, which would be fine imo. we could track that instead of line+column.

Certainly, and we can have both eventually depending on our needs. IMO it's too early to care about code locations, I'd address them after the main codegen part is done. Perhaps, @OmarTawfik or @ggiraldez can answer with more detail.

> are we sure that this string based mlir generation (i.e. using literal names of ops, attributes etc.) is the best we can do? i was discussing with claude, and it mentioned about melior_macro::dialect!? ideally we should work directly with data-structures from tablegen (https://github.com/matter-labs/solx-llvm/blob/main/mlir/include/mlir/Dialect/Sol/SolOps.td, SolBase.td etc.). i will also try to find out, i'm still learning rust 😅

Sure, I'm planning to switch to those eventually. I just don't want to fork melior or spend a lot of time on it yet, as from what I understand, using this macro requires writing custom build scripts to parse the solc dialect *.td and generate Rust bindings from it.
Also, I'm somewhat cautious of using macros historically, as they are not base language and are much harder to debug, especially at early stages of project development.
But if you insist, I may look into it after this PR 😉

PavelKopyl:
My main concern is that Sol dialect types are not supported here — everything is mapped to uint256 instead. I would suggest first adding proper support for all Sol dialect types, potentially in a separate PR. Without this, the generated MLIR is invalid, meaning that for any reasonably complex program we silently produce invalid LLVM IR, since the Sol dialect lowering relies entirely on correct value types. Additionally, the lack of type information prevents automatic comparison between the Sol dialect output generated by solx-mlir and solc-mlir.
For the following test:
```contract c {
  uint[] dsc;
  function f(uint[] calldata src) public {
    dsc = src;
  }
}
```
`solx-mlir` produces:
```
#Constructor = #sol<FunctionKind Constructor>
#Contract = #sol<ContractKind Contract>
#NonPayable = #sol<StateMutability NonPayable>
#Osaka = #sol<EvmVersion Osaka>
module attributes {llvm.data_layout = "E-p:256:256-i256:256:256-S256-a:256:256", llvm.target_triple = "evm-unknown-unknown", sol.evm_version = #Osaka} {
  sol.contract @c {
    sol.state_var @slot_0 slot 0 offset 0 : ui256
    sol.func @"f(uint256[])"(%arg0: ui256) attributes {orig_fn_type = (ui256) -> (), selector = 2076556223 : i32, state_mutability = #NonPayable} {
      %0 = sol.alloca : !sol.ptr<ui256, Stack>
      sol.store %arg0, %0 : ui256, !sol.ptr<ui256, Stack>
      %1 = sol.load %0 : !sol.ptr<ui256, Stack>, ui256
      %2 = sol.addr_of @slot_0 : !sol.ptr<ui256, Storage>
      sol.store %1, %2 : ui256, !sol.ptr<ui256, Storage>
      sol.return
    }
    sol.func @"constructor()"() attributes {kind = #Constructor, orig_fn_type = () -> (), state_mutability = #NonPayable} {
      sol.return
    }
  } {kind = #Contract}
}
```
whereas `solc-mlir`:
```
#Constructor = #sol<FunctionKind Constructor>
#Contract = #sol<ContractKind Contract>
#Default = #sol<RevertStrings Default>
#NonPayable = #sol<StateMutability NonPayable>
#Osaka = #sol<EvmVersion Osaka>
module attributes {llvm.data_layout = "E-p:256:256-i256:256:256-S256-a:256:256", llvm.target_triple = "evm-unknown-unknown", sol.evm_version = #Osaka, sol.revert_strings = #Default} {
  sol.contract @c_16 {
    sol.state_var @dsc_4 slot 0 offset 0 : !sol.array<? x ui256, Storage>
    sol.func @c_16() attributes {kind = #Constructor, orig_fn_type = () -> (), state_mutability = #NonPayable} {
      sol.return
    }
    sol.func @f_15(%arg0: !sol.array<? x ui256, CallData>) attributes {id = 15 : i64, orig_fn_type = (!sol.array<? x ui256, CallData>) -> (), selector = 2076556223 : i32, state_mutability = #NonPayable} {
      %0 = sol.alloca : !sol.ptr<!sol.array<? x ui256, CallData>, Stack>
      sol.store %arg0, %0 : !sol.array<? x ui256, CallData>, !sol.ptr<!sol.array<? x ui256, CallData>, Stack>
      %1 = sol.addr_of @dsc_4 : !sol.array<? x ui256, Storage>
      %2 = sol.load %0 : !sol.ptr<!sol.array<? x ui256, CallData>, Stack>, !sol.array<? x ui256, CallData>
      sol.copy %2, %1 : !sol.array<? x ui256, CallData>, !sol.array<? x ui256, Storage>
      sol.return
    }
  } {kind = #Contract}
}
```

Even though `solx-mlir` could emit `sol.copy` instead of sol.store, the copy operation may not have been properly lowered with no type info.

hedgar2017:
@PavelKopyl but arrays are out of scope of this PR. The PR description only mentions primitive types.

I will remove the placeholders @abinavpp mentioned, but they are harmless since those types are not supported yet anyway.

Speaking of invalid MLIR, we cannot generate it 'silently' because solx-tester always fails on it :)

PavelKopyl:
> @PavelKopyl but arrays are out of scope of this PR. The PR description only mentions primitive types.
> 
> I will remove the placeholders @abinavpp mentioned, but they are harmless since those types are not supported yet anyway.
> 
> Speaking of invalid MLIR, we cannot generate it 'silently' because solx-tester always fails on it :)

Yes, they are out of scope and that is fine. The issue is that we can compile such examples and emit MLIR and LLVM IR that turn out to be invalid, which is unexpected. I agree with @abinavpp let's remove all such placeholders and insert an equivalent of llvm_unreachable("Not yet implemented"). This way we can clearly distinguish tests that are not yet supported from those that are, and for the supported ones we can compare the Sol MLIR output between solx and solc.

solx-mlir/src/context/builder.rs
hedgar2017: Done! Now only emitted for constructors and functions with a selector.

solx-slang/src/ast/contract/function/expression/call/mod.rs
hedgar2017: Yeah, good idea!

solx-slang/src/ast/contract/function/statement/control_flow.rs
hedgar2017: Done, but now some tests started getting into an infinite loop in Mz. It looks like the hang is inside LLVM. I've left a TODO and will be only running M3B3 for now.

solx-slang/src/ast/contract/function/statement/control_flow.rs
hedgar2017: Nice, done!

solx-slang/src/ast/contract/function/statement/control_flow.rs
hedgar2017: Understood, thanks!

ggiraldez:
> > do we have source location tracking in slang ast? we could start forwarding them in the lowering. i remember discussing with Gus that it was byte-offset based, which would be fine imo. we could track that instead of line+column.
> 
> Certainly, and we can have both eventually depending on our needs. IMO it's too early to care about code locations, I'd address them after the main codegen part is done. Perhaps, @OmarTawfik or @ggiraldez can answer with more detail.
> 
v1 does provide `text_offset` information for AST nodes, which includes both byte-offsets and line/column. It's missing file name information, but that's easy to add.

We don't have a working solution in v2 yet, but we're very close. Byte offsets are the easiest to provide because that's what the parser generates for terminal nodes. We had some discussion about providing mapping to line/column as well as part of the AST/CompilationUnit API.

(Sorry @hedgar2017 I messed up your comment, I hit edit instead of quote reply :facepalm:)

solx-slang/src/ast/contract/function/mod.rs
hedgar2017: Fixec!

solx-slang/src/ast/contract/function/expression/call/mod.rs
hedgar2017: Done!

solx-slang/src/ast/contract/function/mod.rs
hedgar2017: It looks like `terminal_text` is the only thing it exposes, but I'm certainly planning to move methods such as this one there. Added a TODO.
@ggiraldez would it make sense to move it as is?

solx-slang/src/ast/contract/function/expression/arithmetic.rs
hedgar2017: Why not! Moved this logic to `TypeConversion`, but let's not overengineer it for now.

solx-slang/src/ast/contract/function/statement/control_flow.rs
```diff
@@ -218,14 +236,30 @@ impl<'state, 'context, 'block> StatementEmitter<'state, 'context, 'block> {
...
+            .emit_sol_condition(condition_boolean, &condition_end);
 
         self.region_pointer = saved_region;
         Ok(Some(block))
     }
+
+    /// Appends a dead block with `sol.yield` to a region whose live block
+    /// already terminated (e.g. with `sol.return`). Matches the solc pattern
+    /// where each `sol.if` region always ends with a `sol.yield` block.
+    fn emit_dead_yield(&self, region: &melior::ir::Region<'context>) {
```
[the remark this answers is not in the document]
hedgar2017: Out of scope for now.

solx-slang/src/ast/contract/function/mod.rs
ggiraldez: You should have typing information on the definition node. For example, I see `type_name_to_mlir_type` is being called on the `TypeName` of each `Parameter`, which is a step too deep. `Parameter::get_type()` should return an [`ast::Type`](https://github.com/NomicFoundation/slang/blob/a18c2eb5004da3c5ee567d429bbda6d47c89b696/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/types.rs#L11) which will have complete type information. If that returns `None` then binding failed and there's probably an error in the Solidity source, because we don't do validation yet.

I think we can do better with types of literals, but for others you should be able to map directly without needing to parse the textual representation of the type.

We don't attach the type to the type name itself, because that would be a meta-type or sorts. That's generally true throughout the AST tree. You should query the identifiers or expressions for the types, not the nodes which the define the types themselves.

hedgar2017: @ggiraldez thanks for the suggestion, worked like a charm! All the ugly code has been removed.

abinavpp (approved):
lgtm. thank you!

abinavpp:
> Sure, I'm planning to switch to those eventually. I just don't want to fork melior or spend a lot of time on it yet, as from what I understand, using this macro requires writing custom build scripts to parse the solc dialect *.td and generate Rust bindings from it.
Also, I'm somewhat cautious of using macros historically, as they are not base language and are much harder to debug, especially at early stages of project development.
But if you insist, I may look into it after this PR 😉

imo, we should do this asap before we have more and more string based codegen. the migration cost only grows with time. the concrete risk with strings: say we rename orig_fn_type to original_fn_type in SolOps.td, or remove an attribute entirely. the c++ side breaks immediately (TableGen regenerates the .h.inc files). But the Rust side keeps compiling - it just silently sets an attribute that nothing reads. You find out at runtime, maybe, if you're lucky.

nebasuke:
> > Sure, I'm planning to switch to those eventually. I just don't want to fork melior or spend a lot of time on it yet, as from what I understand, using this macro requires writing custom build scripts to parse the solc dialect *.td and generate Rust bindings from it.
> > Also, I'm somewhat cautious of using macros historically, as they are not base language and are much harder to debug, especially at early stages of project development.
> > But if you insist, I may look into it after this PR 😉
> 
> imo, we should do this asap before we have more and more string based codegen. the migration cost only grows with time. the concrete risk with strings: say we rename orig_fn_type to original_fn_type in SolOps.td, or remove an attribute entirely. the c++ side breaks immediately (TableGen regenerates the .h.inc files). But the Rust side keeps compiling - it just silently sets an attribute that nothing reads. You find out at runtime, maybe, if you're lucky.

I think this needs a short technical mini-doc before merge. This is a structural choice with longer-term architectural consequences, not just an implementation preference. The migration cost only grows with time, and string-based codegen also creates a risk of silent breakage that tests may not catch early.

hedgar2017:
@abinavpp @nebasuke I can certainly do this migration to TableGen as the next PR (shared some of my thoughts in the Notion doc as well), but I don't think we should delay this one any longer, and just do the migration in another one.

solx-slang/src/ast/contract/function/expression/assignment.rs
```diff
@@ -56,34 +58,60 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
+                    &self.state.builder,
+                )
+                .emit(value, &self.state.builder, &block);
+                self.state
+                    .builder
+                    .emit_sol_store(stored_value, pointer, &block);
+                stored_value
             }
             AssignmentTarget::Storage(slot) => {
-                self.emit_storage_store(slot, value, &block)?;
```
PavelKopyl, line 84: I guess here we need casting to ui256 (as Storage slots are always ui256) instead of storing the value directly. I mean line 111.

solx-slang/src/ast/contract/function/expression/operator.rs
```diff
@@ -101,13 +101,20 @@ impl Operator {
 
     /// Returns the Sol dialect operation name for arithmetic, bitwise, and step operators.
     ///
+    /// When `checked` is true, uses `sol.cadd`/`sol.csub`/`sol.cmul` for
+    /// add/subtract/multiply (Solidity 0.8+ default). Division, modulo,
```
PavelKopyl: We also have sol.cdiv to handle `type(int256).min / -1`.

PavelKopyl (approved):
LGTM, thank you!

solx-slang/src/ast/contract/function/expression/assignment.rs
hedgar2017: Easy fix, done!

solx-slang/src/ast/contract/function/expression/operator.rs
hedgar2017: Easy fix, done!

PR #335, opened 2026-04-16
fix: cargo-deny failure from vulnerable rustls-webpki

hedgar2017:
The failing `cargo-checks` job on PR #334 was not caused by the `docker/build-push-action` bump itself. The job was failing in `cargo deny` because `Cargo.lock` still contained `rustls-webpki 0.103.10`, which is affected by:
- `RUSTSEC-2026-0098`
- `RUSTSEC-2026-0099`

Updating to `0.103.12` resolves the advisory failure with the smallest possible change.

### DIFF
--- a/.cargo/cooldown-allowlist.toml
+++ b/.cargo/cooldown-allowlist.toml
@@ -1,3 +1,3 @@
 [[allow.exact]]
 crate = "rustls-webpki"
-version = "0.103.10"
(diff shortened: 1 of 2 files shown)


nebasuke:
Feel free to add an exception to  `rustls-webpki@0.103.12` for cooldown for a security fix.

nebasuke (approved):

PR #336, opened 2026-04-16
fix: update rand to 0.9.3 for dependabot alert #8

hedgar2017:
Dependabot alert `#8` tracks `GHSA-cq8v-f236-94qc` for `rand`.

In this workspace, the runtime `rand 0.9.x` path comes from `secp256k1` and `twox-hash`, and both accept the patched `0.9.3` release. This change updates that vulnerable runtime path with the smallest possible dependency change.

[The alert](https://github.com/NomicFoundation/solx/security/dependabot/8) is low-priority and not critical for this repo. Can be merged after the cargo cooldown.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,7 +129,7 @@ dependencies = [
  "keccak-asm",
  "paste",
  "proptest",
- "rand 0.9.2",
+ "rand 0.9.3",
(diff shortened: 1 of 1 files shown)


nebasuke (approved):
Thanks, should be able to merge this now

PR #340, opened 2026-04-17
fix(mlir): use ptr_type for sol.alloca alloc_type attribute

hedgar2017:
The `alloc_type` attribute on `sol.alloca` was set to `element_type`, but the TableGen custom builder uses the same type for both `alloc_type` and the result (`Sol_PtrType`). This sets it to `ptr_type` to match the ODS definition.


abinavpp (approved):
thank you!

PR #345, opened 2026-04-20
feat(slang): add assert built-in

hedgar2017:
Wire up `assert(condition)` as a built-in function dispatch in `CallEmitter`.

Previously, `assert()` calls fell through to `resolve_function` and failed with "undefined function: assert".

Newly passing tests:

- `tests/solidity/simple/fuzzed/boolean_assert.sol`
- `tests/solidity/simple/fuzzed/boolean_assert1.sol`


abinavpp (approved):
Thank you!

PR #354, opened 2026-04-21
feat(slang): use slang's `compute_canonical_signature()` for method identifiers

hedgar2017:
Replace manual canonical signature reconstruction from ABI entry inputs with slang's new `compute_canonical_signature()` method (NomicFoundation/slang#1712). Removes the TODO and the manual parameter type extraction.

Depends on NomicFoundation/slang#1712.


abinavpp (approved):
thank you!

PR #356, opened 2026-04-21
refactor(slang): replace `mlir_function_name` with `compute_canonical_signature()`

hedgar2017:
Delete `mlir_function_name`, `type_name_text`, and `elementary_type_text` (~60 lines). Use slang's `compute_canonical_signature()` directly at all call sites.

**Blocked on upstream:** slang's `compute_canonical_signature()` currently returns `None` for constructors, fallback, and receive functions (they have no name in the AST). NomicFoundation/slang#1712 needs to be updated to derive the name from the function kind for these cases. Until then, the `.expect()` calls will panic on those function kinds.

Depends on NomicFoundation/slang#1712, in particular https://github.com/NomicFoundation/slang/pull/1712#discussion_r3116356564


hedgar2017:
Superseded by #549.

PR #359, opened 2026-04-21
fix(slang): use arith.constant for i1 booleans

hedgar2017:
Boolean (`i1`) is signless in MLIR, so it uses `arith.constant` instead of `sol.constant` (which is for signed/unsigned int types). This matches solc's `bExt.genBool()` in `Util.h:69-74` which emits `arith::ConstantOp`.

Adds `emit_arith_constant_bool` to the builder and replaces 5 call sites that were incorrectly using `sol.constant` for `i1` values:
- `TrueKeyword` / `FalseKeyword` literal emission
- `&&` / `||` short-circuit default values
- For-loop implicit `true` condition (infinite loop)


abinavpp (approved):
Thank you!

PR #360, opened 2026-04-21
ci(cache): re-add ccache as artifact-cache fallback

nebasuke:
## Summary

- I've upped the solx cache size to 150GB.
- Re-adds ccache to the `build-llvm` and `build-solc` composite actions as a **fallback behind** the SHA-keyed artifact cache (removed in #246). ccache only runs on artifact-cache miss; the warm-cache fast path (~30 s restore) is unchanged.
- Adds ccache-touch steps to each `cache-warmup.yaml` job so the daily cron extends the ccache LRU even when the artifact cache hits, preventing GHA's 7-day eviction from emptying ccache on quiet periods.
- Adds a `free-disk-space` composite action that reclaims ~24 GB of pre-installed host tooling (dotnet, Android SDK, Haskell, CodeQL) via `container.volumes:` bind-mounts, so hosted `ubuntu-24.04` runners have enough room for cold LLVM builds + ccache on the same disk pool.
- Tunes `LLVM_PARALLEL_LINK_JOBS` per platform — 2 on Linux/Windows and macOS x86 (14–16 GB RAM), 1 on macOS ARM64 — because link memory pressure becomes the bottleneck once compile-phase is served from ccache at 99%+ hit rates. The hosted `macos-15` ARM runner has 3 vCPU / 7 GB RAM; two parallel RelWithDebInfo links at 2–4 GB RSS each push it into swap.
- `max-size: "10G"` per ccache scope (matches pre-#246 value).

## Observed runtimes

Three reference runs on this PR branch show the progression:

- **Cold**: run [#24731422129](https://github.com/NomicFoundation/solx/actions/runs/24731422129) at `fb9f415` — first run, ccache empty, pre-free-disk-space; cargo-checks fails on disk exhaustion during `lld` link of `llvm-opt-fuzzer`.
- **Warm ccache, link-jobs=2**: run [#24745368771](https://github.com/NomicFoundation/solx/actions/runs/24745368771) at `13d56d5` — LLVM ccache populated by the cold run, free-disk-space wired, but parallel links still causing swap on macOS ARM64.
- **Warm ccache + link-jobs tuning**: run [#24751701680](https://github.com/NomicFoundation/solx/actions/runs/24751701680) at `6f85d0a` — `LLVM_PARALLEL_LINK_JOBS=1` on macOS ARM64, 100% ccache hit rate across the board.

### Whole-job wall-clock

| Platform | Cold | Warm ccache, link-jobs=2 | Warm ccache + link-jobs tuned |
|---|---|---|---|
| cargo-checks | 2h 17min, **failed on disk** | 18.5 min | **16.5 min** |
| Linux x86 gnu | 3h 00min | 32.5 min | **15.5 min** |
| Linux ARM64 gnu | 1h 44min | 25 min | **14.5 min** |
| macOS x86 | 3h 28min | 1h 53min | **45 min** |
| macOS ARM64 | 4h 44min | 2h 23min | **21 min** |
| Windows | 4h 31min | 1h 46min | **1h 19min** |

### Build LLVM step only

Isolates the LLVM-building cost from checkout / solc build / tests / etc. Shows that the dramatic macOS ARM64 win is mechanism-specific:

| Platform | Warm ccache, link-jobs=2 | Warm ccache + link-jobs tuned | Speedup |
|---|---|---|---|
| cargo-checks | 7:07 | 4:20 | 1.6× |
| Linux x86 gnu | 7:00 | 3:59 | 1.8× |
| Linux ARM64 gnu | 6:11 | 4:22 | 1.4× |
| macOS x86 | 17:40 | 10:23 | 1.7× |
| **macOS ARM64** | **1h 57min** | **5:54** | **20×** |
| Windows | 46:35 | 42:16 | 1.1× |

Two effects stacking:

1. **ccache hit rate crept from 99.35% → 100%** (confirmed on both macOS legs this run). The previous run had 25 misses out of 3839 cacheable calls; those got saved, so this run hits them. Accounts for the modest 1.4–1.8× speedup on every non-ARM64 leg — roughly 25 cold compiles × ~10–15 s each.
2. **`LLVM_PARALLEL_LINK_JOBS=1` eliminated swap thrashing on macOS ARM64**, which is the 20× speedup. 25 additional cache hits don't explain ~1h 50min of saved wall time; that was the linker stalled on page faults in a 7 GB runner trying to run two ~3 GB RSS link jobs concurrently. Windows' 1.1× result is the control group: 16 GB RAM, no swap pressure, no link-jobs change, speedup limited to cache warming.

ccache itself handles compile, not link. The ~200 link steps per build still run every time; that's the floor. On most platforms link time is tolerable; on macOS ARM64 specifically the interaction between link count, link RSS, and runner RAM was catastrophic before this PR and is now fine.

### Building solc step

solc's ccache key changed in `35949d3` (adding `-{cmake-build-type}-end` per review feedback), which invalidated run 1's saves — run 2 re-populated with the new key format, and run 3 is the first run that actually restores a warm solc ccache at 100% hit rate.

| Platform | Cold (run 1, fb9f415) | Warm ccache (run 3, 6f85d0a) | Speedup |
|---|---|---|---|
| Linux x86 gnu | 19:11 | **1:32** | 12.5× |
| Linux ARM64 gnu | 10:26 | **1:25** | 7.4× |
| macOS x86 | 25:32 | **7:12** | 3.5× |
| macOS ARM64 | 14:07 | **2:31** | 5.6× |
| Windows | 35:50 | **16:03** | 2.2× |

Observations:

- **Linux warm solc is the clean case** — under 2 min thanks to a small link graph (just `solc` and `solc-tests` binaries, vs LLVM's ~200 tool-executables). Almost all of the wall-clock is the link-floor we can't eliminate.
- **Windows warm solc is still 16 min.** Same link-bound story as LLVM: `lld-link` is inherently slow per-executable on Windows and solc + boost together produce enough binaries for that to dominate. No swap here (16 GB RAM, not memory-pressured), just slow links.
- **macOS ARM64 solc doesn't need the link-jobs treatment.** Solc produces few enough binaries that even link-jobs=2 doesn't exceed the 7 GB RAM.

## Fixes and tuning during review

- **Cache-key terminator** (`-end`): `llvm-…-mlir` was a prefix of `llvm-…-mlir-coverage-no-assertions`, so ccache-action's `restore-keys` prefix match was cross-restoring between dev and coverage configs on Linux x64. All LLVM and solc keys now end with an `-end` marker.
- **Touch-step path mismatch**: `cache-warmup.yaml`'s Touch steps used `${{ runner.temp }}/ccache-touch-{llvm,solc}` but ccache-action saves with `path = CCACHE_DIR = ${{ runner.temp }}/ccache-{llvm,solc}`. GHA includes `path` in the cache version hash, so the Touch was a silent no-op. Fixed to real paths.
- **solc key missing `cmake-build-type`**: `warm-solc` (RelWithDebInfo) and `warm-llvm-integration` (Release) were colliding on one key. Key now includes build-type.
- **`Show ccache stats`** → `continue-on-error: true` so a missing-ccache-binary failure can't mask the real build failure.
- **`apt update` handling** — went through a couple of iterations: first bare `apt update`, then `sudo apt-get update -qq` (per review suggestion for portability), then back to plain `apt-get update -qq` once a test run (#24751389733) confirmed the `solx-ci-runner` container image doesn't ship `sudo` at all. Kept the add-then-remove as separate commits so the iteration is visible in history.
- **YAML anchor reality check**: GHA composite-action manifests reject YAML anchors (`ActionManifestManagerLegacy` explicit refusal), so the four duplicated `CCACHE_*` env vars across steps in `build-llvm/action.yml` and `build-solc/action.yml` stay inlined with a sync comment. Workflow files *do* accept anchors, which is used to dedupe the five-entry `container.volumes:` list across the six affected jobs within each of `test.yaml` and `cache-warmup.yaml`.
- **`find -mindepth 1 -delete`** instead of `rm -rf` on the bind-mounted host paths: `rm -rf` on a bind-mount directory fails with EBUSY on the mount point itself (disk reclaim still happens, but log noise and a masked non-zero exit). `find -delete` clears contents, leaves the mount intact, and preserves a meaningful exit code.
- **`xargs -P` for macOS Xcode removal**: the pre-installed runner has ~16 Xcode versions. Sequential `rm -rf` was ~8 min; parallel fan-out cuts it well under two.
- **Windows LLVM tool-disable experiment, landed and reverted**: commit `3106d1e` added `LLVM_BUILD_TOOLS=Off` + `LLVM_INCLUDE_TOOLS=Off` on Windows to skip ~200 unused tool-binary links. Build LLVM step dropped from 42 min → 9:44 (4.3×) on the validation run. Reverted in `90df14b` after `Run tests` failed with `llvm-sys` unable to find `llvm-config` at Rust crate-build time — `llvm-config` is itself a tool binary and got disabled along with the rest. Follow-up to retry with a surgical "build only `llvm-config`" approach tracked in #364.

## Prior art (why this shape)

- #217 — original ccache. Revealed cargo caches evicting LLVM ccache from the 10 GB budget; fixed by removing cargo caches from that path.
- #226 — sccache is broken on MSYS2 with cmake+ninja+clang ([mozilla/sccache#1097](https://github.com/mozilla/sccache/issues/1097), open since 2022). Windows uses ccache.
- #246 — removed ccache entirely in favour of SHA-keyed artifact caching on the (now-outdated) assumption that submodule bumps would be rare. This PR restores only the ccache steps; the 5→4 LLVM config consolidation from #246 stays.

## Cache calculations

Current observed active usage (`gh api repos/NomicFoundation/solx/actions/cache/usage`):

| Entry | Per entry | Copies¹ | Subtotal |
|---|---|---|---|
| `v1-llvm-Windows-X64-RelWithDebInfo-mlir-…` | 9.09 GB | 4 | 36.4 GB |
| `v1-llvm-macOS-X64-RelWithDebInfo-mlir-…` | 2.39 GB | 2 | 4.8 GB |
| `v1-llvm-macOS-ARM64-RelWithDebInfo-mlir-…` | 2.28 GB | 1 | 2.3 GB |
| `v1-llvm-Linux-X64-RelWithDebInfo-mlir-…` | 1.86 GB | 2 | 3.7 GB |
| `v1-llvm-Linux-ARM64-RelWithDebInfo-mlir-…` | 1.82 GB | 2 | 3.6 GB |
| `build-and-test-v2-*` (rust-cache, 3 OSes) | ~1.5 GB | 3 | ~4.7 GB |
| `v1-solc-*` (5 platforms) | ~0.35 GB | 5 | 1.8 GB |
| misc | small | — | ~0.05 GB |
| **Current total** | | | **~59.7 GB / 28 entries** |

¹ GHA scopes caches per-ref: each merge queue branch (`gh-readonly-queue/main/pr-###-…`) creates its own cache copy.

### Projected ccache additions

With `max-size: 10G` per scope, actual on-disk ccache per variant tends to sit at ~2–3 GB for LLVM, ~1 GB for solc. GHA stores the zstd-compressed tarball, typically 40–60 % of on-disk size.

| New ccache entry | On-disk (typical) | Count | Compressed subtotal |
|---|---|---|---|
| LLVM dev (`llvm-<OS>-<arch>-RelWithDebInfo-mlir`), 5 platforms | ~3 GB | 5 | ~8 GB |
| LLVM sanitizer (Linux x86 only) | ~3 GB | 1 | ~1.5 GB |
| LLVM coverage (Linux x86 only) | ~3 GB | 1 | ~1.5 GB |
| LLVM integration / Release (Linux x86 only) | ~3 GB | 1 | ~1.5 GB |
| solc (`solc-<OS>-<arch>-<build-type>`), 5 platforms | ~1 GB | 5 | ~3 GB |
| **New subtotal** | | 13 | **~15–20 GB** |

### Worst-case upper bound

If every ccache scope fills its 10G cap (unlikely, but the theoretical max):

| | On-disk | Compressed in GHA |
|---|---|---|
| 13 scopes × 10G | 130 GB | ~50–80 GB |

### Projected totals

| Scenario | Active cache |
|---|---|
| Current (artifact cache only) | ~60 GB |
| Current + typical ccache | **~75–80 GB** |
| Current + worst-case ccache | **~110–140 GB** |
| Quota (newly raised) | **150 GB** |

Headroom under quota stays comfortable in the typical case and survives the worst case. If observed usage approaches 150 GB after a few weeks, the `max-size` cap is a single-line dial-back.

## Why ccache-touch in cache-warmup?

The composite action gates the ccache steps on `cache-hit != 'true'`. When the daily cron fires on an unchanged submodule SHA, the artifact cache hits → composite skips ccache → the ccache entry's LRU timer isn't refreshed. After 7 quiet days, GHA evicts it. The `Touch ccache` steps use `actions/cache/restore` with `lookup-only: true` — the lookup hits the cache service endpoint, resetting the 7-day access timer, without downloading the ~2–4 GB entry.

## Test plan

- [x] CI passes cold build — confirmed in run #24745368771 once free-disk-space was wired. cargo-checks passes in 16.5 min (was failing on disk with SIGBUS during link).
- [x] LLVM ccache active, populated, and restored across runs. Stats confirm 100% hit rate once warmed.
- [x] solc ccache validated end-to-end on run #24751701680 — 100% hit rate on all five platforms after the key change propagated.
- [x] LLVM wall-clock drops substantially vs cold baseline on every leg. Linux x86: 3h → 15.5 min (12×). Linux ARM64: 1h 44min → 14.5 min (7×). macOS x86: 3h 28min → 45 min (4.6×). macOS ARM64: 4h 44min → 21 min (13×). Windows: 4h 31min → 1h 19min (3.4×).
- [x] `LLVM_PARALLEL_LINK_JOBS=1` on macOS ARM64 validated — Build LLVM step 1h 57min → 5:54 (20×) compared to prior warm-ccache run at link-jobs=2.
- [ ] After merge, watch push-to-main `cache-warmup` run populate ccache entries on main.
- [ ] Monitor `gh api repos/NomicFoundation/solx/actions/cache/usage` over 1–2 weeks — confirm total stays under 150 GB.

## Out of scope / follow-ups

- Publishing prebuilt LLVM binary tarballs from the `solx-llvm` repo (build once there, download by SHA here) — biggest long-term win but requires cross-repo infra.
- Windows LLVM artifact is 9 GB (4× macOS, 5× Linux) — likely PDBs / `.lib` files in `RelWithDebInfo`. Not in scope per prior discussion.
- Capping ninja compile parallelism (`LLVM_PARALLEL_COMPILE_JOBS`) if link-jobs tuning isn't enough on some future platform.
- Reduce Windows LLVM tool count (#364) — attempted in this PR and reverted; worth retrying surgically with `llvm-config` retained. Expected ~4× Windows Build LLVM speedup.

### DIFF
--- a/.github/actions/build-llvm/action.yml
+++ b/.github/actions/build-llvm/action.yml
@@ -89,12 +89,56 @@ runs:
         CARGO_TARGET_DIR: target
       run: ${SFW_PREFIX:-} cargo fetch --locked && ${SFW_PREFIX:-} cargo build --frozen --release --bin solx-dev
 
+    # ccache is layered behind the SHA-keyed artifact cache: it only runs on
+    # artifact-cache miss (i.e. submodule bump) and provides object-level reuse
+    # across neighbouring solx-llvm commits.
+    - name: Define ccache key
+      if: steps.llvm-artifact-cache.outputs.cache-hit != 'true'
+      shell: ${{ runner.os == 'Windows' && 'msys2 {0}' || 'bash' }}
+      id: ccache-key
+      run: |
+        KEY="llvm-${{ runner.os }}-${{ runner.arch }}"
+        [ '${{ inputs.build-type }}' != 'Release' ] && KEY="${KEY}-${{ inputs.build-type }}"
+        [ '${{ inputs.enable-mlir }}' = 'true' ] && KEY="${KEY}-mlir"
+        [ -n '${{ inputs.sanitizer }}' ] && KEY="${KEY}-${{ inputs.sanitizer }}"
+        [ '${{ inputs.enable-coverage }}' = 'true' ] && KEY="${KEY}-coverage"
+        [ '${{ inputs.enable-assertions }}' != 'true' ] && KEY="${KEY}-no-assertions"
+        echo "key=${KEY}" | tee -a "${GITHUB_OUTPUT}"
+
+    # ccache-action requires `apt update` on Linux Docker jobs
+    - name: Prepare ccache installation
+      if: runner.os == 'Linux' && steps.llvm-artifact-cache.outputs.cache-hit != 'true'
+      shell: 'bash'
+      run: apt update
+
+    # sccache is broken with cmake+ninja+clang on MSYS2/Windows
+    # (https://github.com/mozilla/sccache/issues/1097). Using ccache everywhere.
+    - name: Set up compiler cache
+      if: steps.llvm-artifact-cache.outputs.cache-hit != 'true'
+      uses: hendrikmuhs/ccache-action@a1209f81afb8c005c13b4296c32e363431bffea5 # v1.2.14
+      env:
+        CCACHE_DIR: ${{ runner.temp }}/ccache-llvm
+        CCACHE_BASEDIR: ${{ github.workspace }}
+        CCACHE_NOHASHDIR: "true"
+        CCACHE_COMPILERCHECK: "content"
+      with:
+        key: ${{ steps.ccache-key.outputs.key }}
+        restore-keys: ${{ steps.ccache-key.outputs.key }}
+        append-timestamp: true
+        max-size: "4G"
+        verbose: 2
+        save: ${{ github.event_name != 'pull_request' }}
+
     - name: Build LLVM
       if: steps.llvm-artifact-cache.outputs.cache-hit != 'true'
       shell: ${{ runner.os == 'Windows' && 'msys2 {0}' || 'bash' }}
       working-directory: ${{ inputs.working-dir }}
       env:
         VERBOSE: 1
+        CCACHE_DIR: ${{ runner.temp }}/ccache-llvm
+        CCACHE_BASEDIR: ${{ github.workspace }}
+        CCACHE_NOHASHDIR: "true"
+        CCACHE_COMPILERCHECK: "content"
         LIBSTDCPP_SOURCE_PATH: "${{ runner.temp }}/msys64/mingw64/lib/libstdc++.a"
       run: |
         set -x
@@ -111,8 +155,15 @@ runs:
         fi
 
         ./target/release/solx-dev llvm build --build-type ${{ inputs.build-type }} \
-          ${ENABLE_TESTS} ${ENABLE_VALGRIND} ${VALGRIND_OPTIONS} ${ENABLE_ASSERTIONS} ${ENABLE_COVERAGE} ${ENABLE_MLIR} ${SANITIZER} \
+          --ccache-variant=ccache ${ENABLE_TESTS} ${ENABLE_VALGRIND} ${VALGRIND_OPTIONS} ${ENABLE_ASSERTIONS} ${ENABLE_COVERAGE} ${ENABLE_MLIR} ${SANITIZER} \
           --extra-args "-DCMAKE_EXPORT_COMPILE_COMMANDS='Off'" "-DLLVM_PARALLEL_LINK_JOBS='2'" "-DLLVM_OPTIMIZED_TABLEGEN='On'"
+
+    - name: Show ccache stats
+      if: always() && steps.llvm-artifact-cache.outputs.cache-hit != 'true'
+      shell: ${{ runner.os == 'Windows' && 'msys2 {0}' || 'bash' }}
+      env:
+        CCACHE_DIR: ${{ runner.temp }}/ccache-llvm
+      run: ccache --show-stats
 
     - name: Save LLVM artifact cache
       if: steps.llvm-artifact-cache.outputs.cache-hit != 'true' && github.event_name != 'pull_request'
--- a/.github/actions/build-llvm/action.yml
+++ b/.github/actions/build-llvm/action.yml
@@ -111,11 +111,12 @@ runs:
         echo "key=${KEY}" | tee -a "${GITHUB_OUTPUT}"
 
     # ccache-action refreshes the apt package index before installing ccache.
-    # `sudo` works both as root (Docker) and non-root (hosted runners); `-qq` silences routine output.
+    # The solx-ci-runner container doesn't ship sudo (steps run as root), so
+    # we invoke apt-get directly. `-qq` silences routine output.
     - name: Prepare ccache installation
       if: runner.os == 'Linux' && steps.llvm-artifact-cache.outputs.cache-hit != 'true'
       shell: 'bash'
-      run: sudo apt-get update -qq
+      run: apt-get update -qq
 
     # sccache is broken with cmake+ninja+clang on MSYS2/Windows
     # (https://github.com/mozilla/sccache/issues/1097). Using ccache everywhere.
--- a/.github/actions/build-llvm/action.yml
+++ b/.github/actions/build-llvm/action.yml
@@ -125,7 +125,7 @@ runs:
         key: ${{ steps.ccache-key.outputs.key }}
         restore-keys: ${{ steps.ccache-key.outputs.key }}
         append-timestamp: true
-        max-size: "4G"
+        max-size: "10G"
         verbose: 2
         save: ${{ github.event_name != 'pull_request' }}
 
--- a/.github/actions/build-llvm/action.yml
+++ b/.github/actions/build-llvm/action.yml
@@ -120,13 +120,14 @@ runs:
     # sccache is broken with cmake+ninja+clang on MSYS2/Windows
     # (https://github.com/mozilla/sccache/issues/1097). Using ccache everywhere.
     #
-    # The &ccache-llvm-env anchor keeps this block in sync with the Build LLVM
-    # and Show ccache stats steps — composite-action sibling steps don't share
-    # env, so we'd otherwise duplicate four variables three times.
+    # The four CCACHE_* env vars below are duplicated on the Build LLVM and
+    # Show ccache stats steps. YAML anchors/merge-keys would dedupe but GHA's
+    # composite-action manifest loader rejects anchors — keep them in sync by
+    # hand.
     - name: Set up compiler cache
       if: steps.llvm-artifact-cache.outputs.cache-hit != 'true'
       uses: hendrikmuhs/ccache-action@a1209f81afb8c005c13b4296c32e363431bffea5 # v1.2.14
-      env: &ccache-llvm-env
+      env:
         CCACHE_DIR: ${{ runner.temp }}/ccache-llvm
         CCACHE_BASEDIR: ${{ github.workspace }}
         CCACHE_NOHASHDIR: "true"
@@ -144,7 +145,10 @@ runs:
       shell: ${{ runner.os == 'Windows' && 'msys2 {0}' || 'bash' }}
       working-directory: ${{ inputs.working-dir }}
       env:
-        <<: *ccache-llvm-env
+        CCACHE_DIR: ${{ runner.temp }}/ccache-llvm
+        CCACHE_BASEDIR: ${{ github.workspace }}
+        CCACHE_NOHASHDIR: "true"
+        CCACHE_COMPILERCHECK: "content"
         VERBOSE: 1
         LIBSTDCPP_SOURCE_PATH: "${{ runner.temp }}/msys64/mingw64/lib/libstdc++.a"
       run: |
@@ -171,7 +175,11 @@ runs:
       if: always() && steps.llvm-artifact-cache.outputs.cache-hit != 'true'
       continue-on-error: true
       shell: ${{ runner.os == 'Windows' && 'msys2 {0}' || 'bash' }}
-      env: *ccache-llvm-env
+      env:
+        CCACHE_DIR: ${{ runner.temp }}/ccache-llvm
+        CCACHE_BASEDIR: ${{ github.workspace }}
+        CCACHE_NOHASHDIR: "true"
+        CCACHE_COMPILERCHECK: "content"
       run: ccache --show-stats
 
     - name: Save LLVM artifact cache
--- a/.github/actions/build-llvm/action.yml
+++ b/.github/actions/build-llvm/action.yml
@@ -103,6 +103,9 @@ runs:
         [ -n '${{ inputs.sanitizer }}' ] && KEY="${KEY}-${{ inputs.sanitizer }}"
         [ '${{ inputs.enable-coverage }}' = 'true' ] && KEY="${KEY}-coverage"
         [ '${{ inputs.enable-assertions }}' != 'true' ] && KEY="${KEY}-no-assertions"
+        # Terminator: prevents shorter-variant keys from prefix-matching longer ones
+        # via ccache-action's restore-keys (e.g. `...-mlir` matching `...-mlir-coverage-no-assertions`).
+        KEY="${KEY}-end"
         echo "key=${KEY}" | tee -a "${GITHUB_OUTPUT}"
 
     # ccache-action requires `apt update` on Linux Docker jobs
--- a/.github/actions/build-llvm/action.yml
+++ b/.github/actions/build-llvm/action.yml
@@ -177,9 +177,28 @@ runs:
           LINK_JOBS=1
         fi
 
+        # Windows disables LLVM's own tool binaries (opt, llc, llvm-objdump, ...)
+        # because `lld-link` is slow per-executable and those ~200 tool links
+        # dominate Windows wall-clock even at 100% ccache hits on compiles. solx
+        # consumes LLVM via inkwell FFI and never invokes a tool from
+        # target-llvm/target-final/bin/; distro/Xcode packages provide
+        # llvm-cov / llvm-symbolizer / etc. Scoped Windows-only so non-Windows
+        # runs stay identical pending independent validation on those platforms.
+        EXTRA_ARGS=(
+          "-DCMAKE_EXPORT_COMPILE_COMMANDS='Off'"
+          "-DLLVM_PARALLEL_LINK_JOBS='${LINK_JOBS}'"
+          "-DLLVM_OPTIMIZED_TABLEGEN='On'"
+        )
+        if [ "${{ runner.os }}" = "Windows" ]; then
+          EXTRA_ARGS+=(
+            "-DLLVM_BUILD_TOOLS='Off'"
+            "-DLLVM_INCLUDE_TOOLS='Off'"
+          )
+        fi
+
         ./target/release/solx-dev llvm build --build-type ${{ inputs.build-type }} \
           --ccache-variant=ccache ${ENABLE_TESTS} ${ENABLE_VALGRIND} ${VALGRIND_OPTIONS} ${ENABLE_ASSERTIONS} ${ENABLE_COVERAGE} ${ENABLE_MLIR} ${SANITIZER} \
-          --extra-args "-DCMAKE_EXPORT_COMPILE_COMMANDS='Off'" "-DLLVM_PARALLEL_LINK_JOBS='${LINK_JOBS}'" "-DLLVM_OPTIMIZED_TABLEGEN='On'"
+          --extra-args "${EXTRA_ARGS[@]}"
 
     # `continue-on-error` keeps a failing ccache install (→ ccache not on PATH)
     # from masking the real Build LLVM failure in the step summary.
--- a/.github/actions/build-llvm/action.yml
+++ b/.github/actions/build-llvm/action.yml
@@ -110,18 +110,23 @@ runs:
         KEY="${KEY}-end"
         echo "key=${KEY}" | tee -a "${GITHUB_OUTPUT}"
 
-    # ccache-action requires `apt update` on Linux Docker jobs
+    # ccache-action refreshes the apt package index before installing ccache.
+    # `sudo` works both as root (Docker) and non-root (hosted runners); `-qq` silences routine output.
     - name: Prepare ccache installation
       if: runner.os == 'Linux' && steps.llvm-artifact-cache.outputs.cache-hit != 'true'
       shell: 'bash'
-      run: apt update
+      run: sudo apt-get update -qq
 
     # sccache is broken with cmake+ninja+clang on MSYS2/Windows
     # (https://github.com/mozilla/sccache/issues/1097). Using ccache everywhere.
+    #
+    # The &ccache-llvm-env anchor keeps this block in sync with the Build LLVM
+    # and Show ccache stats steps — composite-action sibling steps don't share
+    # env, so we'd otherwise duplicate four variables three times.
     - name: Set up compiler cache
       if: steps.llvm-artifact-cache.outputs.cache-hit != 'true'
       uses: hendrikmuhs/ccache-action@a1209f81afb8c005c13b4296c32e363431bffea5 # v1.2.14
-      env:
+      env: &ccache-llvm-env
         CCACHE_DIR: ${{ runner.temp }}/ccache-llvm
         CCACHE_BASEDIR: ${{ github.workspace }}
         CCACHE_NOHASHDIR: "true"
@@ -139,11 +144,8 @@ runs:
       shell: ${{ runner.os == 'Windows' && 'msys2 {0}' || 'bash' }}
       working-directory: ${{ inputs.working-dir }}
       env:
+        <<: *ccache-llvm-env
         VERBOSE: 1
-        CCACHE_DIR: ${{ runner.temp }}/ccache-llvm
-        CCACHE_BASEDIR: ${{ github.workspace }}
-        CCACHE_NOHASHDIR: "true"
-        CCACHE_COMPILERCHECK: "content"
         LIBSTDCPP_SOURCE_PATH: "${{ runner.temp }}/msys64/mingw64/lib/libstdc++.a"
       run: |
         set -x
@@ -163,11 +165,13 @@ runs:
           --ccache-variant=ccache ${ENABLE_TESTS} ${ENABLE_VALGRIND} ${VALGRIND_OPTIONS} ${ENABLE_ASSERTIONS} ${ENABLE_COVERAGE} ${ENABLE_MLIR} ${SANITIZER} \
           --extra-args "-DCMAKE_EXPORT_COMPILE_COMMANDS='Off'" "-DLLVM_PARALLEL_LINK_JOBS='2'" "-DLLVM_OPTIMIZED_TABLEGEN='On'"
 
+    # `continue-on-error` keeps a failing ccache install (→ ccache not on PATH)
+    # from masking the real Build LLVM failure in the step summary.
     - name: Show ccache stats
       if: always() && steps.llvm-artifact-cache.outputs.cache-hit != 'true'
+      continue-on-error: true
       shell: ${{ runner.os == 'Windows' && 'msys2 {0}' || 'bash' }}
-      env:
-        CCACHE_DIR: ${{ runner.temp }}/ccache-llvm
+      env: *ccache-llvm-env
       run: ccache --show-stats
 
     - name: Save LLVM artifact cache
--- a/.github/actions/build-llvm/action.yml
+++ b/.github/actions/build-llvm/action.yml
@@ -119,10 +119,8 @@ runs:
     # sccache is broken with cmake+ninja+clang on MSYS2/Windows
     # (https://github.com/mozilla/sccache/issues/1097). Using ccache everywhere.
     #
-    # The four CCACHE_* env vars below are duplicated on the Build LLVM and
-    # Show ccache stats steps. YAML anchors/merge-keys would dedupe but GHA's
-    # composite-action manifest loader rejects anchors — keep them in sync by
-    # hand.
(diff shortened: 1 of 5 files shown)


.github/actions/build-llvm/action.yml
```diff
@@ -67,8 +67,10 @@ runs:
         KEY="${KEY}-${SHA}"
         echo "key=${KEY}" | tee -a "${GITHUB_OUTPUT}"
 
+    # TEMP(ci:ccache-test): forced-miss so the ccache path always runs. REVERT BEFORE MERGE.
     - name: Restore LLVM artifact cache
       id: llvm-artifact-cache
+      if: false
```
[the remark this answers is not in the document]
nebasuke: Temp commit: dismissed

.github/actions/build-llvm/action.yml
```diff
@@ -89,12 +91,59 @@ runs:
...
+        CCACHE_BASEDIR: ${{ github.workspace }}
+        CCACHE_NOHASHDIR: "true"
+        CCACHE_COMPILERCHECK: "content"
+      with:
+        key: ${{ steps.ccache-key.outputs.key }}
+        restore-keys: ${{ steps.ccache-key.outputs.key }}
+        append-timestamp: true
+        max-size: "10G"
+        verbose: 2
+        save: true # TEMP(ci:ccache-test): save on PR events too. REVERT BEFORE MERGE to `${{ github.event_name != 'pull_request' }}`.
```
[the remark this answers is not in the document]
nebasuke: Temp commit: dismissed

.github/actions/build-solc/action.yml
```diff
@@ -33,8 +33,10 @@ runs:
         KEY="${KEY}-boost${{ inputs.boost-version }}-${SHA}-script${SCRIPT_HASH:0:8}"
         echo "key=${KEY}" | tee -a "${GITHUB_OUTPUT}"
 
+    # TEMP(ci:ccache-test): forced-miss so the ccache path always runs. REVERT BEFORE MERGE.
     - name: Restore solc artifact cache
       id: solc-artifact-cache
+      if: false
```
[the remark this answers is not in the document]
nebasuke: Temp commit: dismissed

.github/actions/build-solc/action.yml
```diff
@@ -49,19 +51,57 @@ runs:
...
+        CCACHE_BASEDIR: ${{ github.workspace }}
+        CCACHE_NOHASHDIR: "true"
+        CCACHE_COMPILERCHECK: "content"
+      with:
+        key: solc-${{ runner.os }}-${{ runner.arch }}
+        restore-keys: solc-${{ runner.os }}-${{ runner.arch }}
+        append-timestamp: true
+        max-size: "10G"
+        verbose: 2
+        save: true # TEMP(ci:ccache-test): save on PR events too. REVERT BEFORE MERGE to `${{ github.event_name != 'pull_request' }}`.
```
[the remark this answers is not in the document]
nebasuke: Temp commit: dismissed

.github/workflows/cache-warmup.yaml
```diff
@@ -73,6 +73,23 @@ jobs:
...
+      # so the nightly cron keeps ccache alive past GHA's 7-day eviction
+      # window. Primary key is unique → falls through to the restore-keys
+      # prefix match against the ccache-action's timestamped key.
+      # lookup-only: true skips the download; the cache service lookup alone
+      # resets the access timer.
+      - name: Touch LLVM ccache
+        if: always()
+        uses: actions/cache/restore@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
+        with:
+          path: ${{ runner.temp }}/ccache-touch-llvm
```
hedgar2017: All Touch steps use `ccache-touch-llvm` / `ccache-touch-solc` as the `path`, but `hendrikmuhs/ccache-action` saves caches with `path` = `CCACHE_DIR` = `${{ runner.temp }}/ccache-llvm` / `ccache-solc` (confirmed by reading [`common.ts:cacheDir()`](https://github.com/hendrikmuhs/ccache-action/blob/a1209f81afb8c005c13b4296c32e363431bffea5/src/common.ts#L10-L18) — it reads `process.env.CCACHE_DIR`).

GHA includes `path` in the cache version hash, and `restore-keys` prefix matching also checks version. Different path → different version → no match.

**The Touch steps will never hit the real ccache entries, making the LRU refresh a silent no-op.** Fix: use `${{ runner.temp }}/ccache-llvm` and `${{ runner.temp }}/ccache-solc` as paths (matching the ccache-action's `CCACHE_DIR`).

.github/actions/build-solc/action.yml
```diff
@@ -49,19 +51,57 @@ runs:
...
+    - name: Set up compiler cache for solc
+      if: steps.solc-artifact-cache.outputs.cache-hit != 'true'
+      uses: hendrikmuhs/ccache-action@a1209f81afb8c005c13b4296c32e363431bffea5 # v1.2.14
+      env:
+        CCACHE_DIR: ${{ runner.temp }}/ccache-solc
+        CCACHE_BASEDIR: ${{ github.workspace }}
+        CCACHE_NOHASHDIR: "true"
+        CCACHE_COMPILERCHECK: "content"
+      with:
+        key: solc-${{ runner.os }}-${{ runner.arch }}
```
hedgar2017: Solc ccache key is `solc-{os}-{arch}` with no `cmake-build-type` component. `warm-solc` uses `RelWithDebInfo`, `warm-llvm-integration` uses `Release` — different optimization flags produce different object files. ccache's own compiler-flag hashing prevents incorrect output, but the two configs share one cache bucket, reducing effective hit rates (each evicts the other's objects).

The LLVM key already includes build-type; suggest mirroring here:
```suggestion
        key: solc-${{ runner.os }}-${{ runner.arch }}-${{ inputs.cmake-build-type }}
        restore-keys: solc-${{ runner.os }}-${{ runner.arch }}-${{ inputs.cmake-build-type }}
```

.github/actions/build-llvm/action.yml
```diff
@@ -111,9 +160,16 @@ runs:
         fi
 
         ./target/release/solx-dev llvm build --build-type ${{ inputs.build-type }} \
-          ${ENABLE_TESTS} ${ENABLE_VALGRIND} ${VALGRIND_OPTIONS} ${ENABLE_ASSERTIONS} ${ENABLE_COVERAGE} ${ENABLE_MLIR} ${SANITIZER} \
+          --ccache-variant=ccache ${ENABLE_TESTS} ${ENABLE_VALGRIND} ${VALGRIND_OPTIONS} ${ENABLE_ASSERTIONS} ${ENABLE_COVERAGE} ${ENABLE_MLIR} ${SANITIZER} \
           --extra-args "-DCMAKE_EXPORT_COMPILE_COMMANDS='Off'" "-DLLVM_PARALLEL_LINK_JOBS='2'" "-DLLVM_OPTIMIZED_TABLEGEN='On'"
 
+    - name: Show ccache stats
```
hedgar2017, line 182: nit: `ccache --show-stats` will fail with "command not found" if the ccache-action installation step itself failed. Since this is a diagnostic step (`if: always()`), adding `continue-on-error: true` would prevent it from masking the original failure.

.github/actions/build-llvm/action.yml
```diff
@@ -89,12 +91,59 @@ runs:
...
+        verbose: 2
+        save: true # TEMP(ci:ccache-test): save on PR events too. REVERT BEFORE MERGE to `${{ github.event_name != 'pull_request' }}`.
+
     - name: Build LLVM
       if: steps.llvm-artifact-cache.outputs.cache-hit != 'true'
       shell: ${{ runner.os == 'Windows' && 'msys2 {0}' || 'bash' }}
       working-directory: ${{ inputs.working-dir }}
       env:
         VERBOSE: 1
+        CCACHE_DIR: ${{ runner.temp }}/ccache-llvm
```
hedgar2017, line 145: nit: The four `CCACHE_*` env vars are duplicated identically on the ccache-action step (L125–128) and the Build step (L143–146). Duplication is technically necessary (composite step env doesn't propagate to siblings), but they can silently diverge if a fifth var is added to one and not the other. A YAML anchor (`&ccache-llvm-env` / `*ccache-llvm-env`) would keep them in sync. Same applies to `build-solc/action.yml`.

.github/actions/build-llvm/action.yml
```diff
@@ -89,12 +91,59 @@ runs:
...
+        # Terminator: prevents shorter-variant keys from prefix-matching longer ones
+        # via ccache-action's restore-keys (e.g. `...-mlir` matching `...-mlir-coverage-no-assertions`).
+        KEY="${KEY}-end"
+        echo "key=${KEY}" | tee -a "${GITHUB_OUTPUT}"
+
+    # ccache-action requires `apt update` on Linux Docker jobs
+    - name: Prepare ccache installation
+      if: runner.os == 'Linux' && steps.llvm-artifact-cache.outputs.cache-hit != 'true'
+      shell: 'bash'
+      run: apt update
```
hedgar2017: nit: Bare `apt update` works because Linux jobs run as root in the Docker container. `sudo apt-get update -qq` is more portable (works in both root and non-root environments) and quieter. Same applies to `build-solc/action.yml`.

.github/actions/build-llvm/action.yml
nebasuke: Heads-up: this approach doesn't work in GHA, unfortunately. I tried it in 35949d3 and it failed at action-load time in run 24745149414:

```
/home/runner/work/solx/solx/./.github/actions/build-llvm/action.yml:
  Anchors are not currently supported. Remove the anchor 'ccache-llvm-env'
Error: System.ArgumentException: Unexpected type '' encountered while reading 'action manifest root'.
...
   at GitHub.Runner.Worker.ActionManifestManagerLegacy.Load(IExecutionContext executionContext, String manifestFile)
Error: Failed to load /home/runner/work/solx/solx/./.github/actions/build-llvm/action.yml
```

GHA's `ActionManifestManagerLegacy` explicitly rejects YAML anchors/merge-keys in composite-action manifests — it's a hard refusal, not a parser limitation (PyYAML parses these fine, and the workflow-file parser actually does accept anchors, but the action-manifest loader is separate and stricter). Reverted in e60659a, keeping the four `CCACHE_*` env vars inlined on each step with a comment noting why we can't dedupe.

.github/workflows/cache-warmup.yaml
nebasuke: Addressed in 35949d3 — the six Touch steps now point at `${{ runner.temp }}/ccache-llvm` and `${{ runner.temp }}/ccache-solc` to match ccache-action's `CCACHE_DIR`. Thanks for catching this; would've been a silent bug.

.github/actions/build-solc/action.yml
nebasuke: Addressed in 35949d3 — solc ccache key is now `solc-{os}-{arch}-{cmake-build-type}-end`, and the two solc Touch `restore-keys` in `cache-warmup.yaml` are split accordingly (`...-RelWithDebInfo-end` on `warm-solc`, `...-Release-end` on `warm-llvm-integration`). Also adopted the `-end` terminator from the LLVM-key fix (a239dea) for future-proofing against another prefix-collision class.

.github/actions/build-llvm/action.yml
nebasuke: Addressed in 35949d3 — `continue-on-error: true` on the `Show ccache stats` step in both `build-llvm/action.yml` and `build-solc/action.yml`, so a missing-ccache-binary failure can't mask the real build failure.

.github/actions/build-llvm/action.yml
nebasuke: Addressed in 35949d3 — bare `apt update` → `sudo apt-get update -qq` in both composite actions.

nebasuke: Follow-up: this one turned out to backfire. Test run [#24751389733](https://github.com/NomicFoundation/solx/actions/runs/24751389733) failed on the `Prepare ccache installation` step with:

```
/__w/_temp/0e9283d3-…sh: line 1: sudo: command not found
Process completed with exit code 127
```

The `solx-ci-runner` container image doesn't ship `sudo` at all — steps already run as root inside the container, so `sudo` was both unnecessary and broken on Linux legs. Dropped it on top of the stack (comment updated to reflect the actual container behaviour rather than the generic root/non-root portability framing).

Keeping the original "add sudo" commit (35949d3) and the "drop sudo" follow-up as separate commits rather than rewriting history, so the iteration stays visible in the record.

.github/actions/free-disk-space/action.yml
```diff
@@ -0,0 +1,61 @@
...
+          # delete contents instead so the mount stays intact and a real rm
+          # failure can still surface as a non-zero exit.
+          find "$p" -mindepth 1 -delete
+        }
+        cleanup /mnt/free-disk-space/dotnet  &
+        cleanup /mnt/free-disk-space/android &
+        cleanup /mnt/free-disk-space/ghc     &
+        cleanup /mnt/free-disk-space/ghcup   &
+        cleanup /mnt/free-disk-space/codeql  &
+        wait
```
hedgar2017: Bare `wait` with no args returns 0 regardless of child exit codes (bash manual §3.7.5), so a failed `find -delete` on any bind mount leaves the tree populated while the step reports success — the exact ENOSPC mode this action exists to prevent. The comment at L51 ("a real rm failure can still surface as a non-zero exit") is also incorrect as written; it would be true for synchronous calls but parallel backgrounding loses all exit-code information.

Repro confirming the behavior:

```
$ bash -c 'cleanup() { return 1; }; cleanup A & cleanup B & wait; echo $?'
0
```

Fix — capture each PID and wait individually:

```bash
pids=()
cleanup /mnt/free-disk-space/dotnet  & pids+=($!)
cleanup /mnt/free-disk-space/android & pids+=($!)
cleanup /mnt/free-disk-space/ghc     & pids+=($!)
cleanup /mnt/free-disk-space/ghcup   & pids+=($!)
cleanup /mnt/free-disk-space/codeql  & pids+=($!)
rc=0
for pid in "${pids[@]}"; do
  wait "$pid" || rc=$?
done
# ... df -h / ...
exit "$rc"
```

And revise the L51 comment to describe the new per-PID semantics.

.github/actions/free-disk-space/action.yml
```diff
@@ -0,0 +1,61 @@
...
+        echo "Before:"
+        df -h /
+        # Only rm paths that exist and are non-empty. Defends against a
+        # missing `container.volumes:` silently unlinking an empty mount point
+        # on the host, and caps the blast radius to /mnt/free-disk-space/*.
+        cleanup() {
+          local p="$1"
+          case "$p" in
+            /mnt/free-disk-space/*) ;;
+            *) echo "refusing to clean path outside /mnt/free-disk-space/: $p" >&2; return 1 ;;
```
hedgar2017, line 47: `case "$p" in /mnt/free-disk-space/*)` is a shell glob, not a realpath check. `*` matches `/` and path-traversal components, so `/mnt/free-disk-space/../host-dir` passes the guard and `find` then operates on `/host-dir`.

Repro:

```
$ p="/mnt/free-disk-space/../sibling"
$ case "$p" in /mnt/free-disk-space/*) echo "matched";; *) echo "refused";; esac
matched
```

All callers in the PR pass hardcoded literals, so this is unexploitable today. But the action description explicitly claims the rm "operates only on paths under `/mnt/free-disk-space/`, so a missing or misconfigured mount can't target anything outside that prefix" — that claim is false as written.

Two options:

1. Harden the guard to match the doc:
   ```bash
   real=$(realpath -m -- "$p")
   case "$real" in /mnt/free-disk-space/*) ;; *) echo "refusing..."; return 1 ;; esac
   ```
2. Or drop the "caps the blast radius" framing from the description and treat the `case` as best-effort documentation only.

.github/workflows/cache-warmup.yaml
```diff
@@ -120,18 +151,32 @@ jobs:
           boost-version: '1.83.0'
           working-dir: 'solx-solidity'
 
+      - name: Touch solc ccache
```
hedgar2017, line 163: Tangent to this new step, but scoped to the same `warm-solc` job: this job's `container:` block (above, at L126-127) has `image: ${{ matrix.image || '' }}` but no `volumes: *free-disk-volumes`, and the job has no `Free disk space` step. Every other Linux-container job in this file (`warm-llvm`, `warm-llvm-sanitizer`, `warm-llvm-coverage`, `warm-llvm-integration`) carries both.

This may be intentional — the disk-exhaustion failure that motivated `free-disk-space` was LLVM-specific (cold `lld` linking `llvm-opt-fuzzer`), and solc's build + boost footprint is materially smaller. If so, please add a one-line comment on the container block stating that solc doesn't require the cleanup so the next reader doesn't read it as an oversight.

If not intentional, the fix is symmetric with the other jobs:

```yaml
    container:
      image: ${{ matrix.image || '' }}
      volumes: *free-disk-volumes
    ...
    steps:
      ...
      - name: Free disk space
        if: runner.os == 'Linux' && matrix.image != ''
        uses: ./.github/actions/free-disk-space
```

.github/workflows/test.yaml
```diff
@@ -167,19 +191,28 @@ jobs:
...
+            # per inode, so run them concurrently (one worker per bundle).
+            # Soft-fail: a stray rm error shouldn't sink the whole job — rm's
+            # stderr will pinpoint the bad path above.
+            if [ "${#to_remove[@]}" -gt 0 ]; then
+              if printf '%s\0' "${to_remove[@]}" \
+                  | xargs -0 -n1 -P "${#to_remove[@]}" sudo rm -rf; then
+                removed=${#to_remove[@]}
+              else
+                echo "  warning: one or more Xcode removals failed (see rm stderr above)"
+              fi
```
hedgar2017, line 220: `removed=${#to_remove[@]}` is only assigned inside the `then` branch. If any one `sudo rm -rf` in the `xargs` batch fails, `xargs` exits non-zero → the `else` branch runs → `removed` stays at the earlier `removed=0` init, even though N−1 bundles were successfully removed.

The summary line (`echo "Removed ${removed} inactive Xcode version(s)"`) then prints `Removed 0 inactive Xcode version(s)` after freeing ~30 GB of disk, making a mostly-successful cleanup look completely broken.

Fix — set `removed` unconditionally before the xargs call and reserve the `else` branch for the warning:

```bash
if [ "${#to_remove[@]}" -gt 0 ]; then
  removed=${#to_remove[@]}
  if ! printf '%s\0' "${to_remove[@]}" \
      | xargs -0 -n1 -P "${#to_remove[@]}" sudo rm -rf; then
    echo "  warning: one or more Xcode removals failed (see rm stderr above)"
  fi
fi
```

hedgar2017 (changes requested):
Ran another round! Everything is mostly good but a couple of things worth tightening.

.github/actions/free-disk-space/action.yml
nebasuke: Fixed in cc6357e — per-PID `wait "$pid" || rc=$?` capture, so a failed `find -delete` no longer gets masked.

One deliberate deviation from your suggested fix: we emit a GitHub Actions warning annotation (`::warning::…`) and continue, rather than `exit "$rc"`-ing. This isn't just a log line — it surfaces as a yellow annotation on the job summary and in the PR "Checks" tab, so it stays visible without sinking the run. Partial cleanup is usually still enough headroom for the downstream build, and when it isn't, ENOSPC surfaces at a more specific call site than this action can name.

The message is `::warning::free-disk-space: one or more cleanups failed (exit $rc); subsequent steps may hit ENOSPC`.

Also reworded the inline comment at the `find` call — the prior note about "a real rm failure can still surface as a non-zero exit" no longer matches the warn-and-continue contract. Good catch — would've been a silent failure mode otherwise.

.github/actions/free-disk-space/action.yml
nebasuke: Went with your option 2 in 66f71b0 — softened both the action description and the inline comment to describe the `case` guard as best-effort typo-catching, explicitly "NOT a security boundary (the glob matches `..` traversal)". Kept the guard code as-is since all current callers pass hard-coded literals; hardening with `realpath -m --` didn't feel worth the extra fork when the attack surface is our own workflow YAML.

.github/workflows/cache-warmup.yaml
nebasuke: Went with the symmetric fix in 13ff326 rather than just a comment. solc's current footprint (~3–5 GB) fits container headroom today, but the ~30 s the cleanup costs is cheap insurance against future solc/boost growth hitting ENOSPC mid-warmup — and it eliminates the implicit special case the next reader has to reason about. `warm-solc` now carries `volumes: *free-disk-volumes` + the `Free disk space` step, same pattern as the four LLVM warmup jobs.

.github/workflows/test.yaml
nebasuke: Applied your fix in f247529 — `removed=${#to_remove[@]}` set unconditionally before xargs; failure path now just warns. Added a comment block above it stating that the count is off by the number of failed bundles (usually one in practice), so the log wording matches reality rather than claiming zero after ~30 GB was freed.

hedgar2017 (approved):
Thank you sir!

PR #362, opened 2026-04-21
feat(mlir): add lit tests

abinavpp:
## Summary

- Add lit/FileCheck tests for the Slang MLIR frontend (20 tests)
- Add `--emit-mlir=sol`/`--emit-mlir=llvm` to filter output by dialect
- Fix non-deterministic state_var emission order

## Test plan

- `lit solx-mlir/tests/lit/` - 20/20 pass

### DIFF
--- a/solx-mlir/tests/lit/.gitignore
+++ b/solx-mlir/tests/lit/.gitignore
@@ -1,0 +1,2 @@
+Output/
+.lit_test_times.txt
--- a/solx-mlir/tests/lit/address_cast.sol
+++ b/solx-mlir/tests/lit/address_cast.sol
@@ -1,0 +1,11 @@
+// RUN: solx --emit-mlir %s | FileCheck %s
+
+// CHECK: sol.func @"to_address(uint256)"
+// CHECK:   sol.cast %{{.*}} : ui256 to ui160
+// CHECK:   sol.address_cast %{{.*}} : ui160 to !sol.address
+
+contract C {
+    function to_address(uint256 x) public pure returns (address) {
+        return address(uint160(x));
+    }
+}
--- a/solx-mlir/tests/lit/arithmetic.sol
+++ b/solx-mlir/tests/lit/arithmetic.sol
@@ -1,0 +1,110 @@
+// RUN: solx --emit-mlir %s | FileCheck %s
+
+// CHECK: sol.func @"checked_add(uint256,uint256)"
+// CHECK:   sol.cadd %{{.*}}, %{{.*}} : ui256
+
+// CHECK: sol.func @"checked_sub(uint256,uint256)"
+// CHECK:   sol.csub %{{.*}}, %{{.*}} : ui256
+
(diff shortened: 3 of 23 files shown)


solx-core/src/build/contract/mod.rs
```diff
@@ -92,12 +92,19 @@ impl Contract {
...
         output_selection: &solx_standard_json::InputSelection,
+        #[cfg(feature = "mlir")] mlir_filter: Option<&str>,
     ) -> anyhow::Result<()> {
-        writeln!(
-            std::io::stdout(),
-            "\n======= {} =======",
-            self.name.full_path
-        )?;
+        #[cfg(feature = "mlir")]
+        let skip_banner = mlir_filter.is_some();
```
hedgar2017: AFAIU, FileCheck requires the driver to output exactly one output string, or do we have some control over it on the LLVM side?
Could we move skipping this banner there by any chance?

solx-core/src/build/contract/mod.rs
```diff
@@ -232,7 +239,13 @@ impl Contract {
         ) && let Some(stages) = self.mlir.take()
         {
             for (dialect, mlir_text) in stages.iter() {
-                writeln!(std::io::stdout(), "MLIR Dialect {dialect}:\n{mlir_text}")?;
+                if mlir_filter.is_none_or(|f| f == dialect) {
+                    if mlir_filter.is_some() {
```
hedgar2017: Same as above: could it be moved to FileCheck to avoid too much branching in the production output pipeline?

solx-slang/src/ast/contract/mod.rs
```diff
@@ -64,7 +64,9 @@ impl<'state, 'context> ContractEmitter<'state, 'context> {
         );
 
         // Emit sol.state_var declarations for each storage slot.
-        for slot in storage_layout.values() {
+        let mut slots: Vec<_> = storage_layout.values().copied().collect();
```
hedgar2017: Worth adding a comment above to explain why this sorting is needed.

hedgar2017 (changes requested):
Need to remove the dangerous banner side effect from the MLIR filter. The rest LGTM for now!

When we have more tests, we can think how to cut the pipeline short and do not wait until the final output logic will be executed.

abinavpp:
thanks for the review. i redid this test assertions using claude instead of the lit test updater script. the assertions are more focused and hence we don't need any changes in the ir printing. to run the tests, run: `lit solx-mlir/tests/lit`. let's start adding ir tests for each pr's once we merge this. claude works quite well with generating and updating lit tests, but of course like everything else, needs a manual review.

hedgar2017:
@abinavpp thanks! I will start adding such tests from now on.

That thing with unordered dialects printed to the CLI was a good catch. I'll create a separate PR for it!

hedgar2017 (approved):

PR #370, opened 2026-04-24
feat(slang): emit sol.return after sol.revert

hedgar2017:
The Sol dialect does not mark `sol.revert` with the `IsTerminator` trait, so the enclosing block still needs a structural terminator. The frontend previously treated `sol.revert` as control-flow-terminating and appended an immediate `sol.return` inside `emit_sol_revert`. That produced invalid IR whenever revert appeared mid-flow — `revert(...); some_stmt;` would emit `sol.revert; sol.return; some_stmt`, placing ops after a terminator and tripping the verifier.

`sol.revert` is conceptually closer to a noreturn call than a control-flow op: at the source language level it halts execution, but at the dialect level it is just a regular op without the `IsTerminator` trait. Treating it that way in codegen lets the existing function-epilogue path — already responsible for closing fall-through blocks with `sol.return` — supply the structural terminator, with no special-casing for revert.

### DIFF
--- a/solx-mlir/src/context/builder/mod.rs
+++ b/solx-mlir/src/context/builder/mod.rs
@@ -300,8 +300,6 @@ impl<'context> Builder<'context> {
             .into()
     }
 
-    // ==== Terminators ====
-
     /// Emits a `sol.revert` carrying an optional payload.
     ///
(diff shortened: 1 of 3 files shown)


abinavpp (approved):
thank you!

PR #375, opened 2026-04-27
ci: change runner to solx-linux-amd64-self-hosted

nebasuke:
- Switch the runner from `solx-linux-amd64` (per-run) to `solx-linux-amd64-self-hosted` (persistent) in the three workflows that target it: `integration-tests.yaml`, `coverage.yaml`, `sanitizer.yaml`.
- In `integration-tests.yaml`, share the solc build between the PR checkout and the `temp-solx-main` checkout via symlink when their solc submodule SHAs match — same pattern as the existing LLVM-sharing step.
- Refactor the LLVM share step to use the same output+skip pattern (`steps.share-llvm.outputs.shared`) as the new solc step instead of building inline, so the two share/build pairs are symmetric.
- Clear stale dirs/symlinks (`rm -rf` before `ln -snf`) in both share steps so a previous run's main-build output on the persistent self-hosted workspace can't be reused.

## Runs
Integration tests:
* https://github.com/NomicFoundation/solx/actions/runs/25048377992
* https://github.com/NomicFoundation/solx/actions/runs/25053126423


.github/workflows/integration-tests.yaml
```diff
@@ -82,31 +82,60 @@ jobs:
...
         uses: ./.github/actions/build-solc
         with:
           cmake-build-type: Release
           working-dir: 'solx-solidity'
 
+      - name: Share solc with main checkout
+        id: share-solc
+        continue-on-error: true
+        run: |
+          MAIN_SHA=$(git -C temp-solx-main/solx-solidity rev-parse HEAD)
```
hedgar2017, line 150: `build-solc/action.yml:34` keys its artifact cache on submodule SHA + `find solx-dev/src/solc | sha256sum`. The share predicate here matches only the SHA, so a PR that modifies `solx-dev/src/solc` without bumping the submodule pin will share its own solc binary as the "main" baseline and skew the benchmark silently. Suggest also computing and comparing the script hash here:

```bash
SCRIPT_PR=$(find solx-dev/src/solc -type f -print0 | sort -z | xargs -0 cat | sha256sum | cut -d' ' -f1)
SCRIPT_MAIN=$(find temp-solx-main/solx-dev/src/solc -type f -print0 | sort -z | xargs -0 cat | sha256sum | cut -d' ' -f1)
if [ "$MAIN_SHA" = "$PR_SHA" ] && [ "$SCRIPT_MAIN" = "$SCRIPT_PR" ]; then
  ...
```

.github/workflows/integration-tests.yaml
```diff
@@ -82,31 +82,60 @@ jobs:
...
+            ln -snf "$GITHUB_WORKSPACE/solx-solidity/boost" \
+                    temp-solx-main/solx-solidity/boost
+            echo "shared=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "::notice::solc SHAs differ (main=$MAIN_SHA PR=$PR_SHA), building main separately"
+          fi
+
+      # If share-solc failed entirely (output empty), fall through and build.
       - name: Build solc (main)
+        if: steps.share-solc.outputs.shared != 'true'
```
hedgar2017, line 168: Submodule angle: `temp-solx-main/solx-solidity/build` and `boost` live inside the `solx-solidity` submodule, and `actions/checkout`'s `git clean -ffdx` doesn't recurse into submodules. So even when `Checkout main` succeeds, a symlink left by a prior SHA-match run survives. When SHAs differ this run, `share-solc` writes nothing to `$GITHUB_OUTPUT`, this step fires, and the `actions/cache/restore` inside `build-solc` extracts through the symlink into the PR's freshly-built `solx-solidity/build` and `boost`. A defensive `rm -rf temp-solx-main/solx-solidity/{build,boost}` before this step closes it (or Copilot's "delete `temp-solx-main` before the optional checkout" variant — that covers this and the LLVM-side equivalent in one stroke).

hedgar2017 (changes requested):
@nebasuke — stepping back from the per-line comments for a moment, the three problems flagged here (the two stale-symlink fallback paths, plus the build-script-hash dimension that share-solc misses) all come from the same place: the workflow is reimplementing a SHA-keyed mirror in bash, in parallel to the cache-key logic the `build-llvm` and `build-solc` composite actions already own.

Worth considering folding the mirror in as a `mirror-from` input on those actions instead. The workflow then collapses to two ordinary `uses:` calls per artifact:

```yaml
- name: Build LLVM (PR)
  uses: ./.github/actions/build-llvm
  with: { build-type: Release, enable-assertions: 'false' }

- name: Build LLVM (main)
  uses: ./.github/actions/build-llvm
  continue-on-error: true
  with:
    working-dir: temp-solx-main
    mirror-from: '.'
    build-type: Release
    enable-assertions: 'false'

- name: Build solc (PR)
  uses: ./.github/actions/build-solc

- name: Build solc (main)
  uses: ./.github/actions/build-solc
  continue-on-error: true
  with:
    working-dir: temp-solx-main/solx-solidity
    mirror-from: solx-solidity
```

Inside the action, a single new step gated on `inputs.mirror-from != ''`, falling through to the existing cache-restore + build chain when it can't mirror:

```yaml
- name: Mirror artifact from sibling checkout
  id: mirror
  if: inputs.mirror-from != ''
  shell: bash
  run: |
    SRC="${{ inputs.mirror-from }}/target-llvm/target-final"
    DST="${{ inputs.working-dir }}/target-llvm/target-final"
    SRC_KEY=$(<compute the same cache key, with mirror-from's SHAs>)
    [ "$SRC_KEY" = "${{ steps.llvm-cache.outputs.key }}" ] && [ -d "$SRC" ] || {
      echo "mirrored=false" >> "$GITHUB_OUTPUT"; exit 0
    }
    rm -rf "$DST"
    mkdir -p "$(dirname "$DST")"
    cp -al "$SRC" "$DST"
    echo "mirrored=true" >> "$GITHUB_OUTPUT"
```

Existing build steps' `if:` extends to `... && steps.mirror.outputs.mirrored != 'true'`.

Two things this gets that the surgical fixes won't:

- **Cleanup is unconditional and adjacent to the write** — `rm -rf "$DST"` sits one line above `cp -al`, so it can't drift out of the SHA-match branch the way the current `rm -rf` did. Submodule-internal stale state and Copilot's broader "stale `temp-solx-main`" concern collapse into the same fix.
- **The mirror predicate inherits the action's full cache-key dimensions for free** — boost version, MLIR flag, build-script hash, sanitizer — so the share decision can't drift from the cache decision as keys grow.

Also `cp -al` instead of `ln -snf`. Hardlinks under cmake/cargo's rename-on-write conventions don't have the through-write surface symlinks do, and the consumers here are read-only.

Bigger than a surgical fix on this PR — happy to take it as a follow-up if you'd rather close #375 with the `rm -rf` cleanup and the script-hash check in the share predicate (per the inline comments).

nebasuke:
Thanks — going with the surgical path on this PR. Filed the `mirror-from` refactor as #448 to track separately.

**What landed here (path B):**

- `aa3b16d` wipes `temp-solx-main` wholesale before the optional `Checkout main`. That's your "delete `temp-solx-main` before the optional checkout" variant, and it closes both stale-symlink concerns (LLVM and solc submodule-internal) plus Copilot's broader stale-checkout flag in one step.
- `484673c` + `da1d918` add the build-script-hash dimension to both share predicates, mirroring the cache-key composition in `build-{llvm,solc}/action.yml`.
- `0be720f` extracts the `find | sort -z | xargs -0 cat | sha256sum | cut` pipeline into a named `hash_tree` shell function — it's computed twice per step (PR vs main) and the inline one-liner obscured what was being compared.

**Why not path A on this PR:** the architectural win (mirror predicate inheriting the action's cache-key dimensions for free, hardlinks instead of symlinks, cleanup adjacent to the write) is real and worth doing, but it rewrites the composite actions — bigger blast radius than the three defects this PR is fixing. Easier to review the surgical patch against the inline comments and ship that, then refactor cleanly under #448.

.github/workflows/integration-tests.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
       contents: read
       pull-requests: write
       packages: read
-    runs-on: solx-linux-amd64
+    runs-on: solx-linux-amd64-self-hosted
```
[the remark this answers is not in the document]
nebasuke: Addressed in `aa3b16d` — added a `Reset main checkout dir` step that does `rm -rf temp-solx-main` before the optional `Checkout main`. So the share/build(main) steps always run against a freshly-checked-out main (or against nothing, if the checkout itself fails — share-llvm/share-solc both `continue-on-error` and the `Build {LLVM,solc} (main)` fall-through steps re-check via the `shared != 'true'` gate, which is correct either way).

Same step also closes the symmetric submodule-internal stale-symlink concern hedgar flagged on line 168.

.github/workflows/integration-tests.yaml
nebasuke: Done in `484673c`, then refactored in `0be720f`. The predicate is now:

```bash
hash_tree() {
  find "$1" -type f -print0 | sort -z | xargs -0 cat | sha256sum | cut -d' ' -f1
}
...
SCRIPT_PR=$(hash_tree solx-dev/src/solc)
SCRIPT_MAIN=$(hash_tree temp-solx-main/solx-dev/src/solc)
if [ "$MAIN_SHA" = "$PR_SHA" ] && [ "$SCRIPT_MAIN" = "$SCRIPT_PR" ]; then
```

Mirrors the same composition `build-solc/action.yml` uses for its cache key. The same fix is applied to share-llvm (`da1d918`) for the parallel `solx-dev/src/llvm/*` case.

Followup #448 tracks moving this predicate into the composite action so it inherits the cache key directly instead of mirroring it.

.github/workflows/integration-tests.yaml
nebasuke: Took the variant in parens — wholesale `rm -rf temp-solx-main` before the optional checkout, in `aa3b16d`. That closes this case and the LLVM-side equivalent in one step:

> actions/checkout's `git clean -ffdx` doesn't recurse into submodules, so on the persistent self-hosted workspace a prior run can leave stale solx-solidity/{build,boost} dirs or symlinks (pointing at the PR's freshly-built artifacts) inside temp-solx-main. Subsequent SHA-mismatch runs would extract through those symlinks via `actions/cache/restore`, silently corrupting the PR's build output and skewing benchmarks.
>
> Wholesale `rm -rf temp-solx-main` before the optional checkout closes this and the symmetric stale-checkout concern in one stroke.

So the defensive `rm -rf temp-solx-main/solx-solidity/{build,boost}` inline at this step isn't needed — the directory itself doesn't survive between runs.

hedgar2017 (approved):

PR #379, opened 2026-04-28
ci: rework label-gated suites as advisory checks + fix label concurrency

nebasuke:
## Summary

The four label-gated suites (`coverage`, `integration-tests`, `sanitizer`, `slang-tests`) are **advisory opt-in checks**, not merge gates: add the `ci:*` label to run a suite (e.g. the long integration tests) and read the result on the PR, but they don't block merging or run in the merge queue. This PR unifies them on one simple advisory shape and fixes a label-concurrency bug that could silently cancel an opted-in run.

> Originally scoped as "unify so they can all be marked **required**"; we settled on **advisory** instead — required gating for long opt-in suites caused more flakiness than it prevented. The shared submodule-checkout composite action from the original plan landed separately in #383, so it is **not** part of this PR.

## Changes

- **Unify + advisory** (`6a48dc45`): drop the dummy `label-check` job (gate inlined onto `cooldown-check`); remove the `pr-checks` / `re-actors/alls-green` aggregators and their `allowed-skips`, the `merge_group:` triggers, and the now-unused `checks: write` on sanitizer/slang. Each suite posts its own status; absent label → clearly **skipped** rather than a misleading green-via-skip.
- **Per-label concurrency** (`8d50c89e`): put the triggering label into the concurrency group, so a `labeled` event for an *unrelated* label lands in its own group and can't cancel the run started by the matching label. Fixes the "add two labels → one suite gets cancelled/skipped" flakiness. Other PR events (push/synchronize) share the `main` group and cancel-on-new as before.
- **release dedup** (`2f841114`): `release.yaml` had no concurrency block and gates the dry-run on plain `contains(labels, 'ci:release')`, so every `labeled` event launched a fresh full multi-platform build. Add a per-PR concurrency group; `cancel-in-progress` is gated on `pull_request` so real releases (tag push / `workflow_dispatch`) get a per-ref group and are never cancelled.

## ⚠️ Branch protection — must land together with this PR

Remove `PR Checks (Coverage)`, `PR Checks (Integration)`, `PR Checks (Sanitizer)`, `PR Checks (Slang)` from the **required status checks** list. `sanitizer` and `slang` previously reported in the merge queue via `merge_group`; if they stay required after that trigger is removed, **the merge queue will block on checks that never report.**

## Test plan

- [ ] Workflows parse (validated locally).
- [ ] PR with no `ci:*` label: the four suites show as skipped; nothing blocks merge.
- [ ] Add `ci:integration`, then `ci:coverage`: both real runs complete; neither cancels the other.
- [ ] Add all `ci:*` labels at once: each suite's real run completes.
- [ ] Remove a label: no re-trigger (`unlabeled` is not a trigger).
- [ ] `release.yaml`: several `ci:*` labels at once → one dry-run pipeline, not several.

### DIFF
--- a/.github/actions/checkout-submodules/action.yml
+++ b/.github/actions/checkout-submodules/action.yml
@@ -1,0 +1,20 @@
+name: Checkout submodules
+description: >-
+  Initializes and updates all submodules at depth 1, including ones marked
+  inactive in .gitmodules. The repository must already be checked out before
+  calling this action.
+
+inputs:
+  working-directory:
+    description: Directory in which to run the submodule update
+    required: false
+    default: .
+
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      run: |
+        git config --global --add safe.directory '*'
+        git submodule update --init --force --recursive --depth 1 --checkout
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -3,9 +3,10 @@ name: Code coverage
 on:
   pull_request:
     types: [opened, synchronize, reopened, labeled]
+  merge_group:
 
 concurrency:
-  group: ${{ github.repository_id }}-${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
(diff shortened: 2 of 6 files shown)


nebasuke:
Hard to properly fix and have nice required checks. Will revisit later if this becomes more of an issue.

PR #384, opened 2026-04-29
ci(slang): run MLIR lit tests in slang-tests workflow

hedgar2017:
Runs the MLIR lit suite in the solx Slang Tests workflow so AST→MLIR coherence regressions get caught in CI.

### DIFF
--- a/.github/workflows/slang-tests.yaml
+++ b/.github/workflows/slang-tests.yaml
@@ -88,6 +88,24 @@ jobs:
           build-type: RelWithDebInfo
           enable-assertions: 'true'
           enable-mlir: 'true'
+          enable-tests: 'true'
+
+      # TODO: switch to origin/main once the solx-solidity MLIR-codegen PRs
+      # (#119, #120, #121) merge; until then the chain branch carries the
+      # AddressType / sol.for step / arith-bool fixes the lit tests rely on.
+      # TODO: remove this step and the Build solc step below once the lit
+      # tests stop running against solc.
+      - name: Sync solx-solidity for MLIR codegen
+        shell: bash
+        run: |
+          git -C solx-solidity fetch origin chain-mlir-pr-fixes
+          git -C solx-solidity checkout FETCH_HEAD
+
+      - name: Build solc
+        uses: ./.github/actions/build-solc
+        with:
+          cmake-build-type: RelWithDebInfo
+          enable-mlir: 'true'
 
       # bindgen's embedded clang needs to know it's targeting MinGW,
       # otherwise it can't parse MinGW-specific __attribute__ extensions
@@ -96,6 +114,20 @@ jobs:
         if: runner.os == 'Windows'
         shell: bash
         run: echo "BINDGEN_EXTRA_CLANG_ARGS=--target=x86_64-w64-windows-gnu" >> "${GITHUB_ENV}"
+
+      - name: Preserve lit tools before disk cleanup
+        shell: bash
+        run: |
+          LIT=llvm-lit${{ runner.os == 'Windows' && '.py' || '' }}
+          mkdir -p "${RUNNER_TEMP}/lit-tools"
+          cp -r \
+            target-llvm/target-final/bin/FileCheck \
+            target-llvm/target-final/bin/${LIT} \
+            solx-llvm/llvm/utils/lit \
+            solx-solidity/build/solc/solc \
+            "${RUNNER_TEMP}/lit-tools/"
+          echo "${RUNNER_TEMP}/lit-tools" >> "${GITHUB_PATH}"
+          echo "PYTHONPATH=${RUNNER_TEMP}/lit-tools/lit" >> "${GITHUB_ENV}"
 
       - name: Free disk space (remove build artifacts before slang tests)
         shell: bash
@@ -119,6 +151,14 @@ jobs:
           cargo-test-command: 'test-slang'
           results-xml: 'slang-unit-tests-results.xml'
           check-name: 'Slang Unit Tests Results'
+
+      - name: Run MLIR lit tests
+        shell: bash
+        env:
+          SOLX_LIT_TARGET: ${{ matrix.target || '' }}
+        run: |
+          LIT=llvm-lit${{ runner.os == 'Windows' && '.py' || '' }}
+          ${LIT} -v solx-mlir/tests/lit/
 
   pr-checks:
     name: PR Checks (Slang)
--- a/.github/workflows/slang-tests.yaml
+++ b/.github/workflows/slang-tests.yaml
@@ -95,10 +95,7 @@ jobs:
       - name: Sync solx-solidity for MLIR codegen
         shell: bash
         run: |
-          # TEMP: track fix/mlir-default-threadpool until the DefaultThreadPool
-          # cross-platform fix lands on solx-solidity main. Revert to `main` once
-          # the upstream PR merges.
-          git -C solx-solidity fetch origin fix/mlir-default-threadpool
+          git -C solx-solidity fetch origin main
           git -C solx-solidity checkout FETCH_HEAD
 
       - name: Build solc
--- a/.github/workflows/slang-tests.yaml
+++ b/.github/workflows/slang-tests.yaml
@@ -95,7 +95,10 @@ jobs:
       - name: Sync solx-solidity for MLIR codegen
         shell: bash
         run: |
-          git -C solx-solidity fetch origin main
+          # TEMP: track fix/mlir-default-threadpool until the DefaultThreadPool
+          # cross-platform fix lands on solx-solidity main. Revert to `main` once
+          # the upstream PR merges.
+          git -C solx-solidity fetch origin fix/mlir-default-threadpool
           git -C solx-solidity checkout FETCH_HEAD
 
       - name: Build solc
--- a/.github/workflows/slang-tests.yaml
+++ b/.github/workflows/slang-tests.yaml
@@ -90,6 +90,14 @@ jobs:
           enable-mlir: 'true'
           enable-utils: 'true'
 
+      # TODO: remove the Build solc step below once the lit tests stop
+      # running against solc.
+      - name: Build solc
+        uses: ./.github/actions/build-solc
+        with:
+          cmake-build-type: RelWithDebInfo
+          enable-mlir: 'true'
+
       # bindgen's embedded clang needs to know it's targeting MinGW,
       # otherwise it can't parse MinGW-specific __attribute__ extensions
       # in stdlib.h (rust-lang/rust-bindgen#1760).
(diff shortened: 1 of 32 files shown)


.github/workflows/slang-tests.yaml
```diff
@@ -90,6 +90,7 @@ jobs:
           build-type: RelWithDebInfo
           enable-assertions: 'true'
           enable-mlir: 'true'
+          enable-tests: 'true'
```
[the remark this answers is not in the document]
hedgar2017: Done in 4217161: added `enable-tests` to both the LLVM artifact cache key and the ccache key.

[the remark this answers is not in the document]
hedgar2017: Done in 64f7dae: `enable-tests` is now part of both the LLVM artifact and ccache keys.

.github/workflows/slang-tests.yaml
```diff
@@ -99,6 +100,18 @@ jobs:
         shell: bash
         run: echo "BINDGEN_EXTRA_CLANG_ARGS=--target=x86_64-w64-windows-gnu" >> "${GITHUB_ENV}"
 
+      - name: Preserve lit tools before disk cleanup
+        shell: bash
+        run: |
+          mkdir -p "${RUNNER_TEMP}/lit-tools"
+          cp -r \
+            target-llvm/build-final/bin/FileCheck \
+            target-llvm/build-final/bin/llvm-lit \
```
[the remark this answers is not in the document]
hedgar2017: Done in 64f7dae: lit tools now land in `target-llvm/target-final/bin` via cmake's install rule (`LLVM_INSTALL_UTILS=On` for FileCheck plus a copy step in solx-dev's ninja util for the wrapperless llvm-lit), so they ride along with the existing cache.

.github/workflows/slang-tests.yaml
```diff
@@ -88,6 +90,24 @@ jobs:
...
+
+      # TODO: switch to origin/main once the solx-solidity MLIR-codegen PRs
+      # (#119, #120, #121) merge; until then the chain branch carries the
+      # AddressType / sol.for step / arith-bool fixes the lit tests rely on.
+      # TODO: remove this step and the Build solc step below once the lit
+      # tests stop running against solc.
+      - name: Sync solx-solidity for MLIR codegen
+        shell: bash
+        run: |
+          git -C solx-solidity fetch origin chain-mlir-pr-fixes
```
nebasuke: I think this invalidate solc caching for a bit, but otherwise okay.

.github/workflows/slang-tests.yaml
```diff
@@ -88,6 +88,24 @@ jobs:
           build-type: RelWithDebInfo
           enable-assertions: 'true'
           enable-mlir: 'true'
+          enable-tests: 'true'
```
nebasuke: Why do we need enable-tests: 'true'? It also does LLVM_INCLUDE_TESTS=On and LLVM_BUILD_TESTS=On. 

Could we instead split into a narrower `enable-utils` input that toggles only LLVM_BUILD_UTILS / LLVM_INCLUDE_UTILS / LLVM_INSTALL_UTILS, and keeps LLVM_BUILD_TESTS / LLVM_INCLUDE_TESTS off?

nebasuke (commented):
Few comments on approach, looks fine CI wise (with some slower solc build times until #119 etc land).

.github/workflows/slang-tests.yaml
hedgar2017: The temporary branch or in general? I think it should be fine when it's reverted to main.

.github/workflows/slang-tests.yaml
hedgar2017: I need to take a look. Originally we tried to minimize the number of flags in building tools to facilitate DevEx, but now optimizing CI may be more important.

I think it makes sense, and we can tackle it in a separate update.

.github/workflows/slang-tests.yaml
nebasuke: Yeah should just be some temporary main.

.github/workflows/slang-tests.yaml
nebasuke: Discussed, let's do it in another PR!

nebasuke: Looking at the CI, it probably needs to be addressed now. MacOS x86, build LLVM, cold, for Slang is at 5h:03m.  This is without building solc fully, and running tests. There is a good chance a cold run will not pass CI at that point as there is a 6 hour maximum. We don't want to get in a place where we have to manually disable the test workflows/merge queue to get the fix in. 

I would say in general, it's a quite big regression in CI build times.

hedgar2017: Done in #396 — adds the narrower `enable-utils` input you suggested and makes it the default.

hedgar2017: Rebased.

solx-mlir/tests/lit/scoping.sol
```diff
@@ -1,13 +1,14 @@
...
 // CHECK:   sol.alloca : !sol.ptr<ui256, Stack>
 // CHECK:   sol.alloca : !sol.ptr<ui256, Stack>
 
-// CHECK: sol.func @"default_return()"
+// CHECK: sol.func @{{.*default_return.*}}
 // CHECK:   sol.constant 42
-// CHECK:   %[[ZERO:.*]] = sol.constant 0 : ui256
-// CHECK:   sol.return %[[ZERO]] : ui256
+// CHECK:   sol.constant 0 : ui256
+// CHECK:   sol.return
```
abinavpp: ditto. we need to test that we're returning zero here

solx-mlir/tests/lit/literal_types.sol
```diff
@@ -1,24 +1,24 @@
...
 
-// CHECK: sol.func @"neg_int8()"
-// CHECK:   %[[POS:.*]] = sol.constant 1 : ui8
-// CHECK:   %[[CAST:.*]] = sol.cast %[[POS]] : ui8 to si8
-// CHECK:   %[[ZERO:.*]] = sol.constant 0 : si8
-// CHECK:   sol.sub %[[ZERO]], %[[CAST]] : si8
+// solc folds `return -1;` directly to `sol.constant -1 : si8`; solx emits
+// `constant 1 : ui8` -> `cast` -> `sub 0 - 1` (catalog item 9). Both yield -1.
+// CHECK: sol.func @{{.*neg_int8.*}}
+// CHECK:   sol.return %{{.*}} : si8
```
abinavpp: we need to match the original pattern as it's the part that does the negation. maybe we could use check prefixes to add for both solc and solx versions?

solx-mlir/tests/lit/logical.sol
```diff
@@ -1,33 +1,38 @@
...
-// CHECK:   %[[FALSE:.*]] = arith.constant false
-// CHECK:   sol.store %[[FALSE]], %[[RES:.*]] :
+// Short-circuit &&: alloca a result slot, seed it, conditionally store b in
+// the then-branch, then load and return. solx and solc seed the slot
+// differently (constant-false vs the loaded `a`), so we don't pin the seed.
+// CHECK: sol.func @{{.*logical_and.*}}
+// CHECK:   sol.alloca : !sol.ptr<i1, Stack>
+// CHECK:   sol.alloca : !sol.ptr<i1, Stack>
+// CHECK:   %[[RES:.*]] = sol.alloca : !sol.ptr<i1, Stack>
+// CHECK:   sol.store %{{.*}}, %[[RES]]
```
abinavpp, line 11: ditto

abinavpp (commented):
i think we should always try to match the relevant ir and use check-prefixes for now if they're different from solc. wdyt?

hedgar2017:
> i think we should always try to match the relevant ir and use check-prefixes for now if they're different from solc. wdyt?

Sounds good! Let me tackle it.

solx-mlir/tests/lit/scoping.sol
hedgar2017: Done in 47b6f7a — the test now pins both shapes via `CHECK-SOLC`/`CHECK-SOLX`: solc's alloca/store-zero/load/return triple and solx's direct `sol.constant 0` + `sol.return` are both checked, so a regression on either compiler fails the test.

solx-mlir/tests/lit/literal_types.sol
hedgar2017: Done in 47b6f7a along your suggestion — `--check-prefixes=CHECK,CHECK-SOLC` and `--check-prefixes=CHECK,CHECK-SOLX` pin each compiler's `neg_int8` shape: solc's folded `sol.constant -1 : si8`, and solx's `1 : ui8` → cast to `si8` → `sub 0 - 1` (catalog item 9).

solx-mlir/tests/lit/logical.sol
hedgar2017: Done in 47b6f7a — `||` seed pinned per compiler: solc stores the loaded `a`, solx stores `arith.constant true`.

abinavpp (approved):
perfect! thank you!

nebasuke (approved):
Thanks, LGTM now!

.github/workflows/slang-tests.yaml
```diff
@@ -90,6 +90,20 @@ jobs:
           enable-mlir: 'true'
           enable-utils: 'true'
 
+      # TODO: remove this step and the Build solc step below once the lit
+      # tests stop running against solc.
+      - name: Sync solx-solidity for MLIR codegen
+        shell: bash
+        run: |
+          git -C solx-solidity fetch origin main
+          git -C solx-solidity checkout FETCH_HEAD
```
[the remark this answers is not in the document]
hedgar2017: By design.

solx-dev/src/utils.rs
```diff
@@ -217,13 +219,25 @@ pub fn remove(project_directory: &Path, project_name: &str) -> anyhow::Result<()
 }
 
 ///
-/// Call ninja to build the LLVM.
+/// Call ninja to build and install the LLVM. When `--enable-tests` produced
+/// `llvm-lit` in the build tree, also copy it into the install prefix so it
+/// is part of the cached artifact like the rest of the toolchain.
```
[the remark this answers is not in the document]
hedgar2017: Done in 437c5a9 (amended into the CI commit) — comment now reads "When `llvm-lit` exists in the build tree (produced by either `--enable-utils` or `--enable-tests`)", matching the actual condition (`if lit_source.exists()`).

solx-slang/src/ast/contract/mod.rs
```diff
@@ -78,33 +78,28 @@ impl<'state, 'context> ContractEmitter<'state, 'context> {
...
+        let entry = self.state.builder.emit_sol_func(
+            "constructor()",
+            &[],
+            &[],
+            None,
+            solx_mlir::StateMutability::NonPayable,
+            Some(solx_mlir::FunctionKind::Constructor),
+            &contract_body,
+        );
+        self.state.builder.emit_sol_return(&[], &entry);
```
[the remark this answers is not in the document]
hedgar2017: Known limitation, scoped out of this PR. This commit only fixes the layout-ordering bug — that the constructor stub must be emitted **before** the other functions to match solc's MLIR output. Wiring the actual constructor declaration's parameters, body, and `payable` mutability through the slang frontend is separate follow-up work; the lit test `special_functions.sol` documents the current gap (solx emits a no-arg empty constructor while solc emits the full one). The unconditional `constructor()` / `NonPayable` stub here matches what the rest of the slang lowering currently produces — tightening the stub will land together with the rest of the constructor lowering.

PR #386, opened 2026-05-01
fix(slang): skip no-op casts in TypeConversion::emit

hedgar2017:
[`TypeConversion::Address::emit`](https://github.com/NomicFoundation/solx/blob/4dc3a815e695d67e940a991a842edc0ce28cc2bf/solx-slang/src/ast/contract/function/expression/call/type_conversion.rs#L142-L155) always calls `sol.address_cast` and [`TypeConversion::Bool::emit`](https://github.com/NomicFoundation/solx/blob/4dc3a815e695d67e940a991a842edc0ce28cc2bf/solx-slang/src/ast/contract/function/expression/call/type_conversion.rs#L138-L141) always emits `sol.constant 0` + `sol.cmp ne` regardless of whether the source value's type already matches the target. Only `Self::Cast(t)` short-circuits, via [`Builder::emit_sol_cast`'s built-in guard](https://github.com/NomicFoundation/solx/blob/4dc3a815e695d67e940a991a842edc0ce28cc2bf/solx-mlir/src/context/builder/mod.rs#L943).

After [#384](https://github.com/NomicFoundation/solx/pull/384) routed emit/revert arguments through `TypeConversion::from_target_type(...).emit(...)`, every address argument to `sol.emit` and `sol.revert` flows through `Self::Address::emit` and produces a redundant `sol.address_cast %X : !sol.address to !sol.address` between every `sol.caller` (or address-typed `sol.load`) and its consumer. The same redundancy fires for `!a` on an i1 source (`sol.constant false` + `sol.cmp ne` round-trip on `bool`). The LLVM mid-end folds both, so EVM bytecode is unaffected; the Sol-dialect IR has been carrying the noise.

### DIFF
--- a/solx-slang/src/ast/contract/function/expression/call/type_conversion.rs
+++ b/solx-slang/src/ast/contract/function/expression/call/type_conversion.rs
@@ -91,6 +91,15 @@ impl<'context> TypeConversion<'context> {
         }
     }
 
+    /// Returns the MLIR target type this conversion produces.
+    pub fn to_target_type(&self, builder: &solx_mlir::Builder<'context>) -> Type<'context> {
+        match self {
+            Self::Bool => builder.types.i1,
+            Self::Address => builder.types.sol_address,
+            Self::Cast(target_type) => *target_type,
+        }
+    }
+
     /// Resolves the declared Solidity type of a state variable to an MLIR type.
     pub fn resolve_state_variable_type(
         state_variable: &StateVariableDefinition,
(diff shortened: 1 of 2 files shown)


solx-slang/src/ast/contract/function/expression/call/type_conversion.rs
```diff
@@ -134,6 +134,14 @@ impl<'context> TypeConversion<'context> {
     where
         'context: 'block,
     {
+        let target_type = match &self {
+            Self::Bool => builder.types.i1,
+            Self::Address => builder.types.sol_address,
+            Self::Cast(t) => *t,
+        };
```
ggiraldez: Would it make sense to extract this to a function `to_target_type`/`target_type`, symmetric to the existing `from_target_type`? I think that would improve readability.

hedgar2017: Totally, fixed!

ggiraldez (approved):
LGTM!

PR #401, opened 2026-05-04
feat(slang): emit contract-level constant references

hedgar2017:
Adds a `Definition::Constant` arm in the identifier lowering path so references to `uint256 constant FOO = 42;` and similar contract-level constants inline the initializer expression at the use site and cast it to the constant's declared type. Without the cast, slang types the literal at its natural width (e.g. `42` as `ui8`) and downstream arithmetic silently truncates against `uint256` declarations.

Lit test: `solx-mlir/tests/lit/constants.sol`.

Integration test delta (M3B3), `tests/solidity/simple/`:

| | base | tip | Δ |
|---|---|---|---|
| PASSED | 1662 | 1686 | +24 |
| FAILED | 7 | 7 | 0 |
| INVALID | 383 | 376 | −7 |
| Total | 2052 | 2069 | +17 |

Newly fully passing:

- tests/solidity/simple/expression/constant_conditional.sol
- tests/solidity/simple/expression/constant_match.sol
- tests/solidity/simple/expression/mixed_item_constantness.sol
- tests/solidity/simple/loop/range_bitlength_inference/u8_to_u256_const.sol
- tests/solidity/simple/loop/range_bitlength_inference/u8_to_u64_const.sol
- tests/solidity/simple/order/casted_declared_const.sol
- tests/solidity/simple/order/declared_const.sol

### DIFF
--- a/solx-slang/src/ast/contract/function/expression/mod.rs
+++ b/solx-slang/src/ast/contract/function/expression/mod.rs
@@ -177,19 +177,7 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
                         let initializer = constant
                             .value()
                             .ok_or_else(|| anyhow::anyhow!("constant {name} has no initializer"))?;
-                        let slang_type = constant.get_type().ok_or_else(|| {
-                            anyhow::anyhow!("unresolved type for constant: {name}")
-                        })?;
-                        let target_type = TypeConversion::resolve_slang_type(
-                            &slang_type,
-                            None,
-                            &self.state.builder,
-                        );
-                        let (value, block) = self.emit_value(&initializer, block)?;
-                        let cast =
-                            TypeConversion::from_target_type(target_type, &self.state.builder)
-                                .emit(value, &self.state.builder, &block);
-                        Ok((Some(cast), block))
+                        self.emit(&initializer, block)
                     }
                     None => anyhow::bail!("unresolved identifier: {name}"),
                     Some(_) => anyhow::bail!("unsupported identifier reference: {name}"),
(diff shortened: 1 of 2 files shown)


solx-slang/src/ast/contract/function/expression/mod.rs
```diff
@@ -176,6 +176,24 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
                                 .emit_sol_load(pointer, element_type, &block)?;
                         Ok((Some(value), block))
                     }
+                    Some(Definition::Constant(constant)) => {
+                        let initializer = constant
+                            .value()
+                            .ok_or_else(|| anyhow::anyhow!("constant {name} has no initializer"))?;
+                        let slang_type = constant.get_type().ok_or_else(|| {
+                            anyhow::anyhow!("unresolved type for constant: {name}")
+                        })?;
```
abinavpp: when can either of these fail? afaik, every constant in Solidity has both a type and an initializer.

ggiraldez: Typing can fail in the semantic analysis (eg. a reference to an undefined symbol). For the initializer, in the frontend we're creating `ConstantDefinition` both from top-level constants and contract level constant state variable definitions. The top-level constants are guaranteed by the grammar to have an initializer, but the state variables marked with `constant` don't, hence why the `value()` getter returns an `Option<>`. This should be caught in a validation pass in Slang.

ggiraldez (approved):
LGTM!

solx-slang/src/ast/contract/function/expression/mod.rs
hedgar2017: Makes sense, I'll convert all these to panics eventually.

PR #404, opened 2026-05-05
ci(llvm): disable LLVM tools on all platforms via install-distribution

### LINKED ISSUE #364: ci(llvm): reduce Windows LLVM tool count while keeping llvm-config (retry of reverted tool-disable)
## Context

On Windows hosted runners, `lld-link` is slow per-executable. With ccache at 100% hit on compiles, the dominant wall-clock in `build-llvm` is linking the ~200 LLVM tool binaries (`opt.exe`, `llc.exe`, `llvm-objdump.exe`, etc.). solx consumes LLVM purely as a library via inkwell FFI, so none of these tool binaries are invoked at runtime.

An attempt to disable them via `LLVM_BUILD_TOOLS=Off` + `LLVM_INCLUDE_TOOLS=Off` (commit 3106d1e in #360) was reverted because it broke the Rust test step.

## Evidence the approach works (and what broke)

Test run [#24771100328](https://github.com/NomicFoundation/solx/actions/runs/24771100328) at commit 3106d1e:

- **Build LLVM step on Windows: 42 min → 9:44 (4.3×).** The tool-disable mechanism worked exactly as hypothesised — ~200 fewer `lld-link` invocations.
- **Run tests step failed** with:
  ```
  error: No suitable version of LLVM was found system-wide or pointed
  error: could not compile `llvm-sys` (lib) due to 1 previous error
  ```

Root cause: the `llvm-sys` Rust crate (pulled in by inkwell) runs `${LLVM_SYS_211_PREFIX}/bin/llvm-config` at Rust crate-build time to discover include paths, library paths, and link flags. `llvm-config` is an LLVM "tool" under `llvm/tools/llvm-config/`, so `LLVM_INCLUDE_TOOLS=Off` drops its CMakeLists from the configure and the binary isn't produced.

`llvm-config` is the *only* tool binary solx needs (and only at Rust build time, not runtime). Every other tool we disabled is safely unused.

### Option A — surgical override

Keep `LLVM_INCLUDE_TOOLS=On` (so `tools/` subdirectory is configured) but `LLVM_BUILD_TOOLS=Off` (so tools aren't in the ALL target by default), then force just `llvm-config`:

```
-DLLVM_BUILD_TOOLS='Off'
-DLLVM_INCLUDE_TOOLS='On'
-DLLVM_TOOL_LLVM_CONFIG_BUILD='On'
```

LLVM's cmake uses the `LLVM_TOOL_<name>_BUILD` pattern to allow per-tool overrides of `LLVM_BUILD_TOOLS`. Expectation: `llvm-config.exe` is built, the ~200 other tool binaries are not.

Risks to validate in CI:
1. `LLVM_INCLUDE_TOOLS=On` re-emits configure/compile rules for all tools. If LLVM's cmake uses `EXCLUDE_FROM_ALL` correctly, ninja only builds `llvm-config` in practice. If not, the speedup partially erodes. Needs measurement.
2. Some tools may be tangled in `install` targets (e.g. via `llvm_install_symlink`), which could force them in via install-time dependencies. Watch the final `ninja install` step for hidden work.

### Option B — `LLVM_DISTRIBUTION_COMPONENTS` whitelist

Most surgical. Tell cmake: "when building the distribution target, include only these components." Shape roughly:

```
-DLLVM_INCLUDE_TOOLS='On'
-DLLVM_BUILD_TOOLS='Off'
-DLLVM_DISTRIBUTION_COMPONENTS='llvm-libraries;llvm-headers;cmake-exports;llvm-config;mlir-libraries;mlir-headers;lld-libraries'
```

Requires switching `solx-dev llvm build`'s final step from `ninja install` to `ninja install-distribution`. That's a ~one-line change in `solx-dev/src/llvm/build.rs` or equivalent. Clean but touches Rust code.

### Option C — scope to Windows only (same as the reverted commit)

Whichever of A or B lands should retain the Windows-only gate — other platforms weren't link-bound and had no issues with tools enabled.

## Expected impact

Based on the successful half of run #24771100328:

| Metric | Before | With tool-disable | Speedup |
|---|---|---|---|
| Windows Build LLVM step | ~42 min | ~9:44 | **4.3×** |
| Windows whole job | ~1h 19min | ~15–20 min est. | **~4×** |

Applied per Windows CI run on submodule bumps + any path where ccache is the only cache in play.

## Acceptance criteria

- [ ] Windows Build LLVM step completes in under 15 min on a warm-ccache run.
- [ ] `Run tests` step on Windows completes successfully (verifies `llvm-sys` found `llvm-config`).
- [ ] No regression on Linux/macOS legs (confirmed by wall-clock comparison against the last good run).
- [ ] `Show ccache stats` still reports near-100% hit rate on the warm path.

## Out of scope

- Linux/macOS link-phase optimisations — Linux/macOS weren't link-bound in the #360 measurements.
- Broader `ninja install-distribution` rework on all platforms.

## Related

- #360 — added ccache; included the first attempt at Windows tool-disable (commit 3106d1e, reverted).
- #245 — earlier LLVM cmake flag tuning (link-jobs cap, SPLIT_DWARF, OPTIMIZED_TABLEGEN).

nebasuke:
Closes #364. Extends the Windows-only design from #365 to Linux and macOS, since macOS x86 (the macos-15-intel runner) is now also link-bound — see #402's drop to LLVM_PARALLEL_LINK_JOBS=1.

## What

solx never invokes any LLVM tool at runtime; it consumes LLVM as a library through inkwell. So building and linking the ~200 tool binaries (`opt`, `llc`, `lli`, ...) on every cold cache miss is pure wall-clock waste. This applies the install-distribution mechanism across all five platforms, with no Windows-only naming.

`solx-dev/src/llvm/platforms/shared.rs::build_opts_distribution` emits:

- `LLVM_BUILD_TOOLS=Off` — every tool target marked `EXCLUDE_FROM_ALL`.
- `LLVM_INCLUDE_TOOLS=On` — `tools/` stays in the configure pass so the umbrella targets (`llvm-libraries`, `lld-libraries`, optionally `mlir-libraries`) are defined.
- `LLVM_DISTRIBUTION_COMPONENTS` — whitelist of what `install-distribution` actually copies into the prefix: llvm/lld/(optionally mlir) library + header + cmake-export components, plus `FileCheck` when `--enable-utils` is true (used by `solx-mlir/tests/lit/*.sol` RUN lines).

`build_and_install_llvm_config` runs after `install-distribution` on every platform: with tools disabled, cmake doesn't create the `install-llvm-config` target, so `llvm-sys` would have no `llvm-config` to query at Rust build time. The helper invokes `ninja llvm-config` directly and copies the binary into the install prefix's `bin/`. `cfg!(target_os = "windows")` handles the `.exe` suffix; `overwrite: true` makes local reruns succeed (default would error on `AlreadyExists`).

## Wall-clock impact

Build LLVM step, single-sample per cell.

### Warm ccache + cold artifact cache

Both runs went through Build LLVM rather than restoring from artifact cache.

| Platform | Baseline | This PR | Δ |
|---|---:|---:|---:|
| Linux x86 gnu | 6:36 | 3:39 | −45% |
| Linux ARM64 gnu | 6:09 | 2:59 | −52% |
| MacOS x86 | 29:34 | 18:37 | −37% |
| MacOS arm64 | 1:49:28 | 5:31 | −95% |
| Windows | 51:47 | 15:00 | −71% |

Baseline: [run 25415956903](https://github.com/NomicFoundation/solx/actions/runs/25415956903) on `ci-slang-mlir-lit-tests` (no tool-disable). This PR: [run 25407488943](https://github.com/NomicFoundation/solx/actions/runs/25407488943).

The MacOS arm64 95% is genuine and worth flagging: that runner was thrashing under `LLVM_PARALLEL_LINK_JOBS=2` on a 7GB host (this baseline pre-dates the macOS arm64 `LINK_JOBS=1` reduction). Tool-disable cuts total link-phase memory pressure to the point where the host stops swapping. (For macOS x86, this looked like an opportunity to lift `LINK_JOBS=1` back to 2 — but the probe below shows it still thrashes; we keep =1.)

### Cold ccache + cold artifact cache

This PR's cold numbers come from a TEMP cache-key-bust commit forcing both the artifact cache and ccache to miss on every platform. Two samples per platform: the `Slang Tests` and `Tests` workflows running on the same TEMP-bumped SHA.

| Platform | Baseline (cold) | This PR cold, Slang Tests | This PR cold, Tests |
|---|---:|---:|---:|
| Linux x86 gnu | 2h 15min¹ | 2:15:00 | 2:14:40 |
| Linux ARM64 gnu | 1h 29min¹ | 1:16:25 | 1:14:47 |
| MacOS x86 | 3h 33min² | 3:32:09 | 3:04:10 |
| MacOS arm64 | 3h 00min¹ | 1:52:14 | 1:40:11 |
| Windows | 3h 24min¹ | 2:19:16 | 2:40:46 |

Cold runs: Slang Tests [run 25423417125](https://github.com/NomicFoundation/solx/actions/runs/25423417125), Tests [run 25423417118](https://github.com/NomicFoundation/solx/actions/runs/25423417118).

The compile-bound vs link-bound split is the main story:

- **Link-bound platforms** (Windows, MacOS arm64) get the bulk of the savings — these are the ones where tool linking actually dominated cold wall-clock.
- **Compile-bound platforms** (Linux x86, MacOS x86) are essentially flat. Library compilation dominates cold time, and tool-disable doesn't change library compilation. MacOS x86 is also serialised at `LLVM_PARALLEL_LINK_JOBS=1` (see next subsection), so any link-phase savings are serialised anyway.
- **Linux ARM64** is in between — small but real win.

#### macOS x86 LINK_JOBS=2 probe — proves we still need LINK_JOBS=1

A second TEMP commit layered on top of the cache-bust flipped macOS x86 back to `LLVM_PARALLEL_LINK_JOBS=2`, testing the hypothesis that tool-disable's removal of ~200 tool binaries lifts the CPU-bound serialization rationale that motivated #402's drop to 1.

Result: cold time on macOS x86 swings far enough across runs to indicate the runner is thrashing under 2 parallel links, not just running noisier.

| Config (macOS x86 only) | Slang Tests | Tests | Spread |
|---|---:|---:|---:|
| LINK_JOBS=1 | 3:32:09 | 3:04:10 | 28 min |
| LINK_JOBS=2 | 2:46:38 | 4:31:08 | **1h 45min** |

LINK_JOBS=2 produces a 1h 45min spread — over 4× wider than LINK_JOBS=1's 28 min — on the same TEMP-bumped SHA. The fast LJ=2 sample is below LJ=1's range; the slow one is well above it. That bimodality is what thrashing on a CPU-/RAM-marginal runner looks like: 2 parallel `lld` invocations sometimes fit the runner's budget and sometimes overshoot it, producing unpredictable wall-clock.

LINK_JOBS=1 doesn't have that exposure — both samples land within 28 min of each other. **macOS x86 therefore stays at `LLVM_PARALLEL_LINK_JOBS=1` in this PR.** The "tool-disable might let us bump LINK_JOBS back up" hypothesis from the warm-cache narrative above turns out to be wrong for macOS x86; arm64 is untested but should be assumed similar until proven otherwise.

LJ=2 probe runs: Slang Tests [run 25436671658](https://github.com/NomicFoundation/solx/actions/runs/25436671658), Tests [run 25436671659](https://github.com/NomicFoundation/solx/actions/runs/25436671659).

¹ [#396 cold reference](https://github.com/NomicFoundation/solx/actions/runs/25321256321) (post-utils, pre-link-jobs=1 on macOS x86).
² [#402 cold](https://github.com/NomicFoundation/solx/actions/runs/25365589502) — latest macOS x86 baseline, with `LLVM_PARALLEL_LINK_JOBS=1`.

## Cache-key changes (.github/actions/build-llvm/action.yml)

- Added `dev<hash>` dimension (over `solx-dev/src/llvm/**`) to the artifact cache key. Any Rust-side cmake-arg change now invalidates stale artifact-cache entries automatically — previously, modifying `shared.rs` could serve a stale-config-built artifact.
- Kept `-utils` and `-tests` dimensions in **both** the artifact and ccache keys, matching the precedent set by #396. With `--enable-utils=true`, `FileCheck` lands in the install prefix; the cached artifact is materially different.

## First-merge behaviour

The `dev<hash>` dimension is new, so every platform's artifact cache will miss once after this lands on `main`. ccache pools are unchanged (no key dimensions added/removed there). Expect Windows cold rebuild ~3h, others bounded by ccache reuse.

## Out of scope

- `LLVM_BUILD_LLVM_DYLIB` exploration (separate research task).

### DIFF
--- a/solx-dev/src/llvm/platforms/shared.rs
+++ b/solx-dev/src/llvm/platforms/shared.rs
@@ -2,6 +2,9 @@
 //! The shared options for building various platforms.
 //!
 
+use std::path::Path;
+use std::process::Command;
+
 use crate::build_type::BuildType;
 use crate::llvm::platforms::Platform;
 use crate::llvm::sanitizer::Sanitizer;
@@ -184,6 +187,107 @@ pub fn shared_build_opts_coverage(enabled: bool) -> Vec<String> {
         "-DLLVM_BUILD_INSTRUMENTED_COVERAGE='{}'",
         if enabled { "On" } else { "Off" },
     )]
+}
+
+///
+/// Configure an `install-distribution` that ships only what solx needs,
+/// skipping the ~200 LLVM tool binaries (`opt`, `llc`, ...).
+///
+/// solx never invokes any LLVM tool at runtime — it consumes LLVM as a
+/// library via inkwell — so linking those tools is pure wall-clock waste in
+/// CI. `LLVM_BUILD_TOOLS=Off` marks every tool `EXCLUDE_FROM_ALL`;
+/// `LLVM_INCLUDE_TOOLS=On` keeps `tools/` in the configure pass so the
+/// umbrella targets (`llvm-libraries`, `lld-libraries`, etc.) stay defined.
+/// `LLVM_DISTRIBUTION_COMPONENTS` whitelists what `install-distribution`
+/// actually copies into the install prefix — see #364.
+///
+/// `llvm-config` is deliberately **not** in the distribution list: with
+/// `LLVM_BUILD_TOOLS=Off`, cmake's `llvm_add_tool` skips creation of the
+/// `install-llvm-config` target, so referencing it here errors at configure
+/// time. `llvm-sys` still needs `llvm-config` at Rust build time —
+/// `build_and_install_llvm_config` builds it directly via `ninja
+/// llvm-config` and copies the binary into the install prefix afterwards.
+/// `lld-*` is always included because `shared_build_opts_projects` always
+/// enables `lld`. `mlir-*` is included only when `enable_mlir` is true.
+///
+/// When `enable_utils` is true, `FileCheck` is added to the whitelist —
+/// `solx-mlir/tests/lit/*.sol` uses it in RUN lines, so it must land in the
+/// install prefix's `bin/`. LLVM's `add_llvm_utility` registers each utility
+/// under its own install component (named after the target). The component
+/// only exists when `LLVM_BUILD_UTILS=On` (set by `shared_build_opts_utils`
+/// when `--enable-utils` is passed); listing it without that flag errors at
+/// configure time.
+///
+/// `llvm-lit` is **not** listed here despite being needed by the slang-tests
+/// workflow runner: upstream LLVM has no install rule for `llvm-lit` (it's
+/// only `configure_file`'d into the build dir), so referencing it as a
+/// distribution component errors at configure time. How `target-final/bin/
+/// llvm-lit` reaches the install prefix today is independent of this PR.
+///
+pub fn build_opts_distribution(enable_mlir: bool, enable_utils: bool) -> Vec<String> {
+    let mut components = vec![
+        "llvm-libraries",
+        "llvm-headers",
+        "cmake-exports",
+        "lld-libraries",
+        "lld-headers",
+        "lld-cmake-exports",
+    ];
+    if enable_mlir {
+        components.extend(["mlir-libraries", "mlir-headers", "mlir-cmake-exports"]);
+    }
+    if enable_utils {
+        components.push("FileCheck");
+    }
+    vec![
+        "-DLLVM_BUILD_TOOLS='Off'".to_owned(),
+        "-DLLVM_INCLUDE_TOOLS='On'".to_owned(),
+        format!("-DLLVM_DISTRIBUTION_COMPONENTS='{}'", components.join(";")),
+    ]
+}
+
+///
+/// Build `llvm-config` directly (it has no install target under
+/// `LLVM_BUILD_TOOLS=Off`) and copy it into the install prefix's `bin/`.
+///
+/// `llvm-sys` invokes `llvm-config` at Rust build time to discover libs and
+/// flags. Both ninja args and the destination path use forward-slash form on
+/// Windows; callers are expected to pass already-normalized paths (existing
+/// Windows code uses `path_windows_to_unix` for this).
+///
+/// `overwrite: true` lets a second `solx-dev llvm build` against the same
+/// target prefix succeed; `fs_extra::file::copy`'s default would error on
+/// `AlreadyExists` for the second run.
+///
+pub fn build_and_install_llvm_config(build_dir: &Path, target_dir: &Path) -> anyhow::Result<()> {
+    let exe_suffix = if cfg!(target_os = "windows") {
+        ".exe"
+    } else {
+        ""
+    };
+    let bin_name = format!("llvm-config{exe_suffix}");
+    let build_dir_str = build_dir.to_string_lossy();
+
+    crate::utils::command(
+        Command::new("ninja")
+            .arg("-C")
+            .arg(&*build_dir_str)
+            .arg("llvm-config"),
+        "Building llvm-config",
+    )?;
+
+    let source = build_dir.join("bin").join(&bin_name);
+    let dest_dir = target_dir.join("bin");
+    std::fs::create_dir_all(&dest_dir)?;
+    fs_extra::file::copy(
+        &source,
+        dest_dir.join(&bin_name),
+        &fs_extra::file::CopyOptions {
+            overwrite: true,
+            ..Default::default()
+        },
+    )?;
+    Ok(())
 }
 
 ///
--- a/solx-dev/src/utils.rs
+++ b/solx-dev/src/utils.rs
@@ -219,16 +219,27 @@ pub fn remove(project_directory: &Path, project_name: &str) -> anyhow::Result<()
 }
 
 ///
-/// Call ninja to build and install the LLVM. When `llvm-lit` exists in the
-/// build tree (produced by either `--enable-utils` or `--enable-tests`), also
-/// copy it into the install prefix so it is part of the cached artifact like
-/// the rest of the toolchain.
+/// Call ninja to build and install LLVM via the given install target.
 ///
-pub fn ninja(build_dir: &Path) -> anyhow::Result<()> {
+/// All platforms pass `"install-distribution"` together with
+/// `LLVM_DISTRIBUTION_COMPONENTS` to skip the ~200 LLVM tool binaries solx
+/// doesn't use (it consumes LLVM as a library via inkwell). See #364.
+///
+/// When `llvm-lit` exists in the build tree (produced by `--enable-utils`),
+/// it is copied into the install prefix's `bin/` afterwards. Upstream LLVM
+/// has no install rule for `llvm-lit` (it's only `configure_file`'d into the
+/// build dir), so listing it in `LLVM_DISTRIBUTION_COMPONENTS` errors at
+/// configure time; the manual copy bridges that gap so the cached artifact
+/// contains `llvm-lit` like the rest of the toolchain.
+///
+pub fn ninja(build_dir: &Path, install_target: &str) -> anyhow::Result<()> {
     let mut ninja = Command::new("ninja");
     let build_dir_str = build_dir.to_string_lossy();
     ninja.args(["-C", &*build_dir_str]);
-    command(ninja.arg("install"), "Running ninja install")?;
+    command(
+        ninja.arg(install_target),
+        &format!("Running ninja {install_target}"),
+    )?;
     let lit_name = if cfg!(windows) {
         "llvm-lit.py"
     } else {
@@ -237,6 +248,7 @@ pub fn ninja(build_dir: &Path) -> anyhow::Result<()> {
     let lit_source = build_dir.join("bin").join(lit_name);
     if lit_source.exists() {
         let install_bin = LlvmPath::llvm_target_final()?.join("bin");
+        std::fs::create_dir_all(&install_bin)?;
         std::fs::copy(&lit_source, install_bin.join(lit_name))?;
     }
     Ok(())
--- a/.github/actions/build-llvm/action.yml
+++ b/.github/actions/build-llvm/action.yml
@@ -62,19 +62,28 @@ runs:
       working-directory: ${{ inputs.working-dir }}
       run: |
         SHA=$(git -C solx-llvm rev-parse HEAD)
+        # solx-dev owns LLVM-build cmake args (shared.rs + platforms/*.rs).
+        # Hash that tree into the key so any Rust-side flag change invalidates
+        # stale artifact-cache entries automatically. Truncate for readability.
+        DEV_HASH=$(printf %s '${{ hashFiles(format('{0}/solx-dev/src/llvm/**', inputs.working-dir)) }}' | head -c 8)
         KEY="v1-llvm-${{ runner.os }}-${{ runner.arch }}"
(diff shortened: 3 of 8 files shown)


solx-dev/src/llvm/platforms/shared.rs
```diff
@@ -186,6 +189,107 @@ pub fn shared_build_opts_coverage(enabled: bool) -> Vec<String> {
...
+    let bin_name = format!("llvm-config{exe_suffix}");
+    let build_dir_str = build_dir.to_string_lossy();
+
+    crate::utils::command(
+        Command::new("ninja")
+            .arg("-C")
+            .arg(&*build_dir_str)
+            .arg("llvm-config"),
+        "Building llvm-config",
+    )?;
```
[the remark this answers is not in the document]
hedgar2017: Hehe, nice try, but the CI has already passed!

solx-dev/src/utils.rs
```diff
@@ -219,16 +219,27 @@ pub fn remove(project_directory: &Path, project_name: &str) -> anyhow::Result<()
...
+/// doesn't use (it consumes LLVM as a library via inkwell). See #364.
+///
+/// When `llvm-lit` exists in the build tree (produced by `--enable-utils`),
+/// it is copied into the install prefix's `bin/` afterwards. Upstream LLVM
+/// has no install rule for `llvm-lit` (it's only `configure_file`'d into the
+/// build dir), so listing it in `LLVM_DISTRIBUTION_COMPONENTS` errors at
+/// configure time; the manual copy bridges that gap so the cached artifact
+/// contains `llvm-lit` like the rest of the toolchain.
+///
+pub fn ninja(build_dir: &Path, install_target: &str) -> anyhow::Result<()> {
```
hedgar2017, line 235: Every callsite passes the literal `"install-distribution"`, and the doc above explicitly says so — the parameter has no second user. Suggest inlining the constant in the body and dropping the parameter.

solx-dev/src/llvm/platforms/shared.rs
```diff
@@ -186,6 +189,107 @@ pub fn shared_build_opts_coverage(enabled: bool) -> Vec<String> {
...
+/// `llvm-sys` invokes `llvm-config` at Rust build time to discover libs and
+/// flags. Both ninja args and the destination path use forward-slash form on
+/// Windows; callers are expected to pass already-normalized paths (existing
+/// Windows code uses `path_windows_to_unix` for this).
+///
+/// `overwrite: true` lets a second `solx-dev llvm build` against the same
+/// target prefix succeed; `fs_extra::file::copy`'s default would error on
+/// `AlreadyExists` for the second run.
+///
+pub fn build_and_install_llvm_config(build_dir: &Path, target_dir: &Path) -> anyhow::Result<()> {
```
hedgar2017, line 262: All five `platforms/*.rs` files now end with the same two-line block: `crate::utils::ninja(llvm_build_final.as_ref(), "install-distribution")?` followed by `build_and_install_llvm_config(llvm_build_final.as_ref(), llvm_target_final.as_ref())?`. Worth a small wrapper here in `shared` (e.g. `install_distribution(build_dir, install_prefix)`) so a future post-install step lands in one place instead of five.

hedgar2017 (commented):
@nebasuke — wall-clock numbers are wild, mechanism is right. Four small things, none blocking.

solx-dev/src/llvm/platforms/shared.rs
```diff
@@ -186,6 +189,107 @@ pub fn shared_build_opts_coverage(enabled: bool) -> Vec<String> {
...
+        format!("-DLLVM_DISTRIBUTION_COMPONENTS='{}'", components.join(";")),
+    ]
+}
+
+///
+/// Build `llvm-config` directly (it has no install target under
+/// `LLVM_BUILD_TOOLS=Off`) and copy it into the install prefix's `bin/`.
+///
+/// `llvm-sys` invokes `llvm-config` at Rust build time to discover libs and
+/// flags. Both ninja args and the destination path use forward-slash form on
```
hedgar2017, line 254: "Both ninja args and the destination path use forward-slash form on Windows" overstates: `target_dir.join("bin").join(&bin_name)` uses `PathBuf::join`, which appends `\` on Windows, so the destination ends up with mixed separators. `fs_extra::file::copy` accepts that on NTFS so there's no bug — but the doc claim is wrong. Drop "and the destination path", or note the mixed separators are tolerated.

solx-dev/src/llvm/platforms/shared.rs
```diff
@@ -186,6 +189,107 @@ pub fn shared_build_opts_coverage(enabled: bool) -> Vec<String> {
...
+/// Configure an `install-distribution` that ships only what solx needs,
+/// skipping the ~200 LLVM tool binaries (`opt`, `llc`, ...).
+///
+/// solx never invokes any LLVM tool at runtime — it consumes LLVM as a
+/// library via inkwell — so linking those tools is pure wall-clock waste in
+/// CI. `LLVM_BUILD_TOOLS=Off` marks every tool `EXCLUDE_FROM_ALL`;
+/// `LLVM_INCLUDE_TOOLS=On` keeps `tools/` in the configure pass so the
+/// umbrella targets (`llvm-libraries`, `lld-libraries`, etc.) stay defined.
+/// `LLVM_DISTRIBUTION_COMPONENTS` whitelists what `install-distribution`
+/// actually copies into the install prefix — see #364.
```
hedgar2017, line 202: Three bare-number references in the new doc comments:

- `shared.rs:202` (this line) — "see #364" (issue #364)
- `shared.rs:225` — "is independent of this PR" (PR #404, worst of the three — no referent once this lands)
- `utils.rs:226` — "See #364" (issue #364)

No links, just the numbers — readers have to guess which repo and click around to find them. Suggest dropping all three; the WHY each wraps stands on its own, and the provenance lives in the commit message and PR description.

hedgar2017 (approved):

solx-dev/src/utils.rs
nebasuke: Done in 6508adc. Dropped the parameter and inlined `"install-distribution"`; updated the doc accordingly.

solx-dev/src/llvm/platforms/shared.rs
nebasuke: Done in 6508adc. Added `shared::install_distribution(build_dir, target_dir)` that runs `ninja` and then materialises `llvm-config`. All five platform files now go through it. Demoted `build_and_install_llvm_config` to module-private since the wrapper is the only entry point.

solx-dev/src/llvm/platforms/shared.rs
nebasuke: Fixed in 6508adc. Doc now says only the ninja `-C` arg is normalised to forward-slash; the destination keeps native separators (`target_dir.join("bin")` appends `\` on Windows) and `fs_extra::file::copy` tolerates the mix on NTFS.

solx-dev/src/llvm/platforms/shared.rs
nebasuke: Dropped all three in 6508adc. The surrounding doc already explains the WHY in each spot; provenance lives in the commit message and PR description.

PR #410, opened 2026-05-10
feat(slang): resolve UserDefinedValue types via target-type lookup

hedgar2017:
Adds Slang→Sol-dialect resolution for the `UserDefinedValue` variant by following the binder's `target_type_id` to the wrapped elementary type and recursing through `TypeConversion::resolve_slang_type`. UDVT-typed parameters and returns lower to the underlying integer/address/bool MLIR type instead of panicking at the catch-all `unimplemented!("unsupported Slang type")` arm.

Reaches the resolved target via a new `UserDefinedValueType::target_type()` AST accessor in slang ([NomicFoundation/slang `feat-udvt-target-type`](https://github.com/NomicFoundation/slang/commit/8754a8fc4b11f66ef022c4cc7864714f09106297)). CI fetches that rev directly off the fork's branch — no upstream slang PR yet.

Depends on #390.

Lit test: `solx-mlir/tests/lit/user_defined_value_type.sol`.

## Test deltas

`tests/solidity/simple/ -O M3B3` vs base [`71e9071`](https://github.com/NomicFoundation/solx/commit/71e9071859f1a0a13b95ab8352069762835120e9): unchanged at 1662 / 7 / 383 (no UDVT-using fixtures).

`solx-solidity/test/libsolidity/semanticTests/`: 803 / 375 / 1173 → 805 / 377 / 1171.

`.sol` files moved from INVALID to FAILED:

- `solx-solidity/test/libsolidity/semanticTests/operators/userDefined/fixed_point_udvt_with_operators.sol` — `applyInterest:1` produces wrong runtime output.
- `solx-solidity/test/libsolidity/semanticTests/operators/userDefined/operator_parameter_and_return_cleanup_between_calls.sol` — `divAddNoOverflow:1` produces wrong runtime output.

### DIFF
--- a/solx-mlir/tests/lit/user_defined_value_type.sol
+++ b/solx-mlir/tests/lit/user_defined_value_type.sol
@@ -1,0 +1,19 @@
+// RUN: solx --emit-mlir=sol %s | FileCheck %s
+// RUN: solc --mlir-action=print-init %s 2>/dev/null | FileCheck %s
+
+// CHECK: sol.func @{{.*id_u.*}}: ui256) -> ui256
+// CHECK: sol.func @{{.*id_s.*}}: si8) -> si8
+// CHECK: sol.func @{{.*id_a.*}}: !sol.address) -> !sol.address
+// CHECK: sol.func @{{.*id_b.*}}: i1) -> i1
+
+contract C {
+    type U is uint256;
+    type S is int8;
+    type A is address;
+    type B is bool;
+
+    function id_u(U x) public pure returns (U) { return x; }
+    function id_s(S x) public pure returns (S) { return x; }
+    function id_a(A x) public pure returns (A) { return x; }
+    function id_b(B x) public pure returns (B) { return x; }
+}
--- a/solx-slang/src/ast/contract/function/expression/call/type_conversion.rs
+++ b/solx-slang/src/ast/contract/function/expression/call/type_conversion.rs
@@ -197,6 +197,12 @@ impl<'context> TypeConversion<'context> {
                 let max = u8::try_from(member_count - 1).expect("enum member count fits in u8");
                 builder.types.enumeration(max.into())
             }
+            SlangType::UserDefinedValue(udvt) => {
+                let target_type = udvt
+                    .target_type()
+                    .expect("UDVT target type resolved by semantic analysis");
+                Self::resolve_slang_type(&target_type, inherited_location, builder)
+            }
             _ => unimplemented!("unsupported Slang type"),
         }
     }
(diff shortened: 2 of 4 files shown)


solx-mlir/tests/lit/user_defined_value_type.sol
```diff
@@ -0,0 +1,19 @@
+// RUN: solx --emit-mlir=sol %s | FileCheck %s
+// RUN: solc --mlir-action=print-init %s 2>/dev/null | FileCheck %s
+
+// CHECK: sol.func {{.*}}id_u{{.*}}ui256{{.*}}ui256
+// CHECK: sol.func {{.*}}id_s{{.*}}si8{{.*}}si8
+// CHECK: sol.func {{.*}}id_a{{.*}}!sol.address{{.*}}!sol.address
+// CHECK: sol.func {{.*}}id_b{{.*}}i1{{.*}}i1
```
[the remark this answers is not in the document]
hedgar2017: Addressed in `fc92307`: CHECK lines now anchor on `@` and on the signature suffix `: TYPE) -> TYPE`. The single `{{.*id_x.*}}` wildcard absorbs the symbol-name divergence between solx (`@"id_u(uint256)"`) and solc (`@id_u_<node_id>`); using a literal symbol would require `--check-prefix` to split per-emitter.

abinavpp (approved):
thank you!

solx-slang/src/ast/contract/function/expression/call/type_conversion.rs
```diff
@@ -189,6 +189,12 @@ impl<'context> TypeConversion<'context> {
                 let max = u32::try_from(member_count - 1).expect("enum member count fits in u32");
                 builder.types.enumeration(max)
             }
+            SlangType::UserDefinedValue(udvt) => {
+                let target_type = udvt
+                    .target_type()
+                    .expect("UDVT target type resolved by semantic analysis");
+                Self::resolve_slang_type(&target_type, inherited_location, builder)
+            }
```
ggiraldez, line 205: Is this correct? Solidity's UDVT are new types not implicitly convertible to the target type. They are not aliases, but whole new types. You're supposed to use the `.wrap()` and `.unwrap()` built-ins to access the underlying type. They also dispatch most operators through `using` directives. Is `TypeConversion` only relevant for the underlying representation?

ggiraldez (commented):
I'm not sure what's the implication of the `TypeConversion` case. Ignore me if it only affects the underlying representation, as I think this should be fine in that case.

solx-slang/src/ast/contract/function/expression/call/type_conversion.rs
hedgar2017: @ggiraldez Yes, only the underlying representation. `resolve_slang_type` returns the MLIR `Type` — it's the type resolver, not a cast emitter; UDVT and its target share representation, so they map to the same MLIR type.

PR #426, opened 2026-05-11
feat(slang): emit state variable initializers in constructor

hedgar2017:
Emits state variable initializers (`T x = <expr>;` at contract scope) at the top of the constructor body, branched off `feat-slang-require-runtime-message`. Reference-typed slots (struct, array, mapping, bytes, string) are written via `sol.copy` from the evaluated value into the storage reference; primitive slots use `sol.store` after casting to the declared type. Mirrors solc's MLIR.

Array-literal initializers (`uint8[2] k = [1, 2];`) depend on `Expression::ArrayExpression` lowering which lives in a different stack — those are skipped here so the slot stays at its default zero, matching pre-initializer behavior.

Integration test delta (M3B3), `tests/solidity/simple/`:

| | base | tip | Δ |
|---|---|---|---|
| PASSED | 1899 | 1901 | +2 |
| FAILED | 8 | 6 | −2 |
| INVALID | 304 | 304 | 0 |
| Total | 2211 | 2211 | 0 |

Newly fully passing:

- tests/solidity/simple/solidity_by_example/simple/view_and_pure_functions.sol

### DIFF
--- a/solx-mlir/src/context/builder/mod.rs
+++ b/solx-mlir/src/context/builder/mod.rs
@@ -46,6 +46,7 @@ use crate::ods::sol::ConditionOperation;
 use crate::ods::sol::ConstantOperation;
 use crate::ods::sol::ContinueOperation;
 use crate::ods::sol::ContractOperation;
+use crate::ods::sol::CopyOperation;
 use crate::ods::sol::DoWhileOperation;
 use crate::ods::sol::ForOperation;
 use crate::ods::sol::FuncOperation;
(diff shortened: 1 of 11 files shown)


abinavpp:
can you update the branches? this pr has changes from #425

hedgar2017:
> can you update the branches? this pr has changes from #425

It's just some very weird github behavior. Need to do a rebase after every single push.

I will update! Could you just look at the top commit for now?

solx-mlir/src/context/builder/mod.rs
```diff
@@ -734,6 +735,33 @@ impl<'context> Builder<'context> {
...
 
+    /// Emits a `sol.copy` between two references.
+    ///
+    /// Use for source-level assignments that cross data locations (e.g. a
+    /// state-variable initializer copying a memory string literal into the
+    /// declared storage slot).
+    ///
+    /// # Panics
+    ///
+    /// Panics if the MLIR operation cannot be constructed.
```
abinavpp, line 746: do we really need to say this? this is highly unlikely

solx-mlir/tests/lit/state_var_initializer.sol
```diff
@@ -0,0 +1,15 @@
+// RUN: solx --emit-mlir=sol %s | FileCheck %s
```
abinavpp, line 1: solc here? or is it broken for this?

solx-slang/src/ast/contract/function/mod.rs
```diff
@@ -179,6 +194,43 @@ impl<'state, 'context> FunctionEmitter<'state, 'context> {
...
+    /// # Errors
+    ///
+    /// Returns an error if a state-variable initializer has an unresolved
+    /// type or contains unsupported constructs, or if the explicit
+    /// constructor body contains unsupported statements.
+    ///
+    /// # Panics
+    ///
+    /// Panics if an entry block is not attached to a region, which is
+    /// unreachable because `emit_sol_func` always creates a region.
```
abinavpp, line 212: ditto here with panic section. i don't think we should comment about unlikely situations in api's

abinavpp (approved):
lgtm, just some nits

solx-mlir/tests/lit/state_var_initializer.sol
hedgar2017: already added in #427! it's the same test

solx-slang/src/ast/contract/function/mod.rs
hedgar2017: makes sense, but it's just a Rust convention, I can scrap these in another PR across the repo later

PR #428, opened 2026-05-11
feat(mlir): emit wide integer constants without parser round-trip

hedgar2017:
Replaces the `BigInt → decimal → MLIR attribute parser` round-trip in `Builder::emit_constant` with a direct APInt-based construction via a new `solxCreateIntegerAttr` C wrapper in `sol_attr_stubs.cpp`. The Rust side passes the BigInt magnitude as LSB-first `u64` words plus a sign flag; the wrapper builds the APInt and negates it when the value is negative.

Bumps the solx-llvm submodule to bring in the upstream cherry-pick of `mlirIntegerAttrGetFromWords` and the matching accessors from llvm/llvm-project#177733 (matter-labs/solx-llvm#80).

### DIFF
--- a/solx-mlir/sol_attr_stubs.cpp
+++ b/solx-mlir/sol_attr_stubs.cpp
@@ -10,9 +10,14 @@
 
 #include "mlir/Dialect/Sol/Sol.h"
 #include "mlir/IR/MLIRContext.h"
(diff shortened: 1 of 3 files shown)


PavelKopyl (approved):
LGTM.

PR #431, opened 2026-05-11
fix(evmla): track typed stack through SWAPX/DUPX

hedgar2017:
solx EVMLA pipeline produces bytecode that hits `InvalidFEOpcode` at runtime for any external call with ≥15 value-type arguments. All gas is burned. Reproduces at every optimization level (`-O1`/`-O2`/`-O3`/`-Os`/`-Oz`). Empirically verified end-to-end via anvil execution of the repro bytecode.

solc emits `SWAPX` / `DUPX` (pseudo-ops with depth pushed as an immediate constant) when the stack depth exceeds 16. The EVMLA→LLVM lifter has explicit typed-stack handlers for [`SWAP1`..`SWAP16`](https://github.com/NomicFoundation/solx/blob/6742e5dd2a89635ff5a9631623182be29e2c3f8a/solx-evm-assembly/src/ethereal_ir/function/mod.rs#L391-L506) and [`DUP1`..`DUP16`](https://github.com/NomicFoundation/solx/blob/6742e5dd2a89635ff5a9631623182be29e2c3f8a/solx-evm-assembly/src/ethereal_ir/function/mod.rs#L508-L572) but none for the variable-depth variants. The runtime IR-memory swap/dup ran via [`solx-evm-assembly/src/assembly/instruction/stack.rs`](https://github.com/NomicFoundation/solx/blob/6742e5dd2a89635ff5a9631623182be29e2c3f8a/solx-evm-assembly/src/assembly/instruction/stack.rs#L50-L120) and worked correctly; only the typed model `block_stack` went out of sync.

After the desync, the encoder helper's return-cleanup `JUMP` (`PUSH 11; SWAPX; SWAP16; 16×POP; JUMP`) saw a non-tag element where the return tag should be. The lifter's [silent error path at `function/mod.rs:215`](https://github.com/NomicFoundation/solx/blob/6742e5dd2a89635ff5a9631623182be29e2c3f8a/solx-evm-assembly/src/ethereal_ir/function/mod.rs#L215) substituted `@llvm.evm.invalid + unreachable` and truncated the rest of the block. LLVM then correctly inferred the encoder as `noreturn` and DCE'd the downstream `CALL`/`STATICCALL` in the caller — final bytecode falls into bare `INVALID` instead of executing the call.


vladimirradosavljevic (approved):
LGTM, thanks!

PR #439, opened 2026-05-12
feat(solx-dev): add --install-distribution flag for lean CI install

hedgar2017:
Default LLVM install is now the full toolset (`ninja install`); `--install-distribution` opts into `ninja install-distribution` plus `llvm-config` for CI artifacts. `--enable-tests` overrides `--install-distribution` because the LLVM regression suite depends on the full toolset.

Drops `--enable-tools` entirely — local builds get the full toolset by default without needing a flag, and CI selects the lean variant via the new opt-in.

### DIFF
--- a/.github/actions/build-llvm/action.yml
+++ b/.github/actions/build-llvm/action.yml
@@ -202,7 +202,7 @@ runs:
           LINK_JOBS=1
         fi
 
-        ./target/release/solx-dev llvm build --build-type ${{ inputs.build-type }} \
+        ./target/release/solx-dev llvm build --build-type ${{ inputs.build-type }} --install-distribution \
           --ccache-variant=ccache ${ENABLE_UTILS} ${ENABLE_TESTS} ${ENABLE_VALGRIND} ${VALGRIND_OPTIONS} ${ENABLE_ASSERTIONS} ${ENABLE_COVERAGE} ${ENABLE_MLIR} ${SANITIZER} \
           --extra-args "-DCMAKE_EXPORT_COMPILE_COMMANDS='Off'" "-DLLVM_PARALLEL_LINK_JOBS='${LINK_JOBS}'" "-DLLVM_OPTIMIZED_TABLEGEN='On'"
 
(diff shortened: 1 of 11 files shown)


vladimirradosavljevic:
I'd rather add an `install-distribution` option and keep it disabled by default, since the CI is the only case where we intentionally do not need the full LLVM toolset.
I would also expect that anyone using this tool to build LLVM locally would want the full set of tools installed, so introducing a separate option for installing all tools would mostly add unnecessary distraction/complexity.
Also, if --enable-tests is passed, we must install all tools regardless, as the test suite depends on them.

hedgar2017:
> I'd rather add an `install-distribution` option and keep it disabled by default, since the CI is the only case where we intentionally do not need the full LLVM toolset. I would also expect that anyone using this tool to build LLVM locally would want the full set of tools installed, so introducing a separate option for installing all tools would mostly add unnecessary distraction/complexity. Also, if --enable-tests is passed, we must install all tools regardless, as the test suite depends on them.

Thanks!

Should I just remove `--enable-tools` completely then?

vladimirradosavljevic:
> > I'd rather add an `install-distribution` option and keep it disabled by default, since the CI is the only case where we intentionally do not need the full LLVM toolset. I would also expect that anyone using this tool to build LLVM locally would want the full set of tools installed, so introducing a separate option for installing all tools would mostly add unnecessary distraction/complexity. Also, if --enable-tests is passed, we must install all tools regardless, as the test suite depends on them.
> 
> Thanks!
> 
> Should I just remove `--enable-tools` completely then?

Yeah, as I don't think we will need it when we introduce `install-distribution` option.

hedgar2017:
@vladimirradosavljevic done, please check!

vladimirradosavljevic (approved):
Works like a charm. Thank you!

.github/actions/build-llvm/action.yml
```diff
@@ -202,7 +202,7 @@ runs:
           LINK_JOBS=1
         fi
 
-        ./target/release/solx-dev llvm build --build-type ${{ inputs.build-type }} \
+        ./target/release/solx-dev llvm build --build-type ${{ inputs.build-type }} --install-distribution \
```
vladimirradosavljevic, line 205: solx-llvm uses this, right? If that is correct, then we will have to somehow enable tests in that repo.

hedgar2017: @vladimirradosavljevic the flag [is canceled](https://github.com/NomicFoundation/solx/blob/db87d097f1f918ae3cc209311cbbb40a980e9c9b/solx-dev/src/llvm/mod.rs#L55) by `--enable-tests`, so should be fine!

PR #442, opened 2026-05-14
fix(tester): pass debug output dir via SOLX_OUTPUT_DIR env var

hedgar2017:
Commit 367fc1c removed the `--debug-output-dir` solx CLI flag in favor of the `SOLX_OUTPUT_DIR` env var, but missed updating the two `run_solx` subprocess-invocation sites in solx-tester. As a result, passing `-D` / `--debug` to solx-tester now fails immediately with `error: unexpected argument '--debug-output-dir' found`.

This forwards the per-mode debug directory via the env var instead, restoring the IR-dump behavior that `-D` is supposed to trigger.


vladimirradosavljevic (approved):
LGTM, thanks!

PR #453, opened 2026-05-20
refactor(slang): Migrate to `slang_solidity_v2`

ggiraldez:
This PR changes the dependency on Slang to `slang_solidity_v2` pinned to a recent [`main` commit](https://github.com/NomicFoundation/slang/commit/bf34fec7369266c29753a54aecf5b221a1f72e47).

Doing the migration I detected some issues still pending with the public API of Slang:
- ~~Exposing `DiagnosticExtensions` to avoid having to depend on `slang_solidity_v2_common`.~~
- ~~Providing the latest Solidity version supported by Slang to replace `LanguageFacts::LATEST_VERSION`.~~
- Having `LiteralKind::Integer` expose the minimum number of bits for representation.
Please let me know anything I missed.

I will send further PRs as these issues are fixed in Slang.

Another issue I'm not sure about is: the conversion of `NodeId` to `usize` is no longer possible, which forces depending on `slang_solidity_v2` for the function registry. We intend for `NodeId` to be opaque in the Slang API, so the extra dependency seems fine. But on the other hand this happens in `solx-mlir` which I presume should remain independent of Slang. Thoughts?


ggiraldez:
> Another issue I'm not sure about is: the conversion of `NodeId` to `usize` is no longer possible, which forces depending on `slang_solidity_v2` for the function registry. We intend for `NodeId` to be opaque in the Slang API, so the extra dependency seems fine. But on the other hand this happens in `solx-mlir` which I presume should remain independent of Slang. Thoughts?

We briefly discussed this offline and it seems sensible that `NodeId` remains abstract. A possibility we discussed is making `solx-mlir`'s struct generic over the identifier of functions. Happy to update this PR to go that way if you think it's appropriate.

ggiraldez:
Updated the PR to the latest `slang_solidity_v2` commit which solves a couple of the issues found. Also, since now Slang returns `U256` for slot addresses, updated all slot references to the new type.

solx-slang/src/ast/contract/function/expression/call/type_conversion.rs
```diff
@@ -27,6 +29,15 @@ pub enum TypeConversion<'context> {
     Cast(Type<'context>),
 }
 
+fn integer_bits_required(value: &BigInt) -> u32 {
```
hedgar2017: I'm trying to avoid dropping free functions and utils to maintain codebase coherence.

IIRC it's №3 from the PR description, so we can probably leave it here with a TODO about moving to Slang!

hedgar2017 (commented):
Looks really nice, thanks!

Just one issue with a util functions that stands out.

I'll check if all the same tests are passing, then approve!

solx-slang/src/ast/contract/function/expression/call/type_conversion.rs
ggiraldez: > I'm trying to avoid dropping free functions and utils to maintain codebase coherence.

Good to know, makes sense!

> IIRC it's №3 from the PR description, so we can probably leave it here with a TODO about moving to Slang!

Yes exactly, it's item 3. After we merge nomicFoundation/slang#1793 we will have `LiteralType::mobile_type()` in the API which should replace the literal handling branch in `TypeConversion`. In the mean time, I'll move this function into `TypeConversion` and leave a `TODO`.

hedgar2017 (approved):
Merging now, thanks!

I've found some ~37 regressions, but only a one-line fix was needed! Also I've unified new Cargo.toml dep entries.

PR #459, opened 2026-05-28
chore(slang): bump slang_solidity_v2 and fix hex literal lowering

hedgar2017:
Bumps `slang_solidity_v2` to a rev that includes the fix in NomicFoundation/slang#1801, which restores `HexNumberExpression::integer_value()` to parsing the literal text instead of routing through binder typing. Without it, hex literals in address-returning positions panic during lowering with `hex literals always evaluate to integers` — `solx-mlir/tests/lit/literal_types.sol` regressed when the v2 migration landed but the failure was masked because the migration PR did not have the `ci:slang` label.

Adapts to the v2 API changes that landed on slang main since the previous pin:

- `ContractDefinition::compute_abi_with_file_id(file_id)` is gone; the replacement `compute_abi()` derives the file from the contract node.
- `CompilationUnit::get_file_ast_root(file_id)` is gone; the replacement flow is `CompilationUnit::file(file_id)` then `File::ast()`.

The slang rev pin is temporary until NomicFoundation/slang#1801 lands on slang main; a follow-up will move it to the merged commit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -215,7 +215,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ggiraldez (approved):
LGTM! 🚀

PR #460, opened 2026-05-28
feat(slang): emit every contract in a source unit

hedgar2017:
Completes source-unit emission, lowering every contract a file deploys rather than only the first:

- `contract C { … }` at file scope lowers to its own `sol.contract` module, one module per contract as solc emits them, off a single melior context shared by the file; the standard-JSON output keys each by contract name, so the tester's source-order resolution finds the real deploy target instead of whichever contract happened to come first.
- `abstract contract A { … }` and `interface I { … }` lower to nothing, the modules `solc --mlir-action=print-init` also omits; `contract D is C { … }` is skipped alongside them, since a contract base carries state and function bodies to inherit while an interface base carries neither.

### DIFF
--- a/solx-mlir/tests/lit/module.sol
+++ b/solx-mlir/tests/lit/module.sol
@@ -11,8 +11,40 @@
 // CHECK-NEXT:       sol.return %{{.*}} : ui256
 // CHECK:        } {kind = #Contract}
 
+// CHECK:      module attributes {llvm.data_layout
+// CHECK:        sol.contract @{{.*Impl.*}} {
+// CHECK:          sol.func @{{.*h.*}}() -> ui256
+// CHECK:        } {kind = #Contract}
+
+// CHECK:      module attributes {llvm.data_layout
+// CHECK:        sol.contract @{{.*Second.*}} {
+// CHECK:          sol.func @{{.*g.*}}() -> ui256
+// CHECK:        } {kind = #Contract}
+
+// CHECK-NOT:  sol.contract
+
 contract C {
     function f() public pure returns (uint256) {
         return 42;
     }
+}
+
+interface Iface {
+    function h() external pure returns (uint256);
+}
+
+contract Impl is Iface {
+    function h() external pure returns (uint256) {
+        return 1;
+    }
+}
+
+contract Second {
+    function g() public pure returns (uint256) {
+        return 7;
+    }
+}
+
+abstract contract Undeployable {
+    function h() public pure virtual returns (uint256);
 }
--- a/solx-mlir/tests/lit/module.sol
+++ b/solx-mlir/tests/lit/module.sol
@@ -11,12 +11,12 @@
 // CHECK-NEXT:       sol.return %{{.*}} : ui256
 // CHECK:        } {kind = #Contract}
 
-// CHECK:      module attributes {llvm.data_layout
+// CHECK:      module attributes {llvm.data_layout = "E-p:256:256-i256:256:256-S256-a:256:256", llvm.target_triple = "evm-unknown-unknown"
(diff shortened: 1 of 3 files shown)


abinavpp (commented):
can we add a lit test with 2 contracts?

solx-slang/src/slang/mod.rs
```diff
@@ -164,43 +164,47 @@ impl Frontend for Slang {
...
+                        contract_name,
+                        solx_standard_json::output::contract::Contract {
+                            mlir: Some(mlir_stages),
+                            evm: Some(solx_standard_json::output::contract::evm::EVM {
+                                method_identifiers: Some(method_identifiers),
+                                ..Default::default()
+                            }),
+                            ..Default::default()
+                        },
+                    );
```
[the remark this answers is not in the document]
hedgar2017: Out of scope for now.

hedgar2017:
@abinavpp added. `solx-mlir/tests/lit/module.sol` now covers a file holding five definitions and pins all four outcomes under both RUN lines: `C` and `Second` each emit their own module, `contract Impl is Iface` emits (an interface base carries no state or bodies to inherit), and `interface Iface` / `abstract contract Undeployable` emit nothing.

The trailing `CHECK-NOT: sol.contract` is what pins the negative case; it is placed last so its window runs to end of input, and `Undeployable` is named and ordered to sort last under both the source order `solc --mlir-action=print-init` prints in and the alphabetical order solx keys its output map by. Verified in both directions: making that contract deployable fails the fixture under either oracle.

solx-mlir/tests/lit/module.sol
```diff
@@ -11,8 +11,40 @@
 // CHECK-NEXT:       sol.return %{{.*}} : ui256
 // CHECK:        } {kind = #Contract}
 
+// CHECK:      module attributes {llvm.data_layout
```
PavelKopyl: It looks like the string is cut. I guess it should be "{llvm.data_layout = "E-p:256:256-i256:256:256-S256-a:256:256", llvm.target_triple = "evm-unknown-unknown", as in the check 10 lines above.

PavelKopyl (approved):
LGTM.

solx-slang/src/source_unit.rs
```diff
@@ -18,37 +19,44 @@ use solx_utils::EVMVersion;
...
 
 impl<'context> SourceUnitScope<'context> {
-    /// Lowers the unit's contracts, owning the per-file melior scope, into standard-JSON contract
-    /// outputs keyed by contract name. Only the first contract per file compiles through this path
-    /// today; the rest are skipped until inheritance-aware emission lands.
+    /// Lowers every contract the unit deploys into standard-JSON contract outputs keyed by contract
+    /// name, each in its own MLIR module off the file's melior context, as solc emits one module per
+    /// contract. An abstract contract and an interface deploy nothing and produce no module, and a
+    /// contract with a contract base is skipped: an interface base carries no state or bodies to
+    /// inherit, a contract base does.
```
PavelKopyl: To be hones the last sentence in the comment is a bit unclear: I'd expand it a to: 
"Abstract contracts and interfaces are not deployable, so they produce no module. A contract with a contract base is skipped because emission (currently) collects only the contract's own state and functions, and a contract base (unlike an interface base) contributes inherited state and bodies we would silently drop." Just a suggestion)

solx-mlir/tests/lit/module.sol
hedgar2017: Fixed!

solx-slang/src/source_unit.rs
hedgar2017: Fixed!

PR #461, opened 2026-05-28
feat(slang): contract creation and try/catch

hedgar2017:
Lowers contract creation and the `try`/`catch` guarding it to their Sol ops:

- **Contract creation** — `new C(args)` lowers to `sol.new`, its options evaluated before its constructor arguments, the order legacy dispatches a creation in; `{value: v}` and `{salt: s}` ride the op's own operands, an omitted `value` materializing `sol.constant 0` at the op, and named arguments ordering by the constructor's parameters. A creation ABI-encodes its constructor arguments the way an external dispatch does, so a reference-typed argument is encoded out of the location it already lives in and only a scalar converts. A parenthesized callee (`(new C)(x)`, `(i.f{value: v})(y)`) peels to the same emission, carrying whatever options the layers hold.
- **Dynamic allocation** — `new T[](n)`, `new bytes(n)` and `new string(n)` lower to `sol.malloc` with `zero_init` and the size in its declared type, which the lowering masks against, so widening it here would change that cleanup; naming a contract without calling it (`new C;`) denotes its creation function and emits nothing.
- **`try`/`catch`** — the guarded call takes the arm an ordinary call takes with the op's `try_call` flag set, keeping the status the unguarded path drops, and `sol.try` opens one region per declared clause: `catch Panic(uint256)` binds `ui256`, `catch Error(string memory)` and `catch (bytes memory)` bind `!sol.string<Memory>`, and an omitted clause leaves its region blockless, which forwards an unmatched revert instead of swallowing it. A guarded creation carries no status of its own, so one comes from comparing the address `sol.new` produced against zero, since `CREATE` reports failure by returning zero.
- **Cross-object linking** — each code segment's dependencies are derived from the objects its own module names through `evm.dataoffset` / `evm.datasize`, so a creation in a constructor and one in a runtime function land in different lists. `Dependencies` moves to `solx-utils` and owns the rule that puts a segment's runtime child first, replacing three separate spellings of it — two of which compared against a fallback that made every non-suffixed dependency read as the runtime child, and survived only because the Yul and EVM assembly pipelines happen to encounter that child last. The MLIR pipeline encounters it first, so the rule is now pinned by unit tests at its owner rather than inferred from encounter order.
- **Object identity** — a contract's object takes a `<path>:<name>` identifier, which linking keys objects by, so two files declaring the same contract name no longer collapse into one object. The runtime segment's suffix stays `_deployed`, which is the Yul and Sol-to-LLVM convention; EVM legacy assembly and LLVM IR use `.runtime`, and `Dependencies` recognizes both.

Out of scope, each waiting on a construct emitted elsewhere:

- **Creation of a contract with a contract base** — `new D()` where `D` inherits names an object whose emission needs inherited state and bodies, so the reference reaches the linker unresolved rather than failing in the frontend.
- **A guarded public getter** — `try t.stateVariable()` needs the accessor a `public` state variable synthesizes, which no path emits; the unguarded call reports the same.
- **Library calls and the bare library name** — `L.f(x)` needs the `delegatecall` dispatch and the `library_call` ABI that hands a storage argument its slot rather than encoding it, and `address(L)` needs the library's own object behind the link symbol; both panic instead.
- **`type(C).creationCode`** — an object's bytecode as a value is a member access on a type expression, which is not lowered.


abinavpp (commented):
how exactly will we track the dependencies? every time we visit the `new`, `creationCode`, `runtimeCode`? I think it'll be better to combine supports for them here so that we can review how this change will enable those lowering.

solx-mlir/src/context/mod.rs
```diff
@@ -233,40 +240,25 @@ impl<'context> Context<'context> {
...
-        let runtime_llvm = Self::take_nested_module_text(&mut module, runtime_code_identifier)?;
+        let runtime_code_identifier = format!(
+            "{code_identifier}{}",
+            solx_utils::Dependencies::DEPLOYED_OBJECT_SUFFIX
+        );
+        let (runtime_llvm, runtime_dependencies) =
+            Self::take_nested_module(&mut module, runtime_code_identifier.as_str())?;
+        let deploy_dependencies =
+            Self::object_dependencies(&module.as_operation(), code_identifier);
         let deploy_llvm = module.as_operation().to_string();
```
abinavpp, line 251: Unrelated to this PR, but it would be nice if we can avoid this serialization.

solx-mlir/src/context/mod.rs
```diff
@@ -305,4 +297,77 @@ impl<'context> Context<'context> {
...
+            operation.remove_from_parent();
+            drop(unsafe { Operation::from_raw(operation.to_raw()) });
+
+            Some((text, dependencies))
+        })
+        .ok_or_else(|| anyhow::anyhow!("no module with sym_name `{target}` in Sol pass output"))
+    }
+
+    /// The objects `operation`'s code references, read off the intrinsics naming them.
+    fn object_dependencies<'c: 'a, 'a>(
```
abinavpp, line 337: Do we have a GitHub issue tracking sema-level contract-dependency analysis? We need it anyway for the circular-bytecode-reference error.  The cached sema info can be used here instead. Maybe we can add a todo/comment in the issue.

solx-utils/src/dependencies.rs
```diff
@@ -0,0 +1,111 @@
...
+        } else {
+            self.inner.push(dependency);
+        }
+    }
+
+    ///
+    /// Whether `dependency` is this object's own runtime segment, named by either the Yul and
+    /// Sol-to-LLVM suffix or the EVM legacy assembly one.
+    ///
+    fn is_runtime_child(&self, dependency: &str) -> bool {
```
abinavpp, line 53: The runtime child is still recognized by name convention. Every site that builds a `Dependencies` already knows its runtime child's exact identifier, so it could live in a dedicated field (like solc's `UnlinkedObj.runtimePart`)

abinavpp (approved):
mostly nits, otherwise lgtm!

solx-mlir/src/context/mod.rs
hedgar2017: Good idea, added an issue.

solx-mlir/src/context/mod.rs
hedgar2017: Yes, I've been thinking how to do it, created an issue.

solx-utils/src/dependencies.rs
hedgar2017: Yeah, it's used by all 3 pipelines for some historical reason. Let me try in a follow-up PR.

solx-mlir/src/context/mod.rs
hedgar2017: @abinavpp this serialization is "reused" as a means of serializing into process. I think it's reasonable to have it gone away together with the subprocesses when it becomes possible.

PR #462, opened 2026-05-28
feat(slang): external calls and ABI

hedgar2017:
Lowers every external call form and the `abi.*` family to their Sol ops:

- **External method calls** — `i.f(x)` lowers to `sol.ext_call`, dispatched on the callee's ABI selector. `static_call` is taken from the callee's own mutability rather than the call syntax, because that is where solc reads it, and the kind of dispatch from its visibility rather than from a probed value. The status is dropped: a failed call reverts, so only the callee's results reach the expression.
- **Bare calls** — `address.call`, `.delegatecall`, `.staticcall` lower to the matching op, their receiver crossing to a plain address and their `(status, ret_data)` pair re-typing member emission to a value list.
- **Call options** — `{gas: .., value: ..}` resolve once, in the source order solc evaluates them in, after the receiver and before the arguments, the order legacy solc dispatches a call in. An option the call omits materializes its default at the op, which is where solc materializes it: `{value: v}` alone emits `sol.gasleft` after the argument, `{gas: g}` alone emits `sol.constant 0` after the input.
- **External function type** — `!sol.ext_func_ref` through `solx_utils::FunctionReferenceKind`, beside `DataLocation`. A bare reference to a `public` function stays an internal pointer; only `this.f` externalizes, which slang's own `externalize_function_type` decides and the oracle confirms. Uncalled, `i.f` / `this.f` is `sol.ext_func_constant` with `.selector` and `.address` as its accessors; called, it is `sol.ext_icall`, whose arguments evaluate before the callee.
- **`abi.*` encoding and decoding** — `sol.encode` and `sol.decode`, folding a literal signature and an `encodeCall` callee to their selector constants, reading a storage payload through memory, encoding a reference-typed argument out of the data location it already lives in, and accepting the untupled single argument `abi.encodeCall` admits.
- **Contract casts** — no code at all. `Value::convert` already routes an address to a contract and back.


solx-slang/src/contract/function/expression/call/mod.rs
```diff
@@ -298,74 +326,188 @@ impl Call {
...
+            scope.contract.source_unit.function_type(&function_type),
+        );
+        let converted = scope.converted_arguments(arguments, &callee.function_type.parameters);
+        let selector = Value::selector(
+            function_definition
+                .compute_selector()
+                .expect("classification admits only callees with a selector"),
+            MlirType::field(scope.melior),
+            scope,
+        );
```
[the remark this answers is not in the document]
hedgar2017: The selector width differs per op by design, so `fixedbytes<4>` here would fail ODS verification rather than avoid a mismatch:

```tablegen
Sol_ExtCallOp:         UI256:$selector
Sol_EncodeOp:          Optional<Sol_FixedBytes4SelectorType>:$selector
Sol_ExtFuncSelectorOp: results (outs Sol_FixedBytes4SelectorType:$result)
```

`sol.ext_call` takes a word-sized selector; only `sol.encode` and `sol.ext_func_selector` are `bytes4`. solc emits the same thing for this op:

```mlir
%c3017696395_ui256 = sol.constant 3017696395 : ui256
sol.ext_call "f_7"(%6) at %5 gas %3 value %2 selector %c3017696395_ui256 {callee_type = (ui256) -> ui256} : !sol.address, (ui256) -> (i1, ui256)
```

That is why `Value::selector` takes the result type from the caller rather than fixing one: this site passes `field`, the `abi.*` sites pass the `bytes4` selector type, and each matches what its op declares.

solx-slang/src/contract/function/expression/member.rs
```diff
@@ -104,4 +132,22 @@ impl<'contract, 'source_unit, 'context> FunctionScope<'contract, 'source_unit, '
             element_type,
         )
     }
+
+    /// The ABI selector `.selector` yields: an operand naming a function folds to its selector
+    /// constant, while a pointer value carries its own.
+    pub fn external_selector(&mut self, operand: &Expression) -> Value<'context> {
+        if let Expression::MemberAccessExpression(access) = operand
```
abinavpp, line 139: iirc, this fold drops the base expression's side effects: `get().f.selector` never evaluates `get()`. Can you try this?

```solidity
contract C {
    uint256 counter;

    function get() internal returns (C) {
        counter += 1;
        return this;
    }

    function selectorOf() public returns (bytes4) {
        return get().selectorOf.selector;
    }
}
```

solx-slang/src/contract/function/expression/call/mod.rs
```diff
@@ -275,115 +345,261 @@ impl Call {
...
+                        .iter()
+                        .map(|element_type| scope.resolve_type(element_type, None))
+                        .collect(),
+                    other => vec![scope.resolve_type(&other, None)],
+                };
+                let decoded_types: Vec<MlirType<'context>> = result_types
+                    .iter()
+                    .map(|&result_type| {
+                        if result_type.is_address() {
+                            MlirType::address(scope.melior, true)
```
abinavpp: Payability is irrelevant to `sol.decode`'s lowering: address-like types decode through one branch (160-bit mask/validation), payable or not. Or did I miss something?

hedgar2017: I think it should be irrelevant, but solc is doing something weird here:
https://github.com/argotorg/solidity/blob/12aaca16458861e9b622818d49a82c1a7026594e/libsolidity/analysis/TypeChecker.cpp#L574-L576
solc MLIR inherits it, so I've been trying to follow it. Do you know why do we need `<payable>` in the dialect at all? If we need to keep it, we'll have to mirror that solc rule in Slang.

solx-slang/src/contract/function/expression/member.rs
hedgar2017: Thanks, confirmed! Let me see what I can do.

hedgar2017: Fixed, now side-effects are emitted in the same cases solc legacy does. Added tests as well.

solx-slang/src/contract/function/expression/call/mod.rs
hedgar2017: Removed these blocks and added [a Slang commit](https://github.com/NomicFoundation/slang/commit/0be9806fecd4e352dc14d849935e06dbc969a54c) to my temporary branch.
We'll decide if we need to add it to the binder or simply discard as solc compatibility trash.

abinavpp (approved):
lgtm, thank you!

PR #463, opened 2026-05-28
feat(slang): state-variable getters, custom errors, and type metadata

hedgar2017:
Lowers the contract-surface constructs through the Slang→MLIR pipeline: public state-variable getters, custom-error `require`, contract-namespace access, and `type(..)` metadata.

- **Public state-variable getters**: the `getter/` module synthesizes a `sol.func` per public variable off slang's getter type — scalar and reference reads, keyed mapping/array levels (`sol.map` / `sol.gep no_panic_bounds`), struct leaves with non-returnable members skipped, and constant initializer folds; immutable and transient variables are loud `unimplemented!` arms.
- **External getter dispatch**: `c.publicVar(keys)` and `try` over a getter classify through the new `ExternalCallee`, which admits functions and public state variables alike and dispatches getters as static calls.
- **Custom-error `require`**: `require(cond, E(..))` lowers through `sol.require` with the error's signature, bare and namespace-qualified alike.
- **Namespace member access**: a contract namespace delegates value and place access to its members (`C.x = v`, `Library.CONST`).
- **`type(..)` metadata**: integer `min`/`max`, `name`, `interfaceId`, and `creationCode`/`runtimeCode` through the new `sol.object_code` wrapper.
- **Selector folds**: error `.selector` (4-byte) and event `.selector` (32-byte topic) fold to constants, as does `.selector` on a getter reference.

**Tests**: 13 consolidated LIT fixtures (8 dual-oracle against `solc --mlir-action=print-init`, 5 solx-only); full tester goes 8792 → 9468 passing with zero regressed files.


solx-slang/src/contract/function/expression/call/mod.rs
```diff
@@ -227,12 +230,29 @@ impl Call {
...
                     );
                 }
-                if let Some(Definition::Function(function_definition)) =
-                    access.member().resolve_to_definition()
-                    && function_definition.compute_selector().is_some()
+                if let Expression::Identifier(operand) = access.operand()
+                    && matches!(
+                        operand.resolve_to_definition(),
+                        Some(Definition::Contract(_))
+                    )
```
PavelKopyl: I suggest to also handle a library case, something like, 
```
if let Expression::Identifier(operand) = access.operand()
    && matches!(operand.resolve_to_definition(), Some(Definition::Library(_)))
{
    unimplemented!("library calls")
}
```
Otherwise, the compiler should crash with no clear reason on:
```
library MathLib {
    function double(uint256 x) public pure returns (uint256) {
        return x * 2;
    }
}

contract C {
    function run(uint256 v) public pure returns (uint256) {
        return MathLib.double(v);
    }
}
```

solx-slang/src/contract/getter/leaf.rs
```diff
@@ -0,0 +1,78 @@
...
+    ) -> Vec<Value<'context>> {
+        match self {
+            Self::Value(stored_type) => {
+                vec![
+                    place
+                        .load(*stored_type, context)
+                        .convert(results[0], context),
+                ]
+            }
+            Self::Members(indices) => indices
```
PavelKopyl, line 54: It’s probably reasonable to assert that `results` and `indices` have the same length, at least. They are calculated in two different places: `indices` here and `results` in Slang, but the logic behind them should be consistent.

solx-slang/src/contract/getter/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// Replays the levels over the getter's entry arguments, then reads the leaf into the return
+    /// values converted to `results`.
+    pub fn returned_values(
+        &self,
+        mut place: Place<'context>,
+        results: &[MlirType<'context>],
+        entry: Block<'context>,
+        context: &Context<'context>,
+    ) -> Vec<Value<'context>> {
+        for (index, level) in self.levels.iter().enumerate() {
```
PavelKopyl, line 99: I guess we could also check that the lengths of `self.levels` and `entry.argument `match, as mentioned in leaf.rs.

solx-slang/src/contract/getter/leaf.rs
hedgar2017: @PavelKopyl we're deliberately not doing any assertions in codegen to preserve performance and not do something Slang is supposed to do. It would be inconsistent with all the previous work done here.

solx-slang/src/contract/getter/mod.rs
hedgar2017: Ditto.

solx-slang/src/contract/function/expression/call/mod.rs
hedgar2017: The library PR comes right after this one, but no big deal, I can add this branch.

solx-slang/src/contract/mod.rs
```diff
@@ -107,6 +111,13 @@ impl<'context> SourceUnitScope<'context> {
                 for function in node.functions().iter().chain(operator_functions) {
                     scope.function_definition(function);
                 }
+                for member in node.members().iter() {
+                    if let ContractMember::StateVariableDefinition(state_variable) = member
+                        && state_variable.is_externally_visible()
+                    {
+                        scope.state_variable_getter(&state_variable);
+                    }
+                }
```
PavelKopyl: Does it make sense first to collect public state variables and use the list here and below where we calculate ABI list?

hedgar2017: @PavelKopyl yes, good catch, thanks 🚀 
AST node is a shared reference so no heavy copying is done.

solx-slang/src/contract/function/expression/call/mod.rs
hedgar2017: Added!

solx-slang/src/contract/getter/leaf.rs
PavelKopyl: Sure, I do not mean checking this in release built, but as I understand, the fact that `results` and `indices` are synced is based on that logically there are two copies of the filer:
```
                    
                        Type::Mapping(_) | Type::Array(_) | Type::FixedSizeArray(_) => false,
                        Type::Function(function_type) => {
                            FunctionReferenceKind::from(function_type.visibility())
                                == FunctionReferenceKind::External
                        }
```
one here (members() method) and another one in slang. At the moment, everything works as expected and will presumably continue to do so. However, hypothetically, we could lose some indexes if they diverge at some point in the future. I was just thinking that it might be worth enforcing this in debug builds, but if there’s a guarantee that they are always kept in sync, then that’s fine.

PavelKopyl (approved):
LGTM, thanks!

solx-slang/src/contract/getter/leaf.rs
hedgar2017: @PavelKopyl good idea, let me check exactly where it happens!

PR #479, opened 2026-06-16
chore(slang): Update slang_solidity_v2 to latest

ggiraldez:
Makes code changes to handle AST API changes.

This should be used instead of #464, as there were some minor API changes. Please let me know if I should do differently and push those changes in the other PR.


hedgar2017 (approved):
Thank you sir!

Just please set `ci:slang` label for subsequent PRs because we're not running these tests by default yet.

PR #494, opened 2026-07-02
refactor(mlir): introduce the ir/ entity substrate and migrate solx-slang onto it

hedgar2017:
The first refactoring PR of the series: the main target is the `solx-mlir` crate.

1. Dissolves the huge god-object `Builder` and moves all dialect costruction to abstracted solx-mlir interface, keeping AST emission as high-level as possible, otherwise it would get dicey very soon!
2. Splits data ownership between `Value`, `Place`, `Type` (rustc-inspired), `Effect`, `Contract`, `Function` (these are minimal now but will evolve soon.
3. Updates all solx-slang call sites, non-invasively and behavior-preserving.
4. Changes all stringy attributes to ODS strictly-typed interface (uses a melior fork with a fix for now).
5. Bumps deps: solx-llvm, slang, melior (no breaking changes, so the latest API can be used everywhere).


solx-mlir/src/ir/value/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+            &BigInt::from(u8::from(value)),
+            Type::signless(context.mlir_context, solx_utils::BIT_LENGTH_BOOLEAN),
+            context,
+            block,
+        )
+    }
+
+    /// A `string memory` value holding `text` verbatim, via `sol.string_lit`.
+    /// The bytes are taken as-is; a string literal need not be valid UTF-8.
+    pub fn string_literal<B>(text: &str, context: &Context<'context>, block: &B) -> Self
```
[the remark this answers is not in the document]
hedgar2017: Fixed.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
```diff
@@ -35,39 +31,23 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
         let result_type = target_type.unwrap_or_else(|| {
-            let lhs_width = solx_mlir::TypeFactory::integer_bit_width(lhs.r#type());
-            let rhs_width = solx_mlir::TypeFactory::integer_bit_width(rhs.r#type());
+            let lhs_width = lhs.r#type().integer_bit_width();
+            let rhs_width = rhs.r#type().integer_bit_width();
             if lhs_width >= rhs_width {
                 lhs.r#type()
             } else {
                 rhs.r#type()
             }
```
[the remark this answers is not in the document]
hedgar2017: Changed to panic!

solx-mlir/src/context/function.rs
```diff
@@ -28,4 +48,96 @@ impl<'context> Function<'context> {
...
+        if let Some(selector_value) = selector {
+            operation_builder = operation_builder.selector(IntegerAttribute::new(
+                IntegerType::new(context.melior, Type::SELECTOR_BIT_WIDTH).into(),
+                selector_value as i64,
+            ));
+        }
+        if selector.is_some() || matches!(kind, Some(FunctionKind::Constructor)) {
+            operation_builder =
+                operation_builder.orig_fn_type(TypeAttribute::new(function_type.into()));
+        }
```
abinavpp, line 101: we also need to set $id attr for indirect internal function call. is that part of another pr?

solx-mlir/src/context/mod.rs
```diff
@@ -155,28 +122,33 @@ impl<'context> Context<'context> {
...
         Self {
+            melior,
             module,
             function_signatures: HashMap::new(),
-            builder: Builder::new(context),
             current_contract_type: None,
         }
     }
 
+    /// The unknown source location.
```
abinavpp, line 132: add a fixme/todo here. we should not generate any unknown loc moving forward

solx-mlir/src/ir/effect/control_flow.rs
```diff
@@ -0,0 +1,111 @@
...
+            ConditionOperation.condition(condition)
+        );
+    }
+
+    /// Emits `sol.if` and returns the then- and else-region entry blocks.
+    pub fn branch(
+        self,
+        condition: Value<'context, 'block>,
+    ) -> (BlockRef<'context, 'block>, BlockRef<'context, 'block>) {
+        mlir_region_op!(self.context, &self.block, IfOperation.cond(condition); then_region, else_region)
```
abinavpp: try not to generate an else block if there aren't any in the ast

solx-mlir/src/ir/type/array_size.rs
```diff
@@ -0,0 +1,20 @@
+//!
```
abinavpp, line 1: nit: do we need a separate file for this? why not in type/mod.rs?

solx-slang/src/ast/contract/function/expression/access.rs
```diff
@@ -81,34 +73,18 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
-                        base_value.r#type().to_raw(),
-                        0,
-                    ))
-                };
+                let element_type = base_value.r#type().element_type(0);
                 let address =
-                    self.state
-                        .builder
-                        .emit_sol_gep(base_value, index_value, element_type, &block);
+                    Place::from(base_value).gep(index_value, element_type, self.state, &block);
```
abinavpp, line 106: add sol.fixed_bytes_index op for fixed bytes too, or do we have it another pr?

solx-slang/src/ast/contract/function/expression/arithmetic.rs
```diff
@@ -97,65 +77,29 @@ impl<'state, 'context, 'block> ExpressionEmitter<'state, 'context, 'block> {
...
-                        .build()
-                        .into(),
-                    )
-                    .result(0)
-                    .expect("sol.sub always produces one result")
-                    .into();
+                let value = TypeConversion::from_target_type(operand_type, self.state)
+                    .emit(value, self.state, &block);
+                let zero = Value::constant(0, operand_type, self.state, &block);
+                let result = zero.subtract(value, false, self.state, &block);
```
abinavpp: this has to be checked sub! also, could we label (with a comment or something) this checked boolean arg for readability.

abinavpp (commented):
Thank you! I like the shape of this. IIRC, Value == RValue, Place == LValue and Effect == InsertPt (could rename to avoid mixup with mlir effects? not sure about the new name, cursor maybe?) is the right call. It would be nice if we can generate rust from mlir-tblgen to avoid all those ir/value/*.rs files. Just some nits/questions.

solx-mlir/src/context/function.rs
hedgar2017: Yes! These calls will have their dedicated PR.

solx-mlir/src/ir/type/array_size.rs
hedgar2017: This is by design and follows the convention of the rest of this repo. I'm deliberately limiting to 1 entity per file because where are 2 entities there are 3, and it turns into spaghetti 5k LoC modules very quickly.

Also would reward AI for spitting out slop wherever it wants :)

solx-slang/src/ast/contract/function/expression/access.rs
hedgar2017: Another PR, but it makes sense to add it here, will do.

solx-slang/src/ast/contract/function/expression/arithmetic.rs
hedgar2017: Thanks, will do!

solx-mlir/src/context/mod.rs
hedgar2017: Will add! It will wait for the debug info support :)

hedgar2017:
> Thank you! I like the shape of this. IIRC, Value == RValue, Place == LValue and Effect == InsertPt (could rename to avoid mixup with mlir effects? not sure about the new name, cursor maybe?) is the right call. It would be nice if we can generate rust from mlir-tblgen to avoid all those ir/value/*.rs files. Just some nits/questions.

Thanks for the thorough review @abinavpp !

1. Actually, `Effect` was renamed to `Block` in the next refactor PR: https://github.com/NomicFoundation/solx/pull/514. I'm happy do it here to reduce the total number of changes. Let me know if you like `Block` (see the other PR, it now owns both the block logic and these ops' emission).
2. It was my original plan, but I haven't done a lot of Rust macros before, so designing it upfront is not straightforward. The main challenge is conditional arguments, for which I probably need to generate their own wrapper-builder. I'm happy to try in order to nuke so much future diff surface, but I'm not confident it will accommodate everything we'll have in the final codebase. Let me try it anyway when my quota resets!

abinavpp (approved):
lgtm. except the unary minus lowering. thanks a lot!

solx-mlir/src/ir/effect/control_flow.rs
hedgar2017: Thanks, found an elegant way of expressing it in the macro!

solx-slang/src/ast/contract/function/expression/arithmetic.rs
hedgar2017: Done! Also added some missing tests for arithmetic ops (signed exp deferred to another PR).

PR #514, opened 2026-07-06
refactor(mlir): own the block insertion cursor in solx_mlir::Context

hedgar2017:
Move block management onto solx_mlir::Context: the insertion cursor (current_block) lives on the context, Effect folds into the Block entity, and the block-scoped 'block lifetime collapses into 'context. The frontend threads the cursor through &mut Context and drops its melior dependency, so control flow repositions the cursor instead of passing block handles.

It makes MLIR must more encapsulated and its client code significantly shorter (net ~600 lines removed).


abinavpp (approved):
thank you!

PR #523, opened 2026-07-09
refactor(mlir): reduce MLIR dialect boilerplate to macro definitions

hedgar2017:
Implements `macro_rules` that generates methods for all Value, Place, and Block-based MLIR operations we have implemented so far.


abinavpp (approved):
nice! thank you!

PR #524, opened 2026-07-09
perf(core): reuse persistent worker subprocesses across translation units

hedgar2017:
Replaces the fork/exec-per-translation-unit subprocess model with a pool of **persistent worker subprocesses** fed over pipes, and slims the per-unit IPC payload. Previously every translation unit fork/exec'd the 55 MB LLVM-linked binary, CBOR-encoded the whole payload, and each child built an unused N-thread rayon pool.

Local results with the raw openzeppelin standard JSON (average of 5 runs):
| Case | Before (`solx-old`) | After (`solx-new`) | Δ |
|---|---|---|---|
| dwarf | 20.94s | 19.55s | −6.6% |
| no-dwarf | 20.02s | 18.87s | −5.7% |


solx-codegen-evm/src/target_machine.rs
```diff
@@ -27,12 +27,21 @@ impl TargetMachine {
...
+    ///
     pub fn new(
         optimizer_settings: &OptimizerSettings,
         llvm_options: &[String],
     ) -> anyhow::Result<Self> {
-        let mut arguments = Vec::with_capacity(1 + llvm_options.len());
+        let mut arguments = Vec::with_capacity(4 + llvm_options.len());
         arguments.push(Self::TARGET.to_string());
+        if optimizer_settings.metadata_size.is_none() {
+            arguments.push("-evm-metadata-size=0".to_owned());
```
hedgar2017: Why are we adding these? What are the correctness implications? What if we keep it omitted like before?

solx-core/src/process/child.rs
```diff
@@ -0,0 +1,87 @@
...
+//! The subprocess-side worker: reads a session, then compiles jobs until `stdin` closes.
+//!
+
+use std::sync::atomic::Ordering;
+use std::thread::Builder;
+
+use crate::error::Error;
+use crate::project::contract::Contract;
+
+use super::channel::FrameRead;
```
hedgar2017: Replace all super:: with crate:: with full paths.

solx-core/src/process/pool.rs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            self.idle.lock().expect(POISON).push(worker);
+        }
+        Ok(result)
+    }
+
+    ///
+    /// Builds the contract error surfaced when a worker dies while compiling `contract_name`.
+    ///
+    fn worker_failed(
```
hedgar2017: Inline this.

solx-core/src/process/pool.rs
```diff
@@ -0,0 +1,102 @@
...
+        contract_name: &solx_utils::ContractName,
+        job: &Job,
+    ) -> crate::Result<EVMOutput> {
+        let idle = self.idle.lock().expect(POISON).pop();
+        if let Some(worker) = idle
+            && let Ok(result) = self.dispatch(worker, job)
+        {
+            return result;
+        }
+        self.dispatch(
```
hedgar2017: Why is it retrying once? Why not twice or zero times?

solx-core/src/process/session.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// The input contract language.
+    pub language: solx_standard_json::InputLanguage,
+    /// The `solc` compiler version, used only for Solidity and Yul projects.
+    pub solc_version: Option<solx_standard_json::Version>,
+    /// The EVM version to produce bytecode for.
+    pub evm_version: Option<solx_utils::EVMVersion>,
+    /// The mapping of auxiliary identifiers, e.g. Yul object names, to full contract paths.
+    pub identifier_paths: BTreeMap<String, String>,
+    /// Output selection for the compilation.
+    pub output_selection: solx_standard_json::InputSelection,
```
hedgar2017, line 19: Is it safe to rename this type to `OutputSelection`?

solx-core/src/process/worker.rs
```diff
@@ -0,0 +1,72 @@
+//!
+//! A persistent worker subprocess owned by the pool.
+//!
+
+use std::path::Path;
+use std::process::Child;
+use std::process::Command;
+use std::process::Stdio;
+
+use super::channel::FrameRead;
```
hedgar2017: super:: -> crate:: with full paths

solx/tests/cli/recursive_process.rs
```diff
@@ -22,11 +22,7 @@ fn missing_input() -> anyhow::Result<()> {
 fn excess_args() -> anyhow::Result<()> {
     crate::common::setup()?;
 
-    let args = &[
-        "--recursive-process",
-        crate::common::TEST_SOLIDITY_CONTRACT,
```
hedgar2017, line 27: This contract name was deliberately added to be able to view which contract is being compiled through OS process monitors. Is there a way to preserve this without hurting performance?

solx-codegen-evm/src/target_machine.rs
hedgar2017: Correctness, not a cosmetic default. LLVM cl-options are process-global, and a worker now compiles many units in one process, so an `-evm-metadata-size` set for one unit survives into the next. Omitting it was only safe under the old fork/exec-per-unit model, where every unit started at LLVM's default; here a unit with no metadata following one with metadata would inherit the stale size and bake a wrong metadata length into its bytecode — the EVMLA byte-identity check catches exactly this. So it is always emitted: `=0` before `llvm_options` when unset (user options still win), and the computed `={size}` after them when set (computed value wins). The stack-region options need no such healing — a worker is only returned to the idle pool when its job had no spill-area size, so a reused worker never set them.

solx-core/src/process/child.rs
hedgar2017: Done — replaced with full `crate::process::…` paths (same applied in `worker.rs` and `pool.rs`).

solx-core/src/process/pool.rs
hedgar2017: Done — inlined into `execute`; `worker_failed` removed.

solx-core/src/process/pool.rs
hedgar2017: One fresh retry is the minimum that separates the two failure modes. An idle worker can die for reasons unrelated to this job (killed externally, or it exited while parked), so its death must not be attributed to the contract — hence one attempt on a fresh process. A fresh worker is clean, so if it dies on the job the failure is genuine and deterministic for this input; retrying again would only reproduce the same crash. Zero retries would misattribute a stale-worker death to the contract. So: a best-effort idle attempt, then one authoritative fresh attempt.

solx-core/src/process/session.rs
hedgar2017: `solx_standard_json::InputSelection` is the shared solc-standard-JSON type — it models the `settings.outputSelection` field of the *Input*, hence the name — and is referenced across ~14 files. Renaming it to `OutputSelection` is a reasonable clarity refactor, but it is cross-cutting and unrelated to this PR, so I would keep it out of scope here. The field itself is already `output_selection`, which reads correctly.

solx-core/src/process/worker.rs
hedgar2017: Done — full `crate::process::…` paths.

solx/tests/cli/recursive_process.rs
hedgar2017: Yes, and cheaply. The worker holds each `Job` (which carries `contract_name`), so it can set its own process title to the current contract at the top of the job loop — one syscall per translation unit, negligible next to the compile. On Linux that is `prctl(PR_SET_NAME)` via the `libc` we already depend on (visible in `ps`/`top`), or a small crate like `proctitle` for the full path in `ps aux` cross-platform. I left it out of this commit as a separable follow-up, since it adds either platform-specific unsafe or a dependency — happy to fold it in here if you would prefer.

solx-core/src/process/channel.rs
```diff
@@ -0,0 +1,61 @@
+//!
+//! The length-prefixed CBOR frame protocol shared by both sides of the worker channel.
+//!
+
+use std::io::Read;
+use std::io::Write;
+
+/// The size of the frame length prefix.
+const LENGTH_PREFIX_SIZE: usize = size_of::<usize>();
```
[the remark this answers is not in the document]
hedgar2017: `size_of` has been in the prelude since Rust 1.80 (`core::mem::size_of`); this crate is edition 2024 on toolchain 1.96, so it compiles as written and `cargo check`/`clippy` are green. Keeping the prelude form.

solx-core/src/process/channel.rs
```diff
@@ -0,0 +1,61 @@
...
+    fn recv<T>(&mut self) -> anyhow::Result<Option<T>>
+    where
+        T: serde::de::DeserializeOwned,
+    {
+        let mut length_bytes = [0u8; LENGTH_PREFIX_SIZE];
+        match self.read_exact(length_bytes.as_mut_slice()) {
+            Ok(()) => {}
+            Err(error) if error.kind() == std::io::ErrorKind::UnexpectedEof => return Ok(None),
+            Err(error) => anyhow::bail!("Frame length prefix reading error: {error}"),
+        }
```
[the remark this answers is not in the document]
hedgar2017: Fixed. `recv` now takes `BufRead` and checks `fill_buf()` first: an empty buffer is a clean shutdown at a frame boundary → `Ok(None)`, while a short read once a frame has started goes through `read_exact` and surfaces as a truncation error. The worker's stdout is wrapped in a `BufReader` (which also cuts read syscalls); the child side already reads from a `StdinLock`.

Cargo.toml
```diff
@@ -87,7 +87,7 @@ features = ["http-rustls-tls", "test", "signing"]
 # LLVM
 [workspace.dependencies.inkwell]
 git = "https://github.com/NomicFoundation/inkwell"
-rev = "4f9f86e15f43dc8555b8b51d4dda01e60f756f22"
+branch = "az-reset-option-occurrences"
```
nebasuke: I think we should first create a PR for this and review that, rather than pin to a branch.

solx-llvm
nebasuke, line 1: I would suggest doing the solx-llvm bump separately on main to not pollute any other performance checks. Also makes it easier to revert if needed.

nebasuke (changes requested):

nebasuke:
Not my wheelhouse, so feel free to discard if irrelevant, but here's a drive-by Claude comment. 

# Claude comment
Non-blocking, follow-up material: `Pool::execute` drops the worker on any `Err`, including ordinary per-contract compile errors. In that case the child is actually at a clean frame boundary (the worker loop sends the `Err` frame and keeps looping), so we discard a healthy worker and every failing contract re-spawns the 55 MB binary. A cheap refinement would be to return the worker to the pool on job-level `Err` results and only drop it on channel/protocol errors — with the caveat that `StackTooDeep` must keep dropping, since that child exits after replying.

Related, minor: a crashed worker now surfaces as "The worker closed the response channel" without the exit status the old `wait_with_output` path reported (stderr still reaches the terminal via `Stdio::inherit`). Calling `wait()` on that path and including the exit code in the error would recover the old diagnosability.

Cargo.toml
hedgar2017: Never planned to merge it of course. Will create after LLVM folks confirm [this is the right approach](https://github.com/NomicFoundation/solx-llvm/pull/113).

solx-llvm
hedgar2017: I think LLVM has only MLIR dialect updates now, but makes sense, let me create a separate bump.

hedgar2017:
> Not my wheelhouse, so feel free to discard if irrelevant, but here's a drive-by Claude comment.

Thanks, both sound reasonable, fixed!

solx-core/src/process/channel.rs
```diff
@@ -0,0 +1,68 @@
...
+            .is_empty()
+        {
+            return Ok(None);
+        }
+        let mut length_bytes = [0u8; LENGTH_PREFIX_SIZE];
+        self.read_exact(length_bytes.as_mut_slice())
+            .map_err(|error| anyhow::anyhow!("Frame length prefix reading error: {error}"))?;
+        let mut body = vec![0u8; usize::from_le_bytes(length_bytes)];
+        self.read_exact(body.as_mut_slice())
+            .map_err(|error| anyhow::anyhow!("Frame body reading error: {error}"))?;
```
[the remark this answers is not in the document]
hedgar2017: This endpoint is in-process only, so not a real-life use case.

nebasuke:
## macOS validation results

Following up on the concern that persistent-worker changes can regress badly on macOS while looking fine on Linux: I ran this PR's tester suite on macOS runners via a debug harness (#546 — direct `solx-tester` invocation, symbols kept, hang watchdog with `sample`, crash reports uploaded). Summary: **the current tip is green on macOS arm64 and consistently *faster* than main; on x86-64 it is functional with no sign of the hang (a toolchain bug unrelated to this PR required a workaround, and the runner is too slow to complete the suite); the hang we initially observed was real, but belonged to the intermediate state `6bd1976a` and is fixed by the two commits that followed.**

| Code | Runner | Result |
|---|---|---|
| main (baseline) | macos-15 | ✅ 15m33s ([run](https://github.com/NomicFoundation/solx/actions/runs/29364158123)) |
| main (baseline) | macos-15-xlarge | ✅ 5m31s (same run) |
| **#524 @ 1ae05b1d (current)** | macos-15 | ✅ **10m32s** ([run](https://github.com/NomicFoundation/solx/actions/runs/29374444399)) |
| **#524 @ 1ae05b1d (current)** | macos-15-xlarge | ✅ **4m31s** ([run](https://github.com/NomicFoundation/solx/actions/runs/29399344551)) |
| **#524 @ 1ae05b1d (current)** | macos-15-large (x86-64) | ✅ functional — no crash, no hang; suite too slow for the runner to finish ([run](https://github.com/NomicFoundation/solx/actions/runs/29405063434), details below) |
| #524 @ 6bd1976a (old tip) | macos-15-xlarge | 🐌 4h26m silent hang, job timeout ([#545 run](https://github.com/NomicFoundation/solx/actions/runs/29350066312)) |
| #524 @ 6bd1976a (old tip) | macos-15 | 🐌 hang reproduced, watchdog kill at 30 min ([run](https://github.com/NomicFoundation/solx/actions/runs/29401155956), samples in the run artifact) |

### Anatomy of the (already-fixed) hang

`sample` stacks from the old-tip reproduction:

- every solx-tester thread: `Child::wait_with_output → poll`, waiting on top-level `solx` processes that never exit;
- every sampled worker subprocess (`--recursive-process`): idle in `FrameRead::recv → read(stdin)`, waiting for a job frame that never arrives.

So the wedge is inside the top-level `solx` processes, between receiving translation units and dispatching to their (ready) workers. `6bd1976a` resets LLVM cl-options per unit but predates the `solx-llvm` reset-option-occurrences pin (`1ae05b1d`), which is consistent with dispatch threads deadlocking in LLVM's option machinery — and with the hang disappearing exactly when `b8469e03` + the LLVM pin landed. It reproduces on the 3-core standard runner, so it wasn't an xlarge/VM artifact. I didn't capture the parents' own stacks (harness sampled only tester + workers), so the LLVM-side mechanism is inferred rather than observed; happy to grab parent stacks with one more free-runner run if you want it pinned down exactly.

### macOS x86-64

Getting an Intel datapoint required a workaround: solx-tester itself is killed on any Intel mac — including on main — by an unrelated rustc 1.95/1.96 enum-niche miscompilation (#547; fix is the upcoming 1.97.1 toolchain bump). So the harness rebuilt just the tester with nightly rustc while keeping solx on the pinned stable. With that:

- smoke test green; the full suite ran **~78 minutes without a crash or a wedge** before the watchdog killed it — at kill time the tester was mid-`posix_spawn` launching new solx invocations, with per-invocation lifetimes so short that `pgrep` caught zero solx processes in flight. That's the opposite signature of the old-tip hang (frozen solx parents, starved workers).
- the suite is simply too slow to *complete* on the 2012-era Intel minis GitHub provides (the 3-core arm runner needs 10–15 min; this 12-core box couldn't finish even the Ethereum corpus subset in 23 min on main). No wedge, but also no completed run or main-vs-PR timing comparison is attainable on this hardware — a limitation of the runner, not of this PR.

### Other notes

- Linux was green at both tips all along (this PR's own integration runs), so the old hang was genuinely macOS-specific — the platform concern was justified, just already addressed by the later commits.

#545 served its purpose as the first mac experiment and is superseded by the #546 harness; closing it.

nebasuke (approved):
Looks good from my end.

PR #525, opened 2026-07-09
chore: devcontainer on the CI runner image, with LLVM install docs and CI validation

nebasuke:
I've run this locally and resolved a whole bunch of UX issues (too much files getting checked by VSCode, interrupted builds struggling, Boost not working well, etc.).

Key change:
* Split the Docker container into a base, dev and a CI only part. The devcontainer then constructs a definition by importing dev (built on base). Note that this causes ~10 minutes of build time (instead of 5 min for ghcr.io) the first install due to `apt get` installations, but avoids having to access ghcr.io. This will be cached and avoided in subsequent runs (due to the separate dev image).  Cached startup was ~10 seconds locally within vscode.

# Claude summary

## What

Adds a `.devcontainer/` to solx plus a Developer Guide page, so opening the repo in VS Code (or Codespaces / the devcontainer CLI) lands you in the CI toolchain environment — with no registry authentication — and a documented one-command path to building the custom LLVM backend.

- **`base`/`dev`/`ci` stage split in `.github/docker/Dockerfile`** — `base` is the shared system toolchain (system packages, LLVM-21 apt, libm/libstdc++ fixes, Node); `dev` adds the devcontainer bits on top (UID-mapped non-root `dev` user with sudo, shared rustup at `/usr/local/{rustup,cargo}` with the toolchain resolved lazily from `rust-toolchain.toml`, apt valgrind — dev needs a working valgrind, not the `ci` stage's pinned from-source build, and keeping it in the image stops LLVM's baked-in cmake valgrind detection from flapping across rebuilds); `ci` adds the CI-only extras (the root-homed Rust install CI jobs use — pinned 1.96.0, kept in lockstep with `rust-toolchain.toml` by a new renovate customManager — plus valgrind from source and cargo coverage/report tools) and stays the final stage, so the published `solx-ci-runner` image is unchanged (`docker-ci-image.yaml` builds the default target, and buildx skips the `dev` stage when targeting `ci`).
- **The devcontainer builds `--target dev` directly** (`devcontainer.json` points at the runner Dockerfile) — ~5–10 min the first time, layer-cached afterwards. All sources are public, so the private `solx-ci-runner` package never needs to be pulled and there is no digest to keep in sync: the container tracks the checkout's Dockerfile by construction. Single build, single cache domain — an earlier iteration built a separately-tagged base via `initializeCommand`, which busted the child layer cache on every rebuild (containerd image store re-IDs images via provenance timestamps even on full cache hits).
- **`.devcontainer/devcontainer.json`** — named volumes for `target/`, `target-slang/`, `target-llvm/` (per-checkout via `${devcontainerId}`, so parallel checkouts/worktrees never share a build tree), plus shared rustup/cargo/ccache volumes (survive container rebuilds, stay off the slow bind mount on macOS/Windows; the rustup share's first-run concurrency caveat is documented in the guide); `hostRequirements` for the LLVM build (8 CPU / 16 GB / 64 GB); rust-analyzer + clangd extensions, clangd pointed at the `compile_commands.json` `solx-dev` exports into `target-llvm/build-final/`; VS Code watcher/search excludes for the build trees (without them the server crawls the built LLVM tree and integrated terminals lag by seconds).
- **`.devcontainer/bootstrap.sh`** — the explicit "install LLVM" step: branch-aware shallow submodule init (a submodule left on a branch — the fork-hacking flow — is left untouched with its nested submodules still synced; a detached pin is reset with an announcement) → build `solx-dev` → `solx-dev llvm build --enable-assertions --enable-mlir --ccache-variant ccache` → `solx-dev solc build --build-boost --ccache-variant ccache --enable-mlir` (the image ships no system Boost, matching CI's build-solc action). MLIR is on by default (required for `cargo test-slang`; `--no-mlir` to opt out); `--clean` wipes `target-llvm/` first, for build trees configured in a previous container image (CMake caches toolchain facts). `CCACHE_COMPILERCHECK=content` matches CI so image rebuilds with an unchanged clang don't cold-miss the cache. Deliberately *not* run from `postCreateCommand`: the cold build is ~1h and should be a step you choose, not a surprise. `post-create.sh` only does container plumbing and prints the pointer. (An earlier revision git-excluded the Boost trees `solx-dev` drops into the `solx-solidity` submodule; the fork's own `.gitignore` now covers them — see the submodule bump below.)
- **`validate-devcontainer` job in `test.yaml`** (mirroring slang's) — path-gated on `.devcontainer/**`, `.github/docker/**` (the devcontainer builds its base stage from it), `rust-toolchain.toml`, and `test.yaml` itself (so edits to the validation harness get validated too). Shellchecks the scripts, then builds the container via the pinned `devcontainers/ci` action — same auth-free local-build flow as a developer machine — and runs `.devcontainer/smoke-test.sh` inside: non-root user + sudo, writable workspace/volumes, system toolchain on PATH, lazy resolution of the `rust-toolchain.toml` pin. Deliberately does **not** build LLVM. The smoke script doubles as a user-facing health check (linked from the guide's troubleshooting section).
- **`docs/src/developer-guide/00-development-container.md`** — the grounding doc: what the base image contains vs. what is deliberately not baked in (the LLVM backend — per-submodule-bump churn + GB weight), how the LLVM install actually works and where it lands, rebuild-after-bump via the persistent ccache volume, a troubleshooting section grown from dogfooding (interrupted toolchain download, bind-mount hazards, stale base image, VS Code sluggishness), working on the `solx-llvm` fork from inside the container, and an orientation section for Slang contributors (no Hermit; `infra` CLI role split between `bootstrap.sh` and `solx-dev`).

## Why

A natural question from contributors coming from Slang is whether `solx-ci-runner` can be used as a devcontainer. As-is it can't, for non-obvious reasons: the package is private (pull needs a `read:packages` token), rustup is homed in `/root` with no non-root user, and — the real gap — the custom LLVM isn't in the image and is only reachable in CI via the Actions cache, so a fresh container greets you with `llvm-sys` build-script failures and an undocumented hour-long build. This PR closes the gap with a supported, documented setup; the stage split removes the auth problem entirely by building the developer-relevant part of the image locally from public sources.

The whole flow was dogfooded end-to-end (bootstrap → `cargo build --release` → `cargo test-slang`); the Boost fix, the watcher excludes, the submodule git-exclude, and most of the troubleshooting section are findings from that run.

## Notes

- **Post-review hardening**: review feedback plus a self-review pass moved Rust out of `base` (it was ~1.5 GB of `/root`-homed toolchain the dev user couldn't reach) into `ci`, caught the pin up to 1.96.0, and corrected the LLVM-pairing comment along the way — rustc 1.96.0 embeds LLVM 22, one major ahead of the image's LLVM-21 tools; they still read its profraw output, and the comment now says to re-check the pairing on every toolchain bump.
- **Local base vs. frozen CI image**: a locally built base re-resolves apt packages, so it is not bit-for-bit identical to the image CI pulls. Fine for development; CI itself still uses the frozen published image. Keep the stage split honest: dev-relevant tools go in `base`, CI-only tooling in `ci` (noted in the Dockerfile header).
- **Image visibility**: with the local-build flow the devcontainer no longer cares that `solx-ci-runner` is private. Making the package public remains worth considering separately — it would let non-org contributors reproduce the exact CI environment — but nothing in this PR depends on it.
- **Not done here (possible follow-ups)**:
  - Publishing the `target-llvm/target-final` artifact from main (reusing the existing `build-llvm` cache key) so the bootstrap can download instead of building — would cut first-open from ~1h to minutes.
  - A README pointer in `solx-llvm` declaring the solx devcontainer the canonical dev environment for the fork (its regression CI already builds through a solx checkout).
  - ~~`/boost*` in the `solx-solidity` fork's own `.gitignore`~~ — done in NomicFoundation/solx-solidity#158 (cherry-pick of the fork's #111 onto `0.8.34`), merged; this PR bumps the submodule one commit (`5323a55dd` → `79519a9bb`, exactly that fix) and drops the bootstrap's local-exclude workaround.
  - Moving the shared-rustup layout into the `base` stage itself, thinning the devcontainer layer.


docs/src/developer-guide/00-development-container.md
```diff
@@ -0,0 +1,101 @@
...
+
+1. Clone the repository (submodules can be left uninitialized; the bootstrap handles them) and open it in VS Code.
+2. Run **Dev Containers: Reopen in Container**. The first open builds the `dev` image (~5–10 min, layer-cached afterwards), then prints the next step.
+3. Build the toolchain:
+
+   ```shell
+   .devcontainer/bootstrap.sh
+   ```
+
+   This is the ~1 hour step (cold). It is kept out of the automatic container setup precisely because of that cost — you should know when you are paying it.
```
OmarTawfik: nit suggestion:
If the repository is not functional without it, i suggest adding the bootstrap scripts to `postCreateCommand` in `devcontainer.json`. We know when we are paying for it (because it spawns in a dedicated VS Code terminal).

nebasuke: Oh interesting, I was not aware of this option, I'll have a look.

nebasuke: Had a look, and I don't think it works for our use case:

* `devcontainers/ci` will run container lifecycle commands and therefore run the `postCreateCommand`, thereby building LLVM as part of the CI (which is what I was trying to avoid). 
* I am hesitant that this will be good UX. At the moment building the Docker container is relatively light., while building LLVM can take anywhere from 30 minutes up to an hour on your local machine, taking a good amount of RAM and CPU. I don't think this should be running as part of opening VSCode. 
* And the main reason why I think we shouldn't is that while the bootstrap has a very sensible default, it is not the only choice of building LLVM. You might want to run without enabling MLIR, or use the old pipeline, and having a forced build of possibly the wrong LLVM as part of the `postCreateCommand` feels off.

nebasuke:
Build the Docker container: https://github.com/NomicFoundation/solx/actions/runs/29089933569/job/86352587385

.devcontainer/bootstrap.sh
```diff
@@ -0,0 +1,63 @@
...
+        *) echo "unknown option: ${arg} (supported: --no-mlir, --clean)" >&2; exit 1 ;;
+    esac
+done
+
+# ccache's stock 5G cap evicts mid-build on LLVM; the setting persists in the
+# volume-backed CCACHE_DIR.
+ccache --set-config=max_size=20G
+
+echo "==> Initializing submodules (shallow, as in CI)"
+git submodule update --init --recursive --depth 1
```
hedgar2017: Rerunning bootstrap.sh silently resets a manually checked-out submodule branch: `git submodule update --init --recursive` checks out the recorded gitlink SHA on every run, not just the first. A developer who followed the guide's fork-hacking step 1 (`git -C solx-llvm checkout <branch>`) and later reruns bootstrap — as the header's "Idempotent — rerun it after a solx-llvm submodule bump" invites — gets a clean solx-llvm tree moved back to the recorded detached SHA, then rebuilds upstream LLVM believing they built their branch.

Consider initializing only uninitialized submodules (lines prefixed `-` in `git submodule status`), or documenting the interaction here and in the guide's fork-hacking section.

.devcontainer/bootstrap.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Boost artifacts land inside the solx-solidity submodule and are not in the
+# fork's .gitignore; without this local exclude, git status enumerates the
+# extracted Boost tree (tens of thousands of files) and VS Code's git
+# extension drowns in it ("too many changes").
+EXCLUDE_FILE=$(git -C solx-solidity rev-parse --git-path info/exclude)
```
hedgar2017: This works around a gap that is fixable at the source: the org owns the solx-solidity fork, and its `.gitignore` has no boost entry today. Adding `/boost*` to the fork's `.gitignore` (at the next submodule bump) fixes the noisy `git status` for every consumer — host checkouts, CI, other IDEs — after which this git-plumbing workaround and its justification comment can be deleted.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,78 @@
...
+    "hostRequirements": {
+        "cpus": 8,
+        "memory": "16gb",
+        "storage": "64gb"
+    },
+
+    // Build state lives in named volumes: it survives container rebuilds and
+    // stays off the bind mount, which is slow on macOS/Windows hosts.
+    "mounts": [
+        "source=solx-target,target=${containerWorkspaceFolder}/target,type=volume",
```
hedgar2017: These named volumes are docker-global, so every checkout of this repo on one machine shares the same build state. Two checkouts (clone + fork review, or two worktrees) opened in devcontainers mount the same cargo target/registry and rustup volumes: concurrent use contends on cargo locks and mixes artifacts from different sources at identical container paths; alternating use thrashes fingerprints and can surface stale artifacts.

The spec's `${devcontainerId}` variable exists for exactly this — e.g. `source=solx-${devcontainerId}-target`.

.github/docker/Dockerfile
```diff
@@ -98,14 +108,6 @@ RUN set -eux; \
...
-ENV VALGRIND_VERSION=3.23.0
-RUN curl -fsSL -o valgrind-${VALGRIND_VERSION}.tar.bz2 \
-        https://sourceware.org/pub/valgrind/valgrind-${VALGRIND_VERSION}.tar.bz2 \
-    && tar -xf valgrind-${VALGRIND_VERSION}.tar.bz2 \
-    && cd valgrind-${VALGRIND_VERSION} && ./configure && make && make install \
-    && cd .. && rm -rf valgrind-${VALGRIND_VERSION}.tar.bz2 valgrind-${VALGRIND_VERSION}
-
 # ── Rust ─────────────────────────────────────────────────────────────────────
 # Version must match rust-toolchain.toml in the solx repo.
 ARG RUST_VERSION=1.93.0
```
hedgar2017: This violates the invariant stated right above it: `rust-toolchain.toml` pins 1.96.0, not 1.93.0. The baked default toolchain is dead weight — every CI job in the ci image re-downloads and installs 1.96.0 at its first cargo invocation — and the "1.93.0 → LLVM 21" rationale in the LLVM section rots silently with it.

Pre-existing, but this PR adds a lazily-resolving rustup in the dev stage precisely because baked pins go stale, and its new CI filter watches `rust-toolchain.toml` while nothing guards this pin. Bump to 1.96.0, or make the base install lazy the same way the dev stage does.

.github/docker/Dockerfile
```diff
@@ -117,6 +119,67 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
...
+# created in the bind-mounted workspace belong to the host user (the
+# devcontainer runtime remaps the UID/GID to the host's on Linux).
+# ubuntu:26.04 ships a default `ubuntu` user squatting UID 1000; replace it.
+#
+# valgrind comes from apt here (seconds), unlike the ci stage's pinned
+# from-source build — dev needs a working valgrind, not CI's exact version.
+# Keeping it in the image (rather than ad-hoc installs) also keeps LLVM's
+# cmake valgrind detection stable across container rebuilds; the persistent
+# target-llvm build tree bakes it in.
+ARG USERNAME=dev
```
hedgar2017: `ARG USERNAME` is dead parameterization: devcontainer.json hardcodes `"remoteUser": "dev"` and passes no `build.args`, so overriding this ARG only desyncs the two files — the image gets some other user while the devcontainer still remotes in as `dev`, and creation fails or falls back to root. Inline `dev`.

.github/docker/Dockerfile
```diff
@@ -117,6 +119,67 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
...
+# ── Shared rustup/cargo ──────────────────────────────────────────────────────
+# base homes rustup in /root, unreachable for the non-root user. Install a
+# shared one in the official rust-image layout instead; the actual toolchain
+# is not baked in — rustup downloads the version pinned in rust-toolchain.toml
+# on the first cargo invocation, so this layer never goes stale on a
+# toolchain bump.
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:${PATH}
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
```
hedgar2017, line 159: The dev image still carries base's full root-homed Rust install (~1.5 GB: rustup + 1.93.0 default-profile toolchain + rust-src in `/root`) that this stage's own comment declares unreachable — which is why a second rustup gets installed here. The only consumers of the `/root` install are in the ci stage (the cargo-tools build, and CI jobs running as root).

Moving base's Rust section (and the `/root/.cargo/bin` PATH component) into `ci` removes the dead weight from the image every developer builds locally; nothing else in base needs cargo. rust-src looks doubly vestigial: jobs run 1.96.0 per `rust-toolchain.toml`, whose default profile does not include it.

.github/docker/Dockerfile
```diff
@@ -117,6 +119,67 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
...
+# ═════ CI-only stage — nothing below is needed for local development ═════
+FROM base AS ci
+ARG TARGETPLATFORM
+
+# ── Valgrind (for LLVM regression tests under valgrind) ───────────────────
+ENV VALGRIND_VERSION=3.23.0
+RUN curl -fsSL -o valgrind-${VALGRIND_VERSION}.tar.bz2 \
+        https://sourceware.org/pub/valgrind/valgrind-${VALGRIND_VERSION}.tar.bz2 \
+    && tar -xf valgrind-${VALGRIND_VERSION}.tar.bz2 \
+    && cd valgrind-${VALGRIND_VERSION} && ./configure && make && make install \
```
hedgar2017: Still compiles single-threaded — `make` with no `-j`. Every CI-image build (monthly cron, every `.github/docker` change, amd64 and QEMU-emulated arm64) pays minutes of avoidable build time, worst under QEMU. `make -j"$(nproc)"`.

.github/workflows/test.yaml
```diff
@@ -41,14 +42,54 @@ jobs:
...
+      # rust-toolchain.toml because the container resolves it lazily on first
+      # use.
+      - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
+        id: filter-devcontainer
+        with:
+          filters: |
+            devcontainer:
+              - '.devcontainer/**'
+              - '.github/docker/**'
+              - 'rust-toolchain.toml'
```
hedgar2017, line 58: The filter does not cover the validation harness itself: a PR that only edits the validate-devcontainer job or bumps the devcontainers/ci pin in this file yields `devcontainer == 'false'`, the job is skipped, and a broken harness merges — surfacing on the next unrelated devcontainer PR. Adding `.github/workflows/test.yaml` here costs a ~2-minute job on test.yaml edits.

docs/src/developer-guide/00-development-container.md
```diff
@@ -0,0 +1,102 @@
...
+| `solx-ccache` | `/var/cache/solx-ccache` | LLVM compiler cache |
+
+To start truly fresh, remove the volumes (`docker volume rm solx-target-llvm …`) in addition to rebuilding the container.
+
+## Working on the LLVM fork itself
+
+The devcontainer is also the intended environment for hacking on `solx-llvm`: the fork is not built standalone — `solx-dev` owns the CMake configuration (in `solx-dev/src/llvm/`), and `solx-llvm`'s own regression CI drives its builds through a **solx** checkout in the same runner image.
+
+1. Point the submodule at your branch: `git -C solx-llvm checkout <branch>` (after `git -C solx-llvm fetch --unshallow origin <branch>` if needed).
+2. Rebuild: `./target/release/solx-dev llvm build --enable-assertions --enable-mlir --ccache-variant ccache`.
```
hedgar2017: This workflow cannot run the LLVM regression tests it exists to support: neither bootstrap.sh nor this rebuild command passes `--enable-tests`, so the build has `LLVM_INCLUDE_TESTS=Off` and `LLVM_BUILD_UTILS=Off` — no FileCheck, no llvm-lit, no `check-*` targets (solx-dev maps `enable_tests` → `LLVM_INCLUDE_TESTS` and promotes it to `enable_utils`; both default off).

A developer following this section rebuilds their branch, has no way to run lit locally, and discovers breakage only in solx-llvm's CI — despite the paragraph above calling the container "the intended environment for hacking on solx-llvm". Document `--enable-tests` here (and/or add a bootstrap passthrough flag).

docs/src/user-guide/01-installation.md
```diff
@@ -69,6 +69,9 @@ This repository maintains intuitive and stable naming for the executables and pr
 > Building from source is only necessary for development, research, and debugging purposes.
 > Deployment and production use cases should rely only on [the officially released executables](#static-executables).
 
+> For development, the repository also ships a [devcontainer](../developer-guide/00-development-container.md)
+> that provides all dependencies below on the exact toolchain image CI uses.
```
hedgar2017: "the exact toolchain image CI uses" overclaims: the devcontainer builds the image locally from the same definition, and the developer guide itself says it is "not bit-for-bit identical to the frozen image CI pulls". A user chasing an apt-drift discrepancy gets sent the wrong way. Suggest: "the same toolchain image definition CI uses".

hedgar2017 (changes requested):

nebasuke:
Build of fixed up container: https://github.com/NomicFoundation/solx/actions/runs/29209314411

.devcontainer/bootstrap.sh
nebasuke: Fixed in 0d39af02 + 6635c524, though not via init-only-uninitialized: that would
break the other documented flow (rerunning after a submodule bump, where the
tree is initialized but stale and *should* move). The loop now distinguishes:

- on a branch → left untouched, with a notice; nested submodules still sync to
  whatever you have checked out (solx-solidity carries three under `deps/`)
- detached at a different SHA → still reset, but announced. A deliberately
  pinned detached checkout is indistinguishable from a stale post-bump tree, so
  a branch is the protection signal — the fork-hacking section says so now.

.devcontainer/bootstrap.sh
nebasuke: Agreed, turns out you already fixed it on the fork's `main` in #111; it just
never reached `0.8.34`. Cherry-picked onto that branch as
NomicFoundation/solx-solidity#158, now merged.

.devcontainer/devcontainer.json
nebasuke: Took `${devcontainerId}` for the three target volumes (230b86f6). I kept
rustup/cargo/ccache shared deliberately: cargo's registry is lock-protected,
and cross-checkout sharing of the LLVM ccache is the point of persisting it —
that's what makes a second checkout's build "minutes, not an hour".

Two things that came out of double-checking this:

- sharing `target-llvm` across checkouts wouldn't even have worked: CMake bakes
  `${containerWorkspaceFolder}` into the cache, and that path differs per
  checkout basename, so per-checkout is correct there regardless.
- rustup is *not* concurrency-safe (rustup#988 — no cross-process locking), so
  the shared-volume caveat is now documented in the guide's Persistence section
  (e30ca908) rather than claimed away.

.github/docker/Dockerfile
nebasuke: Bumped to 1.96.0 (dd903add). The lazy-resolve alternative can't work in this
stage: the ci stage's `cargo install` runs at image build time with no checkout,
so there's no rust-toolchain.toml for rustup to resolve — the pin has to stay
concrete. To stop it drifting again, a renovate regex customManager now emits
the same `rust`/`rust-version` dependency as the built-in rust-toolchain
manager, so the ARG rides the same update PR (e5416bc3; verified with a local
extraction dry-run).

Checking your premise also surfaced that the adjacent LLVM comment was already
wrong: rustc 1.96.0 embeds LLVM **22**, not 21. The image's LLVM-21
llvm-cov/llvm-profdata still read its profraw output, so coverage works — the
comment now states that actual pairing and tells toolchain bumps to re-check it
(ab7f23e9).

.github/docker/Dockerfile
nebasuke: Inlined (dd903add).

.github/docker/Dockerfile
nebasuke: Moved base's Rust section into ci (dd903add) — agreed the split had it in the
wrong stage. One correction on the aside: rust-src isn't vestigial. The
build-rust and rust-unit-tests actions request it explicitly for `-Zbuild-std`;
they were installing it onto the resolved toolchain themselves, so only the
baked 1.93.0 copy was dead. With the pin now matching rust-toolchain.toml,
baking rust-src means jobs find it preinstalled instead of downloading it.

.github/docker/Dockerfile
nebasuke: Done (dd903add). Also moved the valgrind block above the Rust layer in the ci
stage (ab7f23e9): its pin effectively never changes, while RUST_VERSION bumps
every few weeks and was invalidating the valgrind build for nothing.

.github/workflows/test.yaml
nebasuke: Added (6b2a9dcd). One expectation-setting note: it's pricier than ~2 minutes —
the job builds the base+dev stages from scratch on the runner (no layer cache),
so more like 5–10. It runs in parallel and is never the long pole on a
test.yaml edit, so accepted as is.

docs/src/developer-guide/00-development-container.md
nebasuke: Documented `--enable-tests` in the rebuild command (b9a648e5), plus the
`-DLLVM_PARALLEL_LINK_JOBS='2'` cap that bootstrap already uses — the full
toolset's uncapped links would OOM the guide's own 16 GB host minimum
(e30ca908). I passed on a bootstrap passthrough flag: by that point in the flow
you're invoking solx-dev directly anyway.

docs/src/user-guide/01-installation.md
nebasuke: Taken as suggested (b9a648e5).

.devcontainer/bootstrap.sh
nebasuke: Follow-up: I've walked this back and taken your init-only suggestion after all
(147d2d83). The branch/detached heuristic bought one convenience — auto-refresh
on rerun after a bump — at the cost of a policy rule ("a branch protects your
checkout, detached doesn't") that users only learn from a log line, plus the
subtle bash to implement it. Making the refresh explicit dissolves the
ambiguity the heuristic existed to resolve: the bootstrap now only initializes
never-initialized submodules and never moves an existing checkout. If one
differs from the recorded commit it warns with the exact
`git submodule update --init --recursive --depth 1` to run, and the guide's
submodule-bump section leads with that command.

hedgar2017 (approved):

PR #526, opened 2026-07-09
perf(core): skip legacy-assembly materialization in the via-IR pipeline when unused

### LINKED ISSUE #522: perf(core): skip legacy-assembly materialization in the via-IR pipeline when unused
Claude findings as part of reviewing #504. Feel  free to AI analyse and dismiss. 

## Claude summary

In the via-IR pipeline, solc's `evm.legacyAssembly` tree is deserialized for every
contract even though via-IR compiles from Yul and never consumes it. After #504 that
deserialization is parallelized (a big win — see below), but it is still performed and
then discarded. Skipping it entirely when the legacy assembly is neither used for
compilation nor requested as output should remove the remaining cost from the via-IR
path.

## Background / evidence

#504 changed `OutputContractEVMLegacyAssembly` to capture solc's legacy-assembly JSON
verbatim (`Raw`) during the single-threaded solc-output parse and `materialize()` it in
parallel before use, instead of deserializing it eagerly on the serial head.

A temporary compile-benchmark run over the OpenZeppelin via-IR corpus fixture
(`openzeppelin-contracts-0.34--solx-via-ir-dwarf`) showed how much this parse dominates
via-IR wall-clock:

| binary | mean ± σ | vs main |
|---|---|---|
| pr (#504) | 42.934 s ± 0.475 s | **0.394×** |
| main | 108.942 s ± 1.185 s | 1.000× |
| release | 107.902 s ± 0.347 s | 0.990× |

via-IR `main` is ~2× the legacy `main` for the same project (108.9 s vs 55.5 s) purely
because of this serial legacy-assembly deserialization — dead weight in via-IR, where the
compile IR is Yul. #504 parallelizes it (108.9 → 42.9 s). This issue is the natural next
step: in via-IR, don't parse it at all unless it was explicitly selected as output.

(By contrast, in the **legacy** pipeline the parsed assembly *is* the compile IR, so it
must always be materialized there — this optimization is via-IR-only.)

## Root cause

`Project::try_from_solidity_output` (`solx-core/src/project/mod.rs`) materializes **all**
legacy assemblies unconditionally, before the pipeline split:

```rust
solc_output
    .contracts
    .values_mut()
    .flat_map(|file| file.values_mut())
    .filter_map(|contract| contract.evm.as_mut()?.legacy_assembly.as_mut())
    .collect::<Vec<_>>()
    .into_par_iter()
    .try_for_each(|legacy_assembly| legacy_assembly.materialize())?;
```

But the legacy assembly is only consumed:
- as the **compile IR** in the legacy pipeline (`if !via_ir { … preprocess_dependencies(…) }`,
  and the per-contract `ContractEVMLegacyAssembly::from_contract` build), and
- as **output**, when the contract's selection requests it
  (`legacy_assembly.filter(|_| output_legacy_assembly)`, where
  `output_legacy_assembly = output_selection.check_selection(path, name, EVMLegacyAssembly)`).

In a via-IR compile that doesn't select `evm.legacyAssembly`, neither applies, yet every
contract's tree is still parsed and then dropped.

## Proposed change

Materialize a contract's legacy assembly only when it will actually be used:

- `!via_ir` → always (it's the compile IR; `preprocess_dependencies` needs the parsed form), or
- the contract's output selection requests `EVMLegacyAssembly`.

i.e. fold the `via_ir` flag and a per-contract `check_selection(path, name, EVMLegacyAssembly)`
into the `filter_map` that feeds the parallel `materialize()` pass, so via-IR compiles that
don't ask for legacy assembly skip the parse outright. `Raw` variants that are never
materialized are simply dropped later (they're not serialized on the compile path, and the
output `filter` gates them), so `parsed_mut()`/`into_parsed()` invariants are unaffected.

## Considerations

- **Legacy pipeline must be unaffected** — `preprocess_dependencies` and
  `from_contract` still require the parsed form for every contract, so the gate must keep
  materializing everything when `!via_ir`.
- **`--mlir` feature** already ignores `via_ir` and uses its own IR; the same "only when
  selected as output" logic applies there.
- **Output correctness** — a via-IR compile that *does* select `evm.legacyAssembly` must
  still emit it; the per-contract `check_selection` guard preserves that. Worth a
  standard-JSON test asserting via-IR + `evm.legacyAssembly` selection round-trips.

## Impact

Removes the remaining (post-#504, parallelized) legacy-assembly parse from via-IR compiles
that don't request it — the dominant non-LLVM cost in the via-IR head for large projects.
Quantifiable by re-vendoring the `openzeppelin-contracts-0.34--solx-via-ir-dwarf` fixture
in the compile-benchmark and comparing against the #504 baseline (42.9 s).

Discovered while reviewing #504.

nebasuke:
# Claude summary

## What

Skip parsing solc's `evm.legacyAssembly` tree in the via-IR pipeline when it is
never read.

solc emits legacy assembly and solx captures it (`Raw`) for every contract, but
only the **EVMLA** pipeline consumes it — it lowers the tree into its compile IR.
The **via-IR** pipeline compiles from Yul and never touches it, so for a via-IR
compile that does not select `evm.legacyAssembly` as output, parsing the tree is
pure dead weight on the serial solc-output head.

This gates both the parallel `materialize()` pass and the matching `into_parsed()`
unwrap in `Project::try_from_solidity_output` on `!via_ir || <legacy assembly
selected as output>`. via-IR now leaves the tree `Raw` and skips the parse. The
EVMLA pipeline (and `preprocess_dependencies`) still materialize every contract,
and the output selection still governs what is emitted, so **bytecode and JSON
output are unchanged**.

Closes #522.

## Why

Follow-up to #504. That PR moved the legacy-assembly parse off the serial head
and parallelized it; the compile-benchmark showed how much it costs in via-IR:

| `openzeppelin-contracts-0.34` via-IR | mean ± σ | vs main |
|---|---|---|
| #504 | 42.934 s ± 0.475 s | 0.394× |
| main (pre-#504) | 108.942 s ± 1.185 s | 1.000× |

The remaining ~43 s still includes the (now parallel, but still performed) parse
of an assembly via-IR never uses. Skipping it entirely should push this lower.

## Impact

- Correctness: output-preserving (size + solx-tester gas unchanged); the gate only
  affects whether an *unused* intermediate is parsed.
- Perf: removes the legacy-assembly parse from via-IR compiles that don't request
  it — the dominant non-LLVM cost in the via-IR head for large projects.

## Notes

- **Temp commit to revert before merge:** `test(benchmark): TEMP vendor OZ via-ir
  fixture …` vendors the authentic OZ via-IR corpus fixture so `ci:compile-benchmark`
  measures the delta against the 42.9 s baseline. Drop it once the number is recorded.
- Add the `ci:compile-benchmark` label to get that measurement.
- Not built locally (no LLVM prefix in this environment) — relying on CI for the
  compile/clippy/fmt gates.


nebasuke:
Results don't match the initial issue https://github.com/NomicFoundation/solx/issues/522. 

After deeper analysis it seems the legacy assembly does NOT get propagated in the via-ir path unless explicitly added as additionally requested outputSelection.

PR #527, opened 2026-07-09
ci(benchmark): fetch corpus from published repo (+ via-IR prep)

nebasuke:
This PR will stay in draft for a while, until https://github.com/NomicFoundation/hardhat/pull/8415 is slowly merged into Hardhat (in pieces). This is to prove the concept works, and the approach can be continued in HH3.

# Claude summary

Follow-up to #518 (now merged). Two changes to the compile-time benchmark:

1. **Stop vendoring fixtures — fetch the published corpus instead.** This is the long-term source the #518 README already anticipated.
2. **Prep the via-IR pipeline as an opt-in benchmark dimension** (off by default; the EVMLA pipeline always runs).

## Corpus: fetched, not vendored

The fixtures leave the repo (`tests/benchmark/fixtures/` and its `manifest.json` are deleted). A new *Checkout benchmark corpus* step pulls them from the hardhat-published corpus repo `nomic-foundation-automation/hardhat-benchmark-results`, via a **non-cone sparse checkout** that materializes only the benchmarked files (not the whole corpus). The pinned commit is now the fixtures' provenance — the manifest lives in the corpus repo.

> ⚠️ **Temporary preview location.** This pins to the force-pushed `solx-corpus-preview` branch (hence a commit SHA, not the branch tip). The final home is `solx-corpus/` on that repo's `main`, published from hardhat merge runs — repoint the single `ref:` line once it's live. The pinned commit will need periodic bumping (or may be GC'd) during this preview phase.

## via-IR: prepped, off by default

- `run.sh` gains a repeatable `--variant` (default `solx-legacy-dwarf`) that selects `<scenario>/<variant>.json` per scenario. The compiler pipeline is encoded in the standard JSON, so switching variants needs no solx flag.
- The checkout fetches **both** the legacy and via-IR variants of the three scenarios, so via-IR is present and ready; it's just not timed by default. A new `include_via_ir` `workflow_dispatch` input turns it on.
  - Caveat: `workflow_dispatch` runs upload the artifact but don't post the PR comment (the comment step is PR-gated). A dedicated label is the path if via-IR-with-comment on a PR is wanted later; kept to the dispatch input here to avoid extra label-gating machinery.
- `report.py` drops the hardcoded "EVMLA pipeline" title, since via-IR sections can now appear (each section header already carries the variant).

## Fixtures

Same three scenarios timed as before (EVMLA/legacy-DWARF): `ens-verifiable-factory-solx` (52), `openzeppelin-contracts-0.34` (422), `uniswap-v4-core-solx` (157). **`aave-v4` is never fetched** — its dumps require `EVM_DISABLE_MEMORY_SAFE_ASM_CHECK=1`, which isn't part of standard JSON, so `run.sh`'s validation pass would abort on it.

Also hardened: `--fixtures` is now required (the vendored default is gone), and an empty variant resolution fails loudly instead of feeding hyperfine a literal glob.

## Validation

- End-to-end dry run with stubbed `solx`/`hyperfine` against a fake corpus: both guards fire (missing `--fixtures`; unknown variant); default → 2 legacy fixtures; `--variant solx-legacy-dwarf --variant solx-via-ir-dwarf` → 4 fixtures with correct `<scenario>--<variant>` naming; `report.py` renders all sections with the generalized title.
- `bash -n run.sh` clean.
- `actionlint` clean apart from the pre-existing `solx-linux-amd64` custom-runner-label false positive (from the temporary-runner commit on the base branch).


hedgar2017:

Force-pushed `compile-benchmark-corpus`: `ef7b8d3f` → `922e6b8d`. No conflicts.

`git range-diff` flags the commit as changed, but it is context-only: the `--bin pr=./target/x86_64-unknown-linux-gnu/release/solx` and `--bin main=...` lines are context this commit never touches, and the fork point (`dd05f644`) predates their current form. Both trunks already carry those lines identically. Nothing in the change itself moved.

PR #529, opened 2026-07-10
fix(test-harness): fix the integration error gate and self-verify compiler

nebasuke:
Addressed a couple of issues I discovered while going through the Excel reports. Also added compiler verification based on the issue fixed in #497 . I also bumped the integration test solx to 0.1.4 (from 0.1.3) and made it auto-update on releases via renovate.

# Claude summary

Trust fixes for the foundry/hardhat integration harness and the report converter. None of this changes what the suite measures — it makes the suite fail loudly where it previously lied silently. Updated after the review round (commits c2b0498b and a30295d9); each review comment has an inline reply.

## Harness fixes

1. **The test-regression gate was dead code.** In both runners, the test-failure check required `reference_build_errors < candidate_build_errors`, but the build-errors branch above it `continue`s in exactly that case — the condition could never be true, so a PR that broke real project tests could not fail the workflow. The gate now iterates the attempted-projects list directly (not the build table's keys, which only failure paths used to populate), so it compares test failures for every project that ran — including the common case where both toolchains build clean — and cannot be starved by a producer path that forgets to record.

2. **Hard build failures were invisible.** Both runners `continue`d without recording anything, so a toolchain whose builds all failed simply vanished from the report. This is how `solx-latest` ran 0-for-50 foundry builds in full-matrix mode with no visible trace (see 4). Failures are now recorded in the correctness table for every toolchain and pushed as `BuildFailuresReport` into the benchmark data. A zero-error build that emits no bytecode is ranked worse than any error count in the gate (shown as "no bytecode") so it can neither read as clean nor be absorbed by a reference toolchain's real errors; the spreadsheet reports it as 1 failure.

3. **The harness never verified it tested what it claimed.** #497: hardhat compiled with the bundled solc for months while reporting under solx toolchain names, and no check existed to catch that class of bug. Two self-checks added:
   - Compilers are invoked through a shell shim (`CompilerInvocationShim`, compiled as a tiny native binary because Hardhat 3 content-sniffs the compiler file) that records each invocation's arguments in a marker file, one argument per line. After every successful build the runner asserts the configured compiler was invoked with `--standard-json` — a bare `--version` probe, or an argument merely containing the flag text, does not count as a compile. Hardhat verifies only solx toolchains — its solc columns legitimately use Hardhat's own downloaded compiler. Non-Unix hosts pass through with verification disabled.
   - A shared coverage check (`test::verify_benchmark_coverage`) errors when any attempted project × enabled toolchain pair produced no benchmark data. With the current control flow every iteration records or aborts, so this is explicitly a tripwire for future edits that skip a pair silently — the omission class behind #497.

4. **The `solx-latest` pin lagged a release behind.** `dev-compilers.json` pinned 0.1.3 (based on solc 0.8.30) while the pragma rewrite pins the default `=0.8.34`, so forge rejected the binary for every foundry project; hardhat only survived via a 0.8.30 version override. Bumped to 0.1.4 (verified: advertises `0.8.34+commit.e2cbf92c`), stale hardhat override removed, and a Renovate custom manager now tracks the pin — bump PRs get `ci:integration` + `ci:integration-benchmark-full` so the full matrix proves each new release against the harness.

## Report converter fixes (riding along)

- Worksheets with no data rows are dropped from the workbook instead of being emitted with dangling comparison-formula columns (the hardhat Gas/Size ghost sheets, the solx-tester time sheets).
- `Worksheet::column_identifier` dropped a base-26 digit past column Z (26 → "A" instead of "AA"), silently mis-referencing columns in formulas of wide sheets. Fixed with a boundary unit test.

## Also

- `timeout-minutes: 300` on the integration job so a hung compile no longer squats the self-hosted runner for the 6h default.

## Verification

- `cargo build`/`test`/`clippy`/`fmt` clean on `solx-dev` + `solx-benchmark-converter`; unit tests cover the shim lifecycle end-to-end against the compiled shim binary (version probe rejected, embedded-flag-text argument rejected, real `--standard-json` invocation accepted, reset semantics) and the column-identifier boundary.
- Empty-sheet dropping smoke-tested against the real converter binary (hardhat-shaped input yields only Build Failures / Test Failures / Compilation Time sheets).
- Renovate rule validated with `renovate-config-validator` and `--dry-run=full` both at 0.1.4 (no update, steady state) and with the pin set back to 0.1.3 (proposes 0.1.4, rewrites both URL occurrences, single branch).
- actionlint/shellcheck: only pre-existing findings on untouched lines.

## Caveat for reviewers

The gate fix means previously-invisible test regressions will now actually fail integration runs. The first few labeled PRs may surface pre-existing candidate-vs-main diffs that were always there but never reported — that's the fix working, not a new breakage. Suggest validating this PR itself with `ci:integration` and then `ci:integration-benchmark-full`: the full-matrix run should now show solx-latest foundry columns, a hardhat report without ghost sheets, and will prove the `--standard-json` verification assumption against the real forge/hardhat invocations.


solx-dev/src/test/hardhat/mod.rs
```diff
@@ -58,6 +58,21 @@ pub fn test(
...
+        .filter(|(_identifier, compiler)| !compiler.disabled)
+    {
+        let shim_directory =
+            crate::utils::absolute_path(format!("./temp-compiler-shims/hardhat/{identifier}"))?;
+        let compiler_path = crate::utils::absolute_path(compiler.path.as_str())?;
+        compiler_shims.insert(
+            identifier.clone(),
+            crate::utils::CompilerInvocationShim::new(compiler_path, shim_directory.as_path())?,
+        );
+    }
```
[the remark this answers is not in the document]
nebasuke: Deliberate: shim creation never touches the real compiler binary (the path is baked into the generated source as a literal), so there's no existence/executability requirement and the cost is one ~0.5s `rustc` of a six-line program, only when solc is enabled (full-matrix runs). Keeping creation uniform keeps the hardhat and foundry runners symmetric — foundry *does* use the solc shim via `forge --use` — and avoids adding another `contains("solx")` branch point, the same heuristic that caused #497. The unused-verification case is already handled where it matters: `verify()` is gated to solx toolchains with a comment explaining why.

hedgar2017 (changes requested):
Thanks a lot! Let me slightly refactor it and merge first thing next week.

solx-dev/src/test/hardhat/mod.rs
```diff
@@ -413,18 +461,19 @@ pub fn test(
...
+                .copied();
             let candidate_test_failures = test_correctness_table
                 .get(project)
                 .and_then(|toolchains| toolchains.get(&candidate_toolchain))
-                .copied()
-                .unwrap_or_default();
-            if candidate_test_failures > reference_test_failures
-                && reference_build_errors < candidate_build_errors
+                .copied();
+            if let (Some(reference_test_failures), Some(candidate_test_failures)) =
```
hedgar2017, line 469: The revived test-regression gate can't fire for all-green projects on Hardhat. This comparison sits inside `for project in build_correctness_table.keys()` (line 437), but in this runner `build_correctness_table` is populated **only on the build-failure path** (line 275) — the success path writes just `test_correctness_table`. So a project whose reference and candidate both build successfully is never a key, the loop skips it, and this `candidate_test_failures > reference_test_failures` check never runs for it. Foundry avoids this because it also inserts into `build_correctness_table` on the success path (`foundry/mod.rs:294`).

Failure: reference `solc`, candidate `solx`, project `foo` — both compile, `solx` regresses tests 3-vs-0. `foo` isn't in `build_correctness_table`, so `3 > 0` is never evaluated and CI stays green, which is the exact healthy-build / broken-test case this gate is meant to catch. Iterating the union of build+test project keys (or inserting into `build_correctness_table` on the success path like Foundry) would close it.

solx-dev/src/utils.rs
```diff
@@ -325,6 +325,121 @@ pub fn sed_file<P: AsRef<Path>>(file_path: P, patterns: &[&str]) -> anyhow::Resu
...
+                anyhow::anyhow!("Creating compiler shim directory {directory:?}: {error}")
+            })?;
+            let marker_path = directory.join("invoked");
+            let shim_path = directory.join(compiler_path.file_name().ok_or_else(|| {
+                anyhow::anyhow!("Compiler path {compiler_path:?} has no file name")
+            })?);
+            let source_path = directory.join("shim.rs");
+            let source = format!(
+                "fn main() {{
+    let _ = std::fs::write({marker_path:?}, b\"\");
```
hedgar2017: `verify()` proves the shim was invoked *at least once* since `reset()`, not that it was invoked *for the compilation*. The marker is written unconditionally on every exec here, and Forge/Hardhat issue a `solc --version` probe through the configured path before compiling — that probe alone sets the marker. So a build system that version-probes the configured compiler but then compiles with a different/bundled one (exactly the #497 class) still passes `verify()`. The unit test encodes this weak semantics: it runs a bare `--version` and asserts `verify().is_ok()` (lines 508-516). To actually guard #497 the marker would need to record something only a real compile produces (e.g. seeing `--standard-json` / compile args, or an emitted-artifact path), not merely any invocation.

solx-dev/src/test/foundry/mod.rs
```diff
@@ -445,7 +467,28 @@ pub fn test(
         }
     }
 
-    let benchmark = solx_benchmark_converter::Benchmark::from_inputs(benchmark_inputs.into_iter())?;
+    // A toolchain that ran but contributed nothing would be silently absent
+    // from the report, misreported as clean (see #497).
+    if !benchmark_inputs.is_empty() {
```
hedgar2017: This zero-data self-check can never fire. Every iteration of the project x compiler x codegen loop either pushes at least one `benchmark_inputs` entry (a `BuildFailuresReport` before `continue` at 253-257 / 299-303, or the success-path size/gas/test/time reports) or aborts via `?`. So by the time execution reaches here, every enabled toolchain already has an entry and `any(|input| input.toolchain == toolchain_name)` is always true — the `bail!` is unreachable.

It also doesn't catch the #497 case it names: a project that compiles to 0 contracts with 0 errors hits `built_contracts_count == 0` (line 298), pushes `BuildFailuresReport(0)`, and is then reported as a clean 0-failure toolchain despite emitting no bytecode. A meaningful check would assert a *success* record exists per (project, toolchain), not just any entry.

solx-dev/src/test/hardhat/mod.rs
```diff
@@ -331,7 +358,28 @@ pub fn test(
         }
     }
 
-    let benchmark = solx_benchmark_converter::Benchmark::from_inputs(benchmark_inputs.into_iter())?;
+    // A toolchain that ran but contributed nothing would be silently absent
+    // from the report, misreported as clean (see #497).
+    if !benchmark_inputs.is_empty() {
```
hedgar2017: Same defect as the Foundry runner (`foundry/mod.rs:472`): this zero-data self-check can never fire. Every iteration either pushes a `benchmark_inputs` entry (a `BuildFailuresReport` before `continue` at 279-283, or the success-path TestFailures/CompilationTime/TestingTime reports at 342-357) or aborts via `?`. Reaching this loop therefore guarantees every enabled toolchain already has an entry, so `any(...)` is always true and the `bail!` is dead code. A `npm run compile` that exits 0 while producing no artifacts still records reports and is reported as clean — the very #497 case the comment claims to guard.

solx-dev/src/test/foundry/mod.rs
```diff
@@ -244,6 +265,7 @@ pub fn test(
                 }
             };
             let compilation_time = build_timestamp_start.elapsed().as_millis() as u64;
+            compiler_shim.verify(toolchain_name.as_str(), project_name.as_str())?;
```
hedgar2017, line 270: `verify()?` propagates out of the whole `test()` function, and it sits *before* the graceful `build_errors > 0 || built_contracts_count == 0` continue path (line 298). If `forge build` exits 0 without invoking the shim (e.g. a project with no `.sol` files to compile), this aborts the entire harness run — every remaining project — instead of recording that one (project, toolchain) as a build failure and continuing, which is what every other failure path does. Recording + `continue` here (rather than `?`) would keep one odd project from sinking the whole run.

solx-dev/src/utils.rs
```diff
@@ -325,6 +325,121 @@ pub fn sed_file<P: AsRef<Path>>(file_path: P, patterns: &[&str]) -> anyhow::Resu
...
+    );
+    eprintln!(\"compiler invocation shim: {{error}}\");
+    std::process::exit(127);
+}}
+"
+            );
+            std::fs::write(source_path.as_path(), source).map_err(|error| {
+                anyhow::anyhow!("Writing compiler shim source {source_path:?}: {error}")
+            })?;
+            let mut rustc_command = Command::new("rustc");
```
hedgar2017: `CompilerInvocationShim::new()` compiles the shim with `rustc`, and it's called at harness startup (`foundry/mod.rs:72`, `hardhat/mod.rs:72`) before any project runs. That makes `rustc` a hard runtime dependency of `solx-dev test foundry|hardhat`. A runner carrying prebuilt `solx-dev`/`solx` binaries but with no `rustc` on `PATH` (a build-vs-run environment split) now aborts up front with a shim-compile error rather than the usual project-level diagnosis. The doc comment's assumption that rustc is 'always available since solx-dev is built by cargo' doesn't hold when the test runner isn't the build host.

solx-dev/src/utils.rs
```diff
@@ -325,6 +325,121 @@ pub fn sed_file<P: AsRef<Path>>(file_path: P, patterns: &[&str]) -> anyhow::Resu
...
+        {
+            std::fs::create_dir_all(directory).map_err(|error| {
+                anyhow::anyhow!("Creating compiler shim directory {directory:?}: {error}")
+            })?;
+            let marker_path = directory.join("invoked");
+            let shim_path = directory.join(compiler_path.file_name().ok_or_else(|| {
+                anyhow::anyhow!("Compiler path {compiler_path:?} has no file name")
+            })?);
+            let source_path = directory.join("shim.rs");
+            let source = format!(
```
hedgar2017: The generated shim source embeds `{marker_path:?}` and `{compiler_path:?}` (lines 372, 374) into Rust source via `Debug`. For a non-UTF-8 path (valid on Unix), `Debug` output isn't guaranteed to be a valid Rust string literal, so the generated `shim.rs` can fail to compile with a cryptic `rustc` error instead of a clear path diagnostic. If these paths can ever be non-UTF-8, emit them as byte-string literals (or validate/escape explicitly) rather than relying on `{:?}`.

hedgar2017 (commented):
Automated correctness review of the harness + converter changes. Six inline notes below. Three touch this PR's own goals: the Hardhat test-regression gate is still dead for all-green projects (the common healthy-build / broken-test case), the compiler self-check passes on a bare `--version` probe so it doesn't actually catch the #497 misattribution class, and the zero-data self-check is unreachable. The other three are harness robustness: aborting the whole run on a single project, a new `rustc` runtime dependency, and non-UTF-8 path handling in the shim generator. The base-26 column fix, empty-sheet dropping, and the Renovate manager all check out.

solx-dev/src/test/foundry/mod.rs
```diff
@@ -445,7 +467,28 @@ pub fn test(
...
-    let benchmark = solx_benchmark_converter::Benchmark::from_inputs(benchmark_inputs.into_iter())?;
+    // A toolchain that ran but contributed nothing would be silently absent
+    // from the report, misreported as clean (see #497).
+    if !benchmark_inputs.is_empty() {
+        for ((_identifier, compiler), codegen) in config
+            .compilers
+            .iter()
+            .filter(|(_identifier, compiler)| !compiler.disabled)
+            .cartesian_product(["legacy", "viaIR"])
+        {
```
hedgar2017: An identical loop is in `hardhat/mod.rs`. Should be outlined one level up.

solx-dev/src/utils.rs
```diff
@@ -325,6 +325,121 @@ pub fn sed_file<P: AsRef<Path>>(file_path: P, patterns: &[&str]) -> anyhow::Resu
...
+/// Passing the shim path instead of the real compiler lets test runners verify
+/// after each build that the configured compiler was actually invoked.
+///
+/// The shim must be a real native executable, not a script: Hardhat 3 sniffs
+/// the compiler file's content (`isBinaryFile`) and loads anything that looks
+/// like text as a solc-js module. It is therefore generated as a tiny Rust
+/// program and compiled with `rustc` at harness startup (always available,
+/// since `solx-dev` itself is built by cargo).
+///
+pub struct CompilerInvocationShim {
```
hedgar2017: It's a big entity with its own tests so it's better to move it out of `utils.rs` to its own `shim.rs`.

hedgar2017 (changes requested):

solx-dev/src/test/hardhat/mod.rs
nebasuke: Confirmed, good catch — the gate only covered projects that already had a build failure. Fixed in c2b0498b and a30295d9: the gate now iterates the attempted-projects list directly (in both runners) instead of the build table's keys, so it cannot miss a project regardless of what any producer path recorded; missing build entries default to 0 errors.

solx-dev/src/utils.rs
nebasuke: Strengthened in c2b0498b and a30295d9: the shim now records each invocation's arguments in the marker (one argument per line, newlines escaped), and `verify()` requires an exact `--standard-json` argument — the interface both forge and hardhat drive compilers through. A bare `--version` probe no longer passes, an argument merely containing the flag text doesn't either, and the unit test asserts both.

One correction on the framing: in #497 the configured compiler was never consulted at all — no probe either — so the existence check did catch that instance. What it missed was the stronger probed-but-compiled-elsewhere variant, which this closes.

solx-dev/src/test/foundry/mod.rs
nebasuke: Right on both counts. In c2b0498b and a30295d9:

- A zero-error build that emits no bytecode now records `usize::MAX` in the gate table (rendered as "no bytecode" in gate errors), so it can neither read as clean nor be absorbed by a reference toolchain's real error count; the spreadsheet gets a human-scale count of 1.
- The per-toolchain loops in both runners were replaced by one shared per-(project, toolchain) check (`test::verify_benchmark_coverage`). It is still structurally unreachable today — every iteration records or aborts — and its doc comment says so: it's kept as a tripwire against future control-flow edits that skip a pair without recording it, the omission class behind #497. Happy to drop it entirely instead if you'd rather not carry it.

solx-dev/src/test/hardhat/mod.rs
nebasuke: Same fix as the foundry side — both runners now call the shared `test::verify_benchmark_coverage` (see the reply on `foundry/mod.rs:472` for the reasoning), and the gate fix in the same commits closes the "reported as clean" half.

solx-dev/src/test/foundry/mod.rs
nebasuke: Deliberate, and I'd like to keep it: a verify failure isn't a project-level failure — it means the harness's compiler attribution is broken, so every subsequent number in the run is suspect. Failing the whole run loudly seems right for that. The triggering state is also not reachable benignly: every (project, toolchain) iteration deletes and re-clones the project and forge gets `--force`, so a build cannot succeed without invoking the configured compiler — if the shim saw no compile, the build system genuinely compiled with something else, which is exactly the #497 case. If you'd still prefer record-and-continue for symmetry with the other failure paths, I'll change it.

solx-dev/src/utils.rs
nebasuke: True in principle. In practice the only environment that runs `solx-dev test foundry|hardhat` is the integration workflow, which cargo-builds `solx-dev` on the same runner, so `rustc` is guaranteed there. If a prebuilt-binary runner ever appears, the right behavior is probably to disable verification with a loud warning instead of aborting — I can add that fallback now if you want it.

solx-dev/src/utils.rs
nebasuke: Technically possible, but the paths come from UTF-8 TOML config joined onto the workspace directory, and the failure mode is a loud `rustc` error at harness startup rather than anything silent. I'd rather not add escaping machinery for it; if you want it closed, the cheap fix is a `to_str()` validation with a clear error at shim creation.

solx-dev/src/test/foundry/mod.rs
nebasuke: Done — both runners now call `test::verify_benchmark_coverage`, which derives the toolchain set itself from the enabled compiler names; the codegen list and toolchain-name formula live next to it in `test/mod.rs`.

solx-dev/src/utils.rs
nebasuke: Done in c2b0498b — moved to `solx-dev/src/shim.rs` together with its tests.

hedgar2017 (approved):
Thanks!

PR #532, opened 2026-07-11
test(debug-info): pin cross-source DWARF line attribution

nebasuke:
Regression test for the cross-source DWARF line-attribution fix that lands in #503. See failing test run (before #503 merged) here: https://github.com/NomicFoundation/solx/actions/runs/29149198748

# Claude summary

## What this pins

#503 (merged as 31ff72fc) keys `DebugInfo::ast_nodes` by `(source_id, start_offset)`
instead of `start_offset` alone. Before that change, AST nodes from different sources with
colliding start offsets overwrote each other in one flat map, so a contract's DWARF line
rows could carry another source's line numbers. This only corrupts `debugInfo` *content* —
bytecode is untouched — so no existing CI signal can catch a regression: the
integration-report gate covers size and gas, and the tester's `DwarfValidator` only checks
structural validity (`low_pc` within bytecode bounds). EDR's stack-trace decoding consumes
exactly this line table.

## Test design

The fixture's two sources have byte-identical lengths, so every AST node byte offset in
`B.sol` collides with one in `A.sol`, while their line ranges are disjoint (`Alpha` on
lines 4–6, `Betaa` on lines 8–10). The test decodes each contract's
`evm.deployedBytecode.debugInfo` with `gimli` (already a workspace dependency) and asserts
every non-empty DWARF line-program row stays within its own contract's line range. It
asserts the line-number set rather than snapshotting bytes, so it survives codegen/PC
churn but fails on any attribution regression. The offset collision itself is asserted
inside the test (equal byte lengths, aligned `contract` offsets), so a fixture edit that
breaks the alignment fails loudly instead of turning the test into a vacuous pass.

## Evidence

- On pre-#503 main: `Alpha line 8 is outside of its source range 4..=6` — Alpha's
  function rows carry Betaa's declaration lines (failing run linked above).
- With #503 merged (branch rebased onto 31ff72fc): both cases pass.


hedgar2017 (changes requested):
Let's just gate these with the solc cargo feature so they don't crash the Slang workflow (examples in other tests).

nebasuke:
> Let's just gate these with the solc cargo feature so they don't crash the Slang workflow (examples in other tests).

Indeed, I noticed the fix in your Windows PR when reviewing as well. I am going to wait and double-check until the Windows fix is merged to make sure I can test and not break the Slang workflow. Thanks for the suggestion!

nebasuke:
> Let's just gate these with the solc cargo feature so they don't crash the Slang workflow (examples in other tests).

I think you fixed this yourself already, in the mod.rs as you gated the whole file:

```
#[cfg(feature = "solc")]
mod debug_info;
```

Tested it in Slang as well for completeness, and it's all good.

hedgar2017 (approved):

PR #533, opened 2026-07-11
ci: stop macOS disk cleanup once a free-space target is met

nebasuke:
Make the macOS disk setup clean only up to a certain amount (100GB) in case we get a less disk allocated runner. The clean disk action currently can take up to 11 minutes (~20-30% of a run), which can impact total CI times as macOS x86 is often the slowest run. 

I recorded disk usage by temporarily disabling the whole artefact cache for Mac runners, see numbers below.

# Claude summary
The `Free disk space (macOS)` step frees ~160 GiB on every run, but the job never uses more than a third of that. Since the step was written (#329), the runner images changed under it — the cleanup is now mostly buying space the job doesn't need, at 5–6.5 minutes per run. This PR makes the step free space only until a **100 GiB available** target is met, cheapest phase first, and skip entirely when stock disk already suffices.

## Cost today (run [29150260864](https://github.com/NomicFoundation/solx/actions/runs/29150260864), cached build, 2026-07-11)

| Leg | Step time | Share of job | Breakdown |
|---|---|---|---|
| MacOS arm64 | 5m16s | 29% of 18m09s | simctl 50s · 18 Xcode bundles 4m11s · toolcache ~15s |
| MacOS x86 | 6m38s | 21% of 30m42s | simctl 10s · 18 Xcode bundles 6m12s · toolcache ~17s |

The x86 leg is the workflow's critical path, so its 6m38s is wall-clock on every PR iteration.

## What the job actually needs

When #329 landed, macos-15 ARM64 had ~11 GB free. Today's images have 320–325 GiB disks: **ARM64 starts with ~42 GiB available, x86 with ~108 GiB**. Measured peak demand above the post-cleanup baseline:

- **Cached runs** (run 29150260864): ~23 GiB (ARM64), ~34 GiB (x86).
- **Cold LLVM build, measured directly** (run [29149150239](https://github.com/NomicFoundation/solx/actions/runs/29149150239), x86 artifact-cache miss): **~52 GiB** with the full `build-final` tree, ccache dir, rust-cache, and solc build all on disk simultaneously.
- **Cold ARM64, bracketed from above**: cache-warmup's macOS legs build LLVM cold with the same flags (`RelWithDebInfo` + assertions + MLIR + utils) and **no cleanup step at all**, and succeeded inside stock ~42 GiB on Jun 21, Jul 9, and Jul 10 (ccache dir measured 2.8 GB in the Jul 10 log). test.yaml adds rust-cache (~3–5 GiB) and the solc build (~4–5 GiB) on top → ARM64 cold peak ≈ **50 GiB**, consistent with the x86 measurement.

So x86 never needs the cleanup — even a cold build fits stock disk with ~55 GiB headroom — and ARM64 needs ~10–15 GiB freed for a cold build, not ~160 GiB.

## The change

The step computes available space up front and stops as soon as **100 GiB** is available:

1. **Gate**: already ≥ 100 GiB → log and exit (x86 today: 6m38s → seconds).
2. Simulator-runtime unmounts (fast, ~10 GiB), re-check.
3. Inactive Xcode bundles in **parallel batches of 4**, re-checking free space between batches (ARM64, measured: a single batch sufficed — 1m14s instead of 5m16s).
4. SDK/toolcache phase only if still short; loud warning if the target is unreachable.

Target derivation: 52 GiB measured cold peak + ~48 GiB margin, sized to the stock-free-space variance already observed across image updates (~11 GiB → 42 GiB → 108 GiB). If images shrink again, the step gracefully degrades back to full cleanup rather than breaking.

**Expected effect on cached runs: x86 job 30m42s → ~24m (workflow wall-clock −21%); ARM64 18m09s → ~15m.**

Everything stays synchronous — per-batch `xargs` exit-status handling, truthful "Removed N of M" counts, and the active-Xcode guard and soft-fail semantics are unchanged from the version reviewed in #360.

## Alternatives considered and rejected

- **Gate on an LLVM artifact-cache probe** (`lookup-only` restore before cleanup): precise, but requires duplicating the cache key outside `build-llvm` (silent-divergence risk) and has a probe→restore eviction race that reintroduces exactly the ENOSPC flake class #329 fixed — for ~2 extra minutes on ARM64 only.
- **Background rm + reap step**: biggest theoretical win, but backgrounded cleanup loses exit-code observability (the pattern flagged in #360 review) and needs cross-step PID/status plumbing.

## Cold-build validation (run [29155284630](https://github.com/NomicFoundation/solx/actions/runs/29155284630))

A temporary commit (since dropped) salted the macOS LLVM artifact-cache key so both legs built LLVM cold, and sampled available disk every 15 s across the whole job:

| Leg | Baseline | Job minimum | Peak consumption |
|---|---|---|---|
| x86 — cleanup **skipped** (108 GiB stock) | 108 GiB | 56.9 GiB | **~51 GiB** |
| ARM64 — cleanup stopped at target | ~163 GiB post-cleanup | 125.9 GiB | **~37 GiB** |

The worst measured cold peak (~51 GiB, x86) independently reproduces the ~52 GiB estimate above, so the 100 GiB target holds with ~2x headroom; ARM64 starting from exactly 100 GiB would have bottomed out around 63 GiB free. The builds were ccache-warm, which doesn't change the disk profile — ccache only skips compiler CPU time; the full build tree, install prefix, and ccache dir still land on disk.

The new step logic also behaved as designed on the cold path: x86 skipped in ~1 s ("Already >= 100 GiB available; skipping cleanup"), and ARM64 met the target in 1m14s via the simulator-runtime unmounts plus one Xcode batch ("Removed 4 of 18 inactive Xcode version(s); 163 GiB now available").


.github/workflows/test.yaml
```diff
@@ -201,17 +201,38 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
+
```
hedgar2017: This job has become quite long, almost half of the workflow. Can it be outlined to the free-disk-space actions similarly to the one for Linux above?

hedgar2017 (changes requested):
Thanks!

.github/workflows/test.yaml
nebasuke: Good idea, done!

hedgar2017 (approved):

PR #545, opened 2026-07-14
Integration macos large hack

nebasuke:


hedgar2017:
Left this on `main-solc` after the trunk swap (`main-slang` is now `main`). Rebasing onto `main` blows up on the process files, but only because the branch still carries the pre-review draft of what merged as #524 — the rest of it is on the trunk too, and the leftover delta actually reverts main's pool error handling rather than adding to it. The one commit here that is genuinely not on `main` is `6ac861ae` (the large macOS runners hack). Probably easiest to cherry-pick that onto `main` and close this.

PR #546, opened 2026-07-14
ci: debug solx-tester on macOS runners (main baseline for #524 mac validation)

nebasuke:
# ⚠️ Temporary debug PR — do not merge

## Why

#524 (persistent worker subprocesses) needs macOS validation: worker-thread changes have a history of regressing badly on Mac while looking fine on Linux. #545 tried running the integration suite on large macOS runners on top of #524, and `solx-tester` was **signal-killed within seconds on `macos-15-large`** and **ran for hours without finishing on `macos-15-xlarge`** — but that run can't distinguish "solx-tester has never worked on macOS" from "#524 regressed on macOS". Notably, on the Intel runner **main's own binaries crashed identically** in the baseline-benchmark step, so at least the Intel crash predates #524.

This branch is plain `main`, so whatever it does on macOS **is** the macOS baseline:

- if main's tester also crashes/hangs here → the mac failures are pre-existing, #524 is off the hook (and we've found a real solx-tester mac bug);
- if main runs clean → re-point this branch at #524's tip and the diff is the regression, with diagnostics showing where.

## What it does

One push-triggered workflow (`mac-tester-debug.yaml`), matrix over `macos-15`, `macos-15-large`, `macos-15-xlarge`:

- **No C++ builds**: `build-llvm`/`build-solc` are called with test.yaml's exact macOS parameters, so both hit the finished-install caches saved on `main` (`v1-llvm-macOS-*`, `v1-solc-macOS-*`). Only the Rust workspace compiles; an iteration is ~15–20 min.
- **Signals get names**: `solx-tester` is invoked directly instead of via `solx-dev test solx-tester` — the wrapper collapses a signal death to `subprocess failed without exit code`; here the step prints e.g. `killed by signal 11 (SIGSEGV)`.
- **Symbols kept**: release binaries are built with `strip = none` + line tables so crash reports and `sample` output are readable; LLVM is RelWithDebInfo with assertions on.
- **Staged runs**: environment probe → single-test/single-thread smoke → full suite under a 30-minute watchdog that `sample`s the tester and its `--recursive-process` workers before killing them (for the xlarge-style hang).
- **Diagnostics**: new `.ips` crash reports from `~/Library/Logs/DiagnosticReports` and all `sample` outputs are printed and uploaded as a per-runner artifact.

Uses the `free-disk-space-macos` action from #533 (just merged), which with warm artifact caches usually skips the cleanup entirely.

### DIFF
--- a/.github/workflows/mac-tester-debug.yaml
+++ b/.github/workflows/mac-tester-debug.yaml
@@ -1,0 +1,166 @@
+name: Mac solx-tester debug
+
+# Temporary debugging workflow: get solx-tester running on macOS runners.
+# Context: solx-tester has no macOS CI coverage anywhere; PR #545 showed it
+# being signal-killed on macos-15-large and hanging on macos-15-xlarge, but
+# with PR #524 included, so mac breakage and #524 regressions were
+# indistinguishable. This branch is plain `main`, so its results are the
+# macOS baseline.
+#
+# The build-llvm/build-solc parameters deliberately mirror test.yaml's macOS
+# legs so both hit the finished-install caches saved on main (v1-llvm-macOS-*,
+# v1-solc-macOS-*): nothing is compiled but the Rust workspace.
+
+on:
+  push:
+    branches:
+      - mac-tester-debug
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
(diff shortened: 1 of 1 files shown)


hedgar2017:

Force-pushed `mac-tester-debug`: `9d037377` → `19bbbaff`. No conflicts, and `git range-diff` reports all four commits identical (`=`) — the patches came through unchanged.

PR #548, opened 2026-07-15
solx-tester: drop the archived `web3` crate for alloy-primitives

nebasuke:
* Migrate web3 to alloy-primitives. Note this also unblocks the renovate pin on reqwest. 
* Fix up  solx-solc-test-adapter to adapt to the stricter hex parsing semantics in alloy-primitives

# Claude summary
web3 was used only by the test-only crates solx-tester and solx-solc-test-adapter, purely for its U256/Address/keccak256 types — never by the compiler. Replace them with alloy-primitives, which is already in the tree via revm (revm::primitives re-exports it).

- solx-tester uses revm::primitives::{U256, Address} (no new dep); the web3<->revm bridge helpers in revm_type_conversions collapse to identities and are removed, as are a few now-redundant byte round-trips.
- solx-solc-test-adapter gains a direct alloy-primitives dependency.
- Small helpers (u256_from_hex_str / address_from_hex_str) preserve web3's lenient hex parsing (optional 0x prefix, no EIP-55 checksum), since alloy's FromStr is stricter (decimal / checksum-validated).

solx-compiler-downloader relied on web3's transitive reqwest to supply a TLS backend via cargo feature unification; add an explicit "rustls-tls" feature so HTTPS downloads keep working on their own.

Also drop the rust-web3 git-source allowlist entry (deny.toml), the web3 pinned-fork rule and the now-obsolete reqwest 0.12.x pin (renovate.json), and the dead revm/balance.rs stub.


solx-solc-test-adapter/src/lib.rs
```diff
@@ -42,13 +42,42 @@ const NEW_LINE: &str = "\n";
...
-    address.to_big_endian(&mut bytes);
-    web3::types::Address::from_slice(
+    let bytes = address.to_be_bytes::<{ solx_utils::BYTE_LENGTH_FIELD }>();
+    Address::from_slice(
         &bytes[solx_utils::BYTE_LENGTH_FIELD - solx_utils::BYTE_LENGTH_ETH_ADDRESS..],
     )
 }
+
+///
+/// Parses a hexadecimal string (with optional `0x` prefix, no checksum validation) into a `U256`.
```
hedgar2017: These are now duplicates in both crates. Can they be reused from the adapter? Also worth double-checking if we actually need to reimplement them.

solx-solc-test-adapter/src/lib.rs
```diff
@@ -42,13 +42,42 @@ const NEW_LINE: &str = "\n";
...
+    U256::from_str_radix(value, 16)
+        .map_err(|error| anyhow::anyhow!("Invalid hexadecimal literal `{value}`: {error}"))
+}
+
+///
+/// Parses a hexadecimal address (with optional `0x` prefix, no checksum validation) into an `Address`.
+///
+/// Preserves the lenient hex parsing that `web3`'s `Address::from_str` previously provided.
+///
+pub fn address_from_hex_str(value: &str) -> anyhow::Result<Address> {
```
hedgar2017: Ditto.

hedgar2017 (changes requested):

hedgar2017:
@nebasuke thanks! LGTM minus duplicates.

solx-solc-test-adapter/src/lib.rs
nebasuke: Indeed!. Doing it in phases at the moment, semantics preserving rewrite -> integration test run, reduce the adapter -> integration test run.

nebasuke: Alright, with some minor changes on the solx-solc-test-adapter I've managed to get rid of the helper functions instead.

hedgar2017 (approved):

PR #549, opened 2026-07-15
refactor(slang): scope-strata lowering architecture

hedgar2017:
## Architecture

The Slang frontend lowers Solidity through three typed emission scopes. The source unit scope owns the MLIR context and the registry of lowered function signatures; the contract scope borrows it and carries the contract body, its state variables, and their storage layout; the function scope borrows that and carries one body's variable environment, declared return types, and the checked-arithmetic flag. Every piece of emission state is owned by the scope that mutates it.

Lowering is expressed as inherent methods on these scopes, one module per language construct, each method named after the Slang node it lowers, with hand-written dispatch for each evaluation position an expression can occupy: value, value list, place, and effect. Function calls are classified once by callee kind, and each kind is emitted by its own routine. Type projection belongs to the source unit scope and is reached from the function scope through thin forwarders.

The MLIR crate no longer depends on the Slang frontend crate, keeping all Slang-facing bookkeeping in the frontend. The standard JSON layer provides a constructor for MLIR contract output, and the workspace lock matches main except for that dependency removal.

## Verification

All unit tests pass, the frontend side of the MLIR lit suite passes in full, and the integration tester shows no regressions against main, with one previously failing test now passing.


solx-mlir/src/context/function.rs
```diff
@@ -35,6 +35,9 @@ pub struct Function<'context> {
 }
 
 impl<'context> Function<'context> {
+    /// The `mlir_name` of a synthesized parameterless constructor.
+    pub const CONSTRUCTOR_NAME: &'static str = "@constructor()";
+
```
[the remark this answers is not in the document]
hedgar2017: All good. This is Slang convention, unrelated to solc conventions used in other pipelines.

solx-slang/src/contract/function/expression/call/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+                    unimplemented!(
+                        "callee '{}' does not resolve to a function",
+                        identifier.name()
+                    );
+                };
+                Self::Function(function_definition)
+            }
+            Expression::MemberAccessExpression(access) => {
+                match access.member().resolve_to_built_in() {
+                    Some(BuiltIn::AbiDecode) => Self::Builtin(BuiltIn::AbiDecode),
```
ggiraldez: Is there any reason to special case `abi.decode`? It seems it could be lowered in `member_builtin`.

solx-slang/src/contract/function/expression/call/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+                    _ => Self::MemberBuiltin(access),
```
ggiraldez: I think this may miss some (admittedly edge) valid use cases if `resolve_to_built_in()` returns `None`:

```
contract C {
    struct X {
        function() x;
    }
    function test() internal {
        X memory x;
        x.x();  //  <---
    }
}
```

I don't know if the intention is to handle this in `member_builtin` but it's at least an inconsistency in the naming.

solx-slang/src/contract/function/expression/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+pub mod index_access;
+pub mod keyword;
+pub mod literal;
+pub mod logical;
+pub mod member;
+pub mod tuple;
+pub mod unary;
+
+use slang_solidity_v2::ast::Expression;
+use slang_solidity_v2::ast::Type as SlangType;
```
ggiraldez: nit: I've seen slang's `ast::Type as SlangType` aliased in a couple of files, but for the most part the alias is used on `solx_mlir::Type as MlirType`. For consistency and readability, I'd stick to aliasing one of them.

solx-slang/src/contract/function/expression/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+use solx_mlir::Value;
+
+use crate::scope::function::FunctionScope;
+
+use self::call::Call;
+
+impl<'contract, 'source_unit, 'context> FunctionScope<'contract, 'source_unit, 'context> {
+    /// Lowers an expression to its single MLIR value, routing each kind to its lowering. A call in
+    /// value position takes its one result.
+    pub fn expression(&mut self, node: &Expression) -> Value<'context> {
```
ggiraldez, line 35: Probably not for this PR, but as a small optimization and also to avoid having to represent rational numbers in MLIR, you could query the expression's `type()` to make use of the constant folding that Slang already does and avoid traversing the expression sub-trees. If the computed type is a literal, it should contain the numeric result of the folded constant as an integer.

ggiraldez (approved):
Left some minor comments but it looks good to me. I like the general code organization and at least as far as the Slang lowering code, I find it very easy to read.

solx-mlir/src/ir/value.rs
```diff
@@ -66,25 +55,24 @@ impl<'context> Value<'context> {
...
+        Self::constant(i64::from(value), Type::boolean(context.melior), context)
     }
 
-    /// Coerces to `target_type` per Solidity conversion semantics; a no-op at the target type.
-    ///
-    /// A boolean target compares against zero rather than bit-truncating; an address target
-    /// truncates integers to ui160 before `sol.address_cast`; every other target is a `sol.cast`.
+    /// Coerces to `target_type` under Solidity's implicit conversions with a `sol.cast`; a no-op at
+    /// the target type.
     pub fn coerce(self, target_type: Type<'context>, context: &Context<'context>) -> Self {
```
abinavpp, line 63: inline the impl to convert() and remove coerce()?

solx-mlir/src/ir/value.rs
```diff
@@ -93,21 +81,26 @@ impl<'context> Value<'context> {
...
-        self.cast(target_type, context)
+        self.coerce(target_type, context)
     }
 
-    /// Compares against `other` under `predicate`, first coercing both operands to a common integer
-    /// type: their shared type when equal, otherwise the 256-bit field.
+    /// Compares against `other` under `predicate`, the operands first reconciled to a common type:
+    /// their shared type when equal, otherwise the wider operand's integer type, preserving its
+    /// signedness.
     pub fn compare_coerced(
```
abinavpp, line 90: I think slang's semantics already has a `common_type` in its type registry, and comparison typing carries a TODO (SDR[55]) for checking operand compatibility. Hoping we can consume the binder's common type here instead once that semantic work lands.

solx-slang/src/contract/function/expression/assignment.rs
```diff
@@ -0,0 +1,64 @@
...
+use slang_solidity_v2::ast::AssignmentExpression;
+use slang_solidity_v2::ast::AssignmentExpressionOperator;
+
+use solx_mlir::Value;
+
+use crate::scope::function::FunctionScope;
+
+impl<'contract, 'source_unit, 'context> FunctionScope<'contract, 'source_unit, 'context> {
+    /// `lhs = rhs` and the compound `lhs op= rhs`, lowered through the left operand's place. The
+    /// right operand is fully evaluated before the place's old value is read, matching solc's
```
abinavpp, line 14: let's update the compound assignment test to check for the eval order.

solx-slang/src/contract/function/expression/call/mod.rs
hedgar2017: Indeed! It's some refactoring leftover. Let me fix this together with #2.

solx-slang/src/contract/function/expression/call/mod.rs
hedgar2017: Good catch. Should be just `Member` without `BuiltIn`.

solx-slang/src/contract/function/expression/mod.rs
hedgar2017: Thanks, incomplete sweep!

solx-slang/src/contract/function/expression/mod.rs
hedgar2017: Yes, thanks! That will be in constant folding PRs.

abinavpp (commented):
sorry for the delay, lgtm! just minor nits that we could address later

solx-mlir/src/ir/value.rs
hedgar2017: This is actually an artifact of pre-emptive refactoring. In the subsequent PRs, `coerce` owns the implicit type conversion, and `convert` owns the explicit one. Semantically, implicit is a subset of explicit, I think it's true for Solidity, please correct me if I'm wrong!

solx-mlir/src/ir/value.rs
hedgar2017: Yes, will be happy to migrate, created an issue.

solx-slang/src/contract/function/expression/assignment.rs
hedgar2017: Good idea, let me land it into #552!

hedgar2017: Added!

Issue #550, opened 2026-07-15
Windows: fresh llvm-sys builds cannot link against the bundled static libstdc++.a under MSYS2 GCC 16 — CI is one cache-fingerprint change from breaking


nebasuke:
## Summary

Any Windows build that actually recompiles `llvm-sys` fails to link `solx` with:

```
ld.exe: target-llvm/target-final/lib/libstdc++.a(string-inst.o): multiple definition of
  `std::__cxx11::basic_string<…>::_M_replace_cold(…)';
  libllvm_sys-*.rlib(ExecutionEngine.cpp.obj): …/mingw64/include/c++/16.1.0/bits/basic_string.tcc:545: first defined here
collect2.exe: error: ld returned 1 exit status
```

This affects dev and release profiles alike. `test.yaml`'s Windows leg is currently green **only** because its Swatinem cargo cache keeps `llvm-sys` "Fresh" — the moment anything dirties `llvm-sys`'s cargo fingerprint (profile/env change, `RUSTFLAGS`, an llvm-sys or inkwell bump, a cache eviction), Windows CI breaks with no code change to blame. We hit this while building a cross-OS solx-tester workflow (cross-OS plan, Stage 1).

## Root cause

`build-llvm/action.yml` copies MSYS2's `libstdc++.a` into the LLVM install tree (`LIBSTDCPP_SOURCE_PATH` → `target-llvm/target-final/lib/libstdc++.a`, handled by `solx-dev/src/llvm/platforms/x86_64_windows_gnu.rs`) so release binaries can link C++ statically and be self-contained. Because `llvm-sys`'s build script emits `-L target-llvm/target-final/lib`, that *static* archive shadows MSYS2's own `-lstdc++` resolution (which would prefer the shared `libstdc++.dll.a`) for every crate link in the workspace.

Under MSYS2's current GCC 16, this is fatal for freshly compiled C++ objects: GCC 16 added `basic_string::_M_replace_cold` and emits it as a COMDAT in user objects, which collides with the strong definition in the static archive's explicit-instantiation objects (`string-inst.o` / `wstring-inst.o`) whenever those members get pulled into the link.

**It is not version drift.** We ran the controlled experiment:

| Bundled `libstdc++.a` | llvm-sys objects | Result |
|---|---|---|
| from cache (older GCC) | fresh, GCC 16.1.0, release | ❌ multiple definition ([run](https://github.com/NomicFoundation/solx/actions/runs/29415024519)) |
[...]

Issue #547, opened 2026-07-14
solx-tester SIGSEGV on macOS x86-64: rustc 1.95/1.96 enum-niche miscompilation (rust-lang/rust#159035) — fix by bumping to 1.97.1


nebasuke:
I read about https://github.com/rust-lang/rust/issues/159035 a few days ago on Hackernews, and now I regret thinking it would be unlikely enough to apply to us. So here you go, trying to get the integration tests to run on both Mac runners actually triggered it. 

## Claude Summary

`solx-tester` is killed by SIGSEGV within seconds on **macOS x86-64** whenever it processes the Ethereum semantic-test corpus. The root cause is not in solx or solx-tester: it is a **rustc/LLVM enum-niche miscompilation** present in our pinned toolchain (1.96.1) and in 1.95.0, tracked upstream as [rust-lang/rust#159035](https://github.com/rust-lang/rust/issues/159035) (P-critical, `I-miscompile`, LLVM fix merged upstream, **1.97.1 point release with the LLVM bump confirmed by the release team**).

**Proposed fix: bump `rust-toolchain.toml` to 1.97.1 as soon as it ships.** No code changes needed. The debug branch below verifies the bump in a single push.

## Impact

- macOS x86-64 is the only platform where we have observed the crash; it fires deterministically on the first Ethereum-corpus `Call` input, single-threaded or multi-threaded, so solx-tester effectively cannot run on Intel macs today. (This has no CI coverage, which is why it went unnoticed — it surfaced while validating #524 on macOS runners, PR #545.)
- Per the upstream analysis, the miscompile family either crashes loudly or copies data into memory that Rust considers uninitialized (never validly read), so Linux/arm builds passing our suites are unlikely to be silently affected — but staying on an affected toolchain is not comfortable.

## Crash signature

```
exception: EXC_BAD_ACCESS (EXC_I386_GPFLT), "UNKNOWN_0xD at 0x0000000000000000"
frame #0: solx-tester`try_from_ethereum [inlined] new at runtime.rs:54:9 [opt]
->  0x1000ee62c <+3420>: movaps %xmm0, -0xf8(%rbp)
```

[...]

PR #553, opened 2026-07-15
ci: cross-OS solx-tester suite on free macOS/Windows runners (ci:cross-os label + weekly cron)

nebasuke:
## What

A label-gated + weekly-cron workflow running the **full `solx-tester` suite on free macOS (arm64 + Intel) and Windows runners**, plus the `run-tester-suite` composite action it is built from. The workflow consumes the `build-toolchain` composite (landed separately in #581) for its standard-config LLVM/solc prologue, so it adds no toolchain-build steps of its own.

- **Triggers**: PR label `ci:cross-os` (on-demand, for platform-sensitive changes like process/IPC work), `workflow_dispatch`, and a Monday 06:00 UTC cron on main as a drift net. Advisory only — must never become a required check.
- **Scope**: solx-tester only. Cross-OS legs test *platform behavior* (process management, IPC, paths, toolchain quirks); output quality stays on the Linux integration suite, and Hardhat/Foundry are deliberately not replicated.
- **Cost**: $0 — `macos-15`, `macos-15-intel`, and `windows-2025` are free for public repos; LLVM/solc come from the finished-install caches saved on main, so nothing heavier than the Rust workspace compiles.

## Depends on

**#581 (merged)** — extracted the standard-config toolchain prologue (submodules → SFW → msys → macOS-disk → LLVM → solc) into the shared `build-toolchain` composite. This workflow is its sixth consumer; the only change to the composite here is adding `cross-os-tests.yaml` to its consumer-list description. #581 also folded in #550's fix option 2, so the bundled static `libstdc++.a` drop now happens *inside* `build-toolchain` for every standard-config Windows build — this workflow inherits the working shared-`libstdc++` link with no step of its own.

## Why

One week of evidence that this layer catches what Linux cannot:

- the #524 worker-pool hang on macOS (present at an intermediate tip, fixed before merge — confirmed by this harness's watchdog samples, [PR comment](https://github.com/NomicFoundation/solx/pull/524#issuecomment-4980190960));
- the rustc 1.96.1 enum-niche miscompile killing solx-tester on Intel macs (#547, since fixed by the 1.97.1 bump in #556 — the Intel leg here is what keeps verifying it, since test.yaml's Intel leg never runs solx-tester);
- the Windows libstdc++ link fragility (#550), found the day this workflow first ran solx-tester on Windows — which had **zero prior executions** there; it passed 74,605 cases once the `libstdc++` drop (now in `build-toolchain`) and the corpus-dir retention below were in place.

## How it works

`run-tester-suite` (composite action): builds solx + solx-tester (symbols kept), smoke test, then the suite under a parametrized watchdog. A run that neither crashes nor finishes gets diagnosed before the kill — on macOS, `sample` stacks of the tester, the top-level solx parents, *and* the pooled workers, plus new crash reports from `DiagnosticReports`; on Windows, a `tasklist` snapshot. Signal deaths are reported by name (`killed by signal 11 (SIGSEGV)`) instead of the `solx-dev` wrapper's "subprocess failed without exit code". Stage outcomes are aggregated at the end so diagnostics always run.

Windows specifics:
- the bundled static `libstdc++.a` drop (#550) is **inherited from `build-toolchain`**, applied to all standard-config Windows test builds — not a step this workflow adds;
- keep `solx-solidity/test/` through the disk cleanup (the Ethereum corpus is read from it at run time) — this workflow's own `Free disk space (Windows)` step.

## Verification

[Run 29429860239](https://github.com/NomicFoundation/solx/actions/runs/29429860239): `macos-15` and `windows-2025` legs green, full suites — ~22 min and ~37 min job time respectively. That run predates the rebase onto `build-toolchain`, but the composite encodes byte-identical LLVM/solc params, so the tester results carry over; a fresh `ci:cross-os` run on the rebased branch confirms the composite path.

The `macos-15-intel` leg has no post-1.97.1 timing yet (pre-fix runs exceeded 75 min without finishing), so its watchdog (110 min) and job timeout (240 min) are a probe budget rather than the ~2x margin the other legs get. First runs establish the baseline; then tighten, or scope down to `--path tests/solidity` if the full suite can't finish on the free Intel minis.

## Follow-ups (not in this PR)

- Tune the Intel leg's watchdog/timeout once real runs establish a full-suite baseline (see Verification).
- #550's durable fix (option 1): stop bundling `libstdc++.a` into the LLVM tree in solx-dev and make release packaging link it statically on its own. That's a solx-dev + release.yaml change whose verification surface is a release build, so it stays out of CI-only work like this. When it lands, `build-toolchain`'s drop step gets deleted — and its plain `rm` fails loudly if the bundling disappears first, so the workaround can't linger silently.
- The `mac-tester-debug*` / `win-tester-spike` branches and #546 get retired once this lands.

### DIFF
--- a/.github/actions/build-toolchain/action.yml
+++ b/.github/actions/build-toolchain/action.yml
@@ -4,10 +4,11 @@ description: >-
   checkout, Socket Firewall, platform prep (MSYS2 on Windows, disk cleanup
   on macOS), then the LLVM and solc builds. The build parameters below
   define the standard-config cache keys shared by test.yaml,
-  slang-tests.yaml, and cache-warmup.yaml — changing them here re-keys all
-  of those workflows together. Workflows needing a different LLVM
-  configuration (sanitizer, coverage, integration, release) call
-  build-llvm/build-solc directly and pair with their own cache-warmup job.
+  slang-tests.yaml, cross-os-tests.yaml, and cache-warmup.yaml — changing
+  them here re-keys all of those workflows together. Workflows needing a
+  different LLVM configuration (sanitizer, coverage, integration, release)
+  call build-llvm/build-solc directly and pair with their own cache-warmup
+  job.
 
 inputs:
   llvm:
--- a/.github/actions/build-toolchain/action.yml
+++ b/.github/actions/build-toolchain/action.yml
@@ -3,11 +3,12 @@ description: >-
   Shared CI prologue for the standard toolchain configuration: submodule
   checkout, Socket Firewall, platform prep (MSYS2 on Windows, disk cleanup
   on macOS), then the LLVM and solc builds. The build parameters below
-  define the standard-config cache keys shared by test.yaml and
-  cache-warmup.yaml — changing them here re-keys both workflows together.
-  Workflows needing a different LLVM configuration (sanitizer, coverage,
-  integration, release) call build-llvm and build-solc directly as needed
-  and pair with their own cache-warmup job.
+  define the standard-config cache keys shared by test.yaml,
+  cross-os-tests.yaml, and cache-warmup.yaml — changing them here re-keys
+  all of those workflows together. Workflows needing a different LLVM
+  configuration (sanitizer, coverage, integration, release) call build-llvm
+  and build-solc directly as needed and pair with their own cache-warmup
+  job.
 
 inputs:
   llvm:
--- a/.github/actions/run-tester-suite/action.yml
+++ b/.github/actions/run-tester-suite/action.yml
@@ -1,0 +1,174 @@
+name: Run solx-tester suite
+description: |
+  Builds solx and solx-tester and runs a smoke test plus a (optionally
+  filtered) suite under a watchdog. A run that neither crashes nor finishes
+  is diagnosed before being killed: on macOS the tester and its solx
+  processes are `sample`d and new crash reports are collected; elsewhere the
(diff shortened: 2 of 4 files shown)


hedgar2017:

Force-pushed `cross-os-tester`: `3eead44a` → `f126ab21`. One conflict, in `.github/actions/build-toolchain/action.yml`, and I resolved it — **please check this, it is your prose.**

The conflict is the description comment. Your version added `cross-os-tests.yaml` to the list `test.yaml, slang-tests.yaml, cross-os-tests.yaml, and cache-warmup.yaml`, but `slang-tests.yaml` no longer exists — the slang line folded it into `test.yaml`, and the trunk had also reworded the surrounding sentence. I resolved to the trunk wording with your addition kept and `slang-tests.yaml` dropped:

```
  define the standard-config cache keys shared by test.yaml,
  cross-os-tests.yaml, and cache-warmup.yaml — changing them here re-keys
  all of those workflows together. Workflows needing a different LLVM
  configuration (sanitizer, coverage, integration, release) call build-llvm
  and build-solc directly as needed and pair with their own cache-warmup
  job.
```

`git range-diff` shows two adaptations and both are benign: this comment, and a `CLAUDE.md` hunk whose context shifted because the trunk had already removed the `ci:slang` label line above your `ci:cross-os` addition. The third commit is identical (`=`). Your own additions are unchanged.

PR #554, opened 2026-07-16
fix(tester): resolve the Slang deploy target in source order

hedgar2017:
Makes slang-ast tester measurement honest:

- The deploy target is resolved from the Slang AST in source order over all three deployable node types (contract, library, interface), replacing the alphabetical pick that deployed the wrong contract in multi-contract sources; failed resolution is now an error instead of a fallback.
- Tests the Slang frontend cannot reproduce are filtered out: `pragma abicoder v1` sources (Slang always encodes v2) and the Yul test tree. All additions are gated on `slang-ast`; the default tester is unchanged.
- The `<path>:<name>` contract identifier is owned by `ContractName::full_path` and routed through it at all tester call sites.

### DIFF
--- a/solx-tester/src/compilers/output_ext.rs
+++ b/solx-tester/src/compilers/output_ext.rs
@@ -67,15 +67,13 @@ pub fn get_last_contract(
                 }
             }
 
-            // TODO: Slang frontend produces a CST instead of the solc AST, so
-            // `last_contract_name` cannot extract the name from the AST
-            // `nodes` array. Fall back to the contracts map directly.
             #[cfg(feature = "slang-ast")]
             for (path, _source) in sources.iter().rev() {
-                if let Some(contracts) = output.contracts.get(path) {
-                    if let Some((name, _)) = contracts.last_key_value() {
-                        return Ok(format!("{path}:{name}"));
-                    }
+                if let Some(contracts) = output.contracts.get(path)
+                    && let Some(source) = output_sources.get(path)
+                    && let Some(name) = last_slang_object_name(source, contracts)
+                {
+                    return Ok(format!("{path}:{name}"));
                 }
             }
 
@@ -176,4 +174,32 @@ fn last_contract_name(
         )
         .next_back()
         .ok_or_else(|| anyhow::anyhow!("The last contract not found in the AST"))
+}
+
+///
+/// Returns the name of the last deployable object in the Slang AST that was compiled: a contract,
+/// library, or interface. The Slang AST gives each its own node type, unlike the solc AST where
+/// all three are `ContractDefinition`s distinguished by `contractKind`.
+///
+#[cfg(feature = "slang-ast")]
+fn last_slang_object_name(
+    source: &solx_standard_json::output::source::Source,
+    contracts: &BTreeMap<String, solx_standard_json::output::contract::Contract>,
+) -> Option<String> {
+    source
+        .ast
+        .as_ref()?
+        .get("members")?
+        .as_array()?
+        .iter()
+        .rev()
+        .filter(|member| {
+            matches!(
+                member["type"].as_str(),
+                Some("ContractDefinition" | "LibraryDefinition" | "InterfaceDefinition")
+            )
+        })
+        .filter_map(|member| member["name"]["text"].as_str())
+        .find(|name| contracts.contains_key(*name))
+        .map(str::to_owned)
 }
--- a/solx-tester/src/compilers/output_ext.rs
+++ b/solx-tester/src/compilers/output_ext.rs
@@ -8,10 +8,11 @@ use std::collections::HashMap;
 use solx_standard_json::InputLanguage;
 use solx_standard_json::Output;
 use solx_standard_json::OutputError;
-#[cfg(feature = "slang-ast")]
-use solx_standard_json::output::contract::Contract;
 use solx_standard_json::output::source::Source;
 use solx_utils::ContractName;
+
+#[cfg(feature = "slang-ast")]
+use crate::compilers::solidity::slang_ast::SlangAst;
 
 ///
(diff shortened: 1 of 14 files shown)


solx-tester/src/compilers/solidity/mode.rs
```diff
@@ -71,9 +75,38 @@ impl Mode {
     }
 
     ///
-    /// Checks if the mode is compatible with the source code pragmas.
+    /// Checks if the mode is compatible with the source code pragmas. Under the Slang frontend,
```
OmarTawfik, line 80: Checking source code lines/whitespace splitting can lead to both false positives (processing commented out pragmas) and false negatives (missing pragmas that have other trivia in-between). I suggest doing this with a tree visitor/walk for correctness.

Note that a future version of Slang will be emitting errors on `pragma abicoder v1` to make this clear to the user. If this is blocking you, we can also prioritize this diagnostics group on our side.

solx-tester/src/compilers/output_ext.rs
```diff
@@ -177,3 +179,31 @@ fn last_contract_name(
...
+) -> Option<String> {
+    source
+        .ast
+        .as_ref()?
+        .get("members")?
+        .as_array()?
+        .iter()
+        .rev()
+        .filter(|member| {
+            matches!(
```
OmarTawfik: I wonder if there is a technical reason why the Slang AST is not used directly, to get type checking/exhaustive matching on these nodes. Is there something we can improve on our end?

hedgar2017: Thanks for the suggestion!

Just for historical reasons actually.

solx-tester has been using raw standard JSON and never serialized into static types.
Let me check how to do it cleanly!

solx-tester/src/compilers/solidity/mode.rs
hedgar2017: Would be great! I'm not sure if I'd like a diagnostic at once. Maybe you could expose a getter with enum so I can just query AST for the encoder type and decide what to do with it?

hedgar2017: Walked manually for now. Please take a look!

solx-tester/src/compilers/output_ext.rs
hedgar2017: @OmarTawfik please check! Turns out there is no overhead in the release build.

There may be some duplication because due to some dependency issues solx-tester cannot depend on solx-slang, but I hope we'll be able to unify it one day when solc is gone.

solx-tester/src/compilers/solidity/mode.rs
OmarTawfik: > I'm not sure if I'd like a diagnostic at once

The correctness concern is that our own ABI API only returns V2, so any input specifying V1 would return unexpected/wrong results. My understanding is that both solx/slang have no plans to support v1. Is that correct?

The diagnostic itself won't block any other analysis, and you should be able to match on the error kind and ignore it if needed. Thoughts?

solx-tester/src/compilers/solidity/slang_ast.rs
```diff
@@ -0,0 +1,164 @@
...
+                };
+                match directive.pragma() {
+                    Pragma::AbicoderPragma(pragma) => match pragma.version() {
+                        AbicoderVersion::AbicoderV1Keyword(_) => v1_pinned = true,
+                        AbicoderVersion::AbicoderV2Keyword(_) => v2_declared = true,
+                    },
+                    Pragma::ExperimentalPragma(pragma) => {
+                        if matches!(
+                            pragma.feature(),
+                            ExperimentalFeature::ABIEncoderV2Keyword(_)
```
OmarTawfik: I suggest also checking the `ExperimentalFeature::StringLiteral` form.
It should be consolidated when we finish the pragma diagnostics area.

solx-tester/src/compilers/solidity/slang_ast.rs
```diff
@@ -0,0 +1,164 @@
...
+                        ) {
+                            v2_declared = true;
+                        }
+                    }
+                    Pragma::VersionPragma(_) => {}
+                }
+            }
+        }
+        v1_pinned && !v2_declared
```
OmarTawfik: nit: if both are true, it should be invalid input/error in both solx/slang. Do you support such tests?
If not, why checking `v2_declared`?

solx-tester/src/compilers/output_ext.rs
OmarTawfik: Looking great!

OmarTawfik (approved):
Thanks!
Left one more comment on [checking string feature names](https://github.com/NomicFoundation/solx/pull/554#discussion_r3613034246). But otherwise, LGTM!

solx-tester/src/compilers/solidity/mode.rs
hedgar2017: Yes, all v1 tests are ignored!

From what I remember from prior discussions, no plans to support it indeed, but I may be having outdated info!

solx-tester/src/compilers/solidity/slang_ast.rs
hedgar2017: Turned out I can drop the v2 checks entirely. I made them because some multi-file tests have mixed encoding apparently, but it looks like I should just ignore them all.

Thanks!

solx-tester/src/compilers/solidity/slang_ast.rs
hedgar2017: Turned out I can drop the v2 checks entirely. I made them because some multi-file tests have mixed encoding apparently, but it looks like I should just ignore them all.

Thanks!

PR #556, opened 2026-07-16
chore: bump Rust toolchain to 1.97.1

### LINKED ISSUE #547: solx-tester SIGSEGV on macOS x86-64: rustc 1.95/1.96 enum-niche miscompilation (rust-lang/rust#159035) — fix by bumping to 1.97.1

nebasuke:
Closes #547.

Bumps the pinned Rust toolchain to **1.97.1**, which carries the upstream LLVM fix for the enum-niche miscompilation that was crashing `solx-tester` on macOS x86-64.

# Claude summary

## Why

`solx-tester` is killed by SIGSEGV within seconds on **macOS x86-64** whenever it processes the Ethereum semantic-test corpus. The root cause is not in solx: it's a rustc/LLVM enum-niche miscompilation present in 1.95/1.96 ([rust-lang/rust#159035](https://github.com/rust-lang/rust/issues/159035), P-critical `I-miscompile`), resolved by the LLVM point-bump in the **1.97.1** release. See #547 for the full autopsy (lldb crash site, toolchain probes showing 1.95/1.96 crash and fixed-LLVM nightly clean).

No source changes are needed — this is a toolchain pin bump.

## What changed

| File | Change |
|---|---|
| `rust-toolchain.toml` | `1.96.1` → `1.97.1` (drives native runners; resolves lazily in-container) |
| `.github/docker/Dockerfile` | `ARG RUST_VERSION` → `1.97.1`, keeping the baked CI-image toolchain aligned with the pin; updated the adjacent LLVM comment |
| `CLAUDE.md` | toolchain line |

## Coverage-job safety check

The Dockerfile comment flags that the coverage job's `llvm-cov`/`llvm-profdata` (system LLVM 21) must read the profraw format emitted by instrumented binaries. 1.97.1 embeds **LLVM 22.1.6** vs 1.96.1's **22.1.2** — same major, so the profraw format compatibility is unchanged. No coverage-tooling bump required.

## CI runner image

This PR also bumps `ARG RUST_VERSION`, so the `solx-ci-runner` image is rebuilt and the pinned digest repinned across the workflows that run in-container:

- [ ] Rebuild the image (`docker-ci-image.yaml`) and repin the runner digest (currently `sha256:2c8db6ea…3d896583`, 19 references across `sanitizer`, `test`, `release`, `integration-tests`, `compile-benchmark`, `slang-tests`, `cache-warmup`, `coverage`).

Until the repin lands, in-container jobs still get 1.97.1 — the container resolves `rust-toolchain.toml` lazily on first use — at the cost of a per-job toolchain download.

## Verification

- [ ] Green on the `mac-tester-debug` `macos-15-large` leg with the real 1.97.1 toolchain (no nightly hack) — confirms the Intel SIGSEGV is gone (#547 checklist).
- Local `cargo check` is green under 1.97.1 for the core crates + `solx-codegen-evm`.

### DIFF
--- a/.github/docker/Dockerfile
+++ b/.github/docker/Dockerfile
@@ -51,7 +51,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 # ── LLVM 21 (system toolchain for coverage, sanitizers, lipo) ────────────────
 # llvm-cov / llvm-profdata must read the profraw format emitted by
-# instrumented Rust binaries. rustc 1.96.1 embeds LLVM 22 — one major ahead
+# instrumented Rust binaries. rustc 1.97.1 embeds LLVM 22 — one major ahead
 # of these tools — but still emits format versions LLVM 21 reads. Re-check on
 # every toolchain bump: a format bump breaks the coverage job with an opaque
 # profdata error.
@@ -182,7 +182,7 @@ RUN curl -fsSL -o valgrind-${VALGRIND_VERSION}.tar.bz2 \
 # weight in the dev image. rust-src serves the -Zbuild-std actions
 # (build-rust, rust-unit-tests). On a bump, re-check the LLVM section
 # comment in base (rustc's embedded LLVM major moves).
-ARG RUST_VERSION=1.96.1
+ARG RUST_VERSION=1.97.1
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
         | sh -s -- -y --default-toolchain "${RUST_VERSION}" --profile default \
     && . "$HOME/.cargo/env" \
--- a/.github/workflows/cache-warmup.yaml
+++ b/.github/workflows/cache-warmup.yaml
@@ -38,10 +38,10 @@ jobs:
             runner: macos-15
           - name: "Linux x86 gnu"
             runner: ubuntu-24.04
-            image: ghcr.io/nomicfoundation/solx-ci-runner@sha256:2c8db6eea1393f7b1824d632597f9ac3f8ca7ae34d63b301e80c11a13d896583
+            image: ghcr.io/nomicfoundation/solx-ci-runner@sha256:cd5a37f2630fdf1898ddb2ca11f8c2ecc4d572c60011fe923b27b1625f92ba15
           - name: "Linux ARM64 gnu"
             runner: ubuntu-24.04-arm
-            image: ghcr.io/nomicfoundation/solx-ci-runner@sha256:2c8db6eea1393f7b1824d632597f9ac3f8ca7ae34d63b301e80c11a13d896583
+            image: ghcr.io/nomicfoundation/solx-ci-runner@sha256:cd5a37f2630fdf1898ddb2ca11f8c2ecc4d572c60011fe923b27b1625f92ba15
           - name: "Windows"
             runner: windows-2025
     runs-on: ${{ matrix.runner }}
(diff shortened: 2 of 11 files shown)


nebasuke:
Docker image build: https://github.com/NomicFoundation/solx/actions/runs/29504609619

hedgar2017 (approved):

PR #561, opened 2026-07-17
fix(cli): support Windows absolute paths in --libraries

futurehua:
## Summary

Fix `--libraries` parsing for library specifiers that contain Windows absolute paths.

Library specifiers use the following format:

```text
<ContractPath>:<ContractName>=<LibraryAddress>
```

The parser previously split the path at the first :, which caused the drive separator in a value such as C:\contracts\Math.sol:Math=... to be interpreted as the separator before the contract name.

The parser now splits at the final :, preserving the complete source path while continuing to extract the contract name correctly.

A regression test covers a library specifier containing a Windows absolute path.

### DIFF
--- a/solx-utils/src/libraries.rs
+++ b/solx-utils/src/libraries.rs
@@ -75,12 +75,8 @@ impl TryFrom<&[String]> for Libraries {
             let path = path_and_address
                 .next()
                 .ok_or_else(|| anyhow::anyhow!("Library #{index} path is missing."))?;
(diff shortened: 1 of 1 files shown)


nebasuke (approved):
Thanks for your contribution, looks good! 

@hedgar2017 Any concerns?

hedgar2017 (approved):
Hmm nice one! Thank you sir @futurehua!

PR #562, opened 2026-07-17
ci: verify CLI docs examples against the release binary

nebasuke:
Add regenerable examples in our docs using `trycmd`, so we don't end up manually regenerating them every time. 

Regenaration is done via:

```
TRYCMD=overwrite cargo test -p solx --test mod docs_examples
```

I had the change the format to fit this new regenerable approach, and I tried to keep it the same as much as possible while keeping a good reading experience. 
## Claude Summary

Every output block in the CLI user guide is generated from input files that were never committed, and the examples have silently drifted across releases: the 0.1.4 regeneration was made from a pre-tag build, so its examples matched neither the released 0.1.4 binary nor 0.1.5. The drift #558 fixes was found by hand; this PR makes the check mechanical so it cannot recur.

**The guide itself is the test definition.** The examples are [trycmd](https://docs.rs/trycmd) console cases: every ```console block in `docs/src/user-guide/02-command-line-interface.md` runs exactly as documented — `$ solx …` invokes the binary under test, `$ ls '<dir>'` lists its artifacts — in a temporary sandbox seeded from `02-command-line-interface.in/` (the committed fixtures: `Simple.sol` recovered byte-for-byte against the source keccak256 embedded in the documented metadata example, plus `Simple.yul`/`Simple.ll` so the `--yul`/`--llvm-ir` examples are real). There is no parallel table of invocations to maintain, and newly documented examples are covered automatically. Because it's part of the CLI suite, it runs in `test.yaml` on every PR across all platforms — including docs-only PRs, for which a dedicated paths filter forces `build-and-test` when the guide or its fixtures change.

`...` on its own line elides a run of output lines, `[..]` matches anything within a line, and `? failed` documents a non-zero exit. Blank lines and whitespace are compared exactly. The guide itself now opens with this legend and with the `Simple.sol`/`Simple.yul` sources via `{{#include}}`, so the page is followable by a reader, not just by the harness.

## Doc reshaping for the trycmd dialect

All output-preserving (validated against the release binary at conversion time; the snapshot tracks 0.1.6):

- Usage/output block pairs merge into single ```console blocks with `$` prompts — the dialect mdBook highlights as a shell session (same convention as the Cargo book).
- Benchmark timings read `solx_Compilation: [..]ms` instead of a plausible literal value.
- The IR-emission sections (`--evmla`, `--ethir`, `--emit-llvm-ir`) write to per-section output dirs: all commands of the page share one sandbox, and `./build/` listings would otherwise accumulate earlier sections' artifacts. The legend states this.
- The `--overwrite` refusal example shows its non-zero exit as a `? failed` line, and the previously untested `--overwrite` usage block became a live example: it runs against the `./build/` dir populated by the `--output-dir` section, and now says so.
- One honesty fix the exact matcher forced: `--ast-json` shows the trailing empty contract section it always emitted (the old elision matcher never checked output tails).

## Review round

hedgar's review caught two harness-level defects and a set of assertion-depth gaps; all addressed:

- **The `$ ls` steps never ran.** trycmd resolves only registered binaries — it never searches `PATH` — so the five listing steps were unresolvable and silently counted as passing (`Skipped` is `ok` in trycmd's book). `ls` is now resolved from `PATH` at test time and registered like `solx`; the first live comparison immediately caught the one real drift (the hand-written listings lacked the trailing newline `ls` emits). All 35 steps of the guide now execute.
- **Docs-only PRs skipped the test.** The guide lives under `docs/`, which the `code` filter excludes, so a PR editing only the guide skipped `build-and-test` entirely. A `filter-docs-examples` paths filter is now OR'd into the job condition and the `pr-checks` allowed-skip expression.
- A fence-count assertion guards trycmd's other silent-green mode: it exits 0 when it recognizes zero cases, e.g. after a fence-tag restructuring.
- `--bin`/`--bin-runtime` pin the full bytecode except the 34-byte metadata digest (which hashes the solc commit hash and therefore differs between dev and release builds). Abbreviated AST and DWARF payloads say so in prose; real DWARF structure is verified in `solx/tests/debug_info/`.
- The asm emitter padded mnemonics to a fixed width even with no value following, leaving load-bearing trailing spaces in the documented `--asm-solc-json` output. Fixed at the source: padding is emitted only when a value follows.
- The CBOR trailer's version bytes stay **literal** by design: version-stamp drift is precisely the historical failure mode (#558), and the pinned bytes are what makes the `ci:release` dry-run prove the release binary stamps the tagged version. The cost is a two-line doc update per release, handled by direct edit rather than `TRYCMD=overwrite` (documented in the developer guide) so the digest wildcard on the same line survives.
- `toml_edit` moved to 0.25.13 — the cooldown that motivated the earlier lockfile hold-back has elapsed. (The previous description's "toml_edit/toml_writer pinned one patch back" was wrong for `toml_writer`, which was already at the latest release.)

## Regenerating after an intentional output change

```bash
TRYCMD=overwrite cargo test -p solx --test mod docs_examples
```

rewrites the stale blocks in place, then you review the doc diff. Full instructions live in `docs/src/developer-guide/01-testing.md`, including the two caveats: overwrite expands inline `[..]` wildcards on lines that no longer match (restore benchmark timings and DWARF hex by hand), and version bumps should edit the CBOR trailer hex directly instead of blessing — the deterministic version bytes share their line with the build-dependent digest wildcard.

## Validation

All 35 steps of the guide execute and pass against a locally built 0.1.6 binary — none skipped — run repeatedly for determinism, sandbox left pristine. Full CLI suite (406 tests) green; clippy, rustfmt, actionlint, and cargo-cooldown-check clean; mdBook build confirms the `{{#include}}` directives render. Pending: a `ci:release` dry-run re-validates the guide against actual release binaries — the first with #565's `.mod`-free Windows solc build.

## What Windows needed

The first CI round failed only on Windows — three platform deltas, covered by one commit:

- The CRLF checkout rewrote `Simple.sol`, changing its keccak256 and every documented output derived from it (metadata source hash and urls, AST byte offsets, CBOR ipfs digests). Fixed with a `.gitattributes` LF pin on the fixture dirs.
- LLVM prints `.file "Simple"` where other platforms show `.file "Simple.sol:Simple"` — the colon in the module name parses as a Windows drive separator. Wildcarded.
- The Windows solc build stamped its version `0.8.34+commit.<hash>.mod`, shifting the documented version string and — because the metadata JSON embeds it — the ipfs digests in the `--metadata`/`--metadata-hash`/`--ir` examples. Wildcarded here; the stamp itself (a released-binary reproducibility problem beyond this test) was removed by #565, since merged. The wildcards stay regardless: the commit-hash field they absorb changes on every solx-solidity re-pin, as do the digests hashing it.

The previous revision claimed the Windows job's MSYS2 environment provides GNU `ls` via `PATH` — wrong in a useful way, since trycmd never consults `PATH` on any platform and the `ls` steps were being skipped everywhere (see the review round above). They are now resolved explicitly at test time, which on Windows finds MSYS2's GNU `ls`.


solx/tests/cli/docs_examples.rs
```diff
@@ -0,0 +1,28 @@
...
+//! line elisions, but inline `[..]` wildcards in rewritten regions are
+//! expanded to the literal output and must be restored by hand — watch for
+//! benchmark timings and for working-directory paths hex-encoded inside
+//! DWARF output, which differ on every run.
+//!
+
+#[test]
+fn docs_examples() {
+    trycmd::TestCases::new()
+        .register_bin(
```
hedgar2017, line 31: Only `solx` is registered, so the five `$ ls '<dir>'` steps in the guide (lines 331, 363, 396, 496, 873) never run, and the test still reports green.

Traced through the pinned `trycmd 1.2.0`. `BinRegistry::resolve_name` (`registry.rs:44-56`) tries the registered map, then `cargo_bin_opt(name)` — a Cargo-built binary in the target dir. It never searches `PATH`, deliberately. So `ls` comes back as an unresolved `Bin::Name`, and `run_step` hits this arm before `to_command` is ever called:

```rust
// runner.rs:371-375
Some(crate::schema::Bin::Name(_name)) => {
    // Unhandled by resolve
    snapbox::debug!("bin={:?} not found", _name);
    assert_eq!(output.spawn.status, SpawnStatus::Skipped);
    return Ok(output);
}
```

`Ok`, not `Err` — and `SpawnStatus::is_ok` counts `Skipped` as passing (`runner.rs:747-750`). The step renders as `ignored` in the log while the case reports `ok`, and the expected listing written under each `$ ls` is compared against nothing. `prior_step_failed` is only set on `Err`, so the following `solx` steps run normally and nothing looks wrong.

Net effect: every `--output-dir` / `SOLX_OUTPUT_DIR` example asserts only that the preceding `solx` invocation exited 0. The empty `.out/` does not compensate — its comparison is subset-matching, so files present in the sandbox but not listed there are ignored, not flagged.

Fix: drop the `ls` blocks and express the assertion as real `.out/` fixtures, which is trycmd's native "these files exist" mechanism and also sidesteps the GNU/BSD/Windows `ls` formatting differences that `LC_ALL=C` was presumably meant to tame. Registering a resolved `ls` path works too, but keeps the formatting exposure.

solx/tests/cli/docs_examples.rs
```diff
@@ -0,0 +1,28 @@
...
+            "solx",
+            assert_cmd::cargo::cargo_bin!(env!("CARGO_PKG_NAME")).to_path_buf(),
+        )
+        .env("LC_ALL", "C")
+        .case("../docs/src/user-guide/02-command-line-interface.md");
```
hedgar2017: Two gaps in what this line can catch.

A rename or move of the `.md` fails loudly, which is right. But if the file stays and its ```console fences stop being recognized — a fence language tag changed, the blocks restructured — trycmd takes the empty branch at `runner.rs:46`, prints "There are no trycmd tests enabled yet", and exits 0. The suite goes green with nothing run. Assert a nonzero recognized-case count before the run; there is no public trycmd API for that today, so it is worth filing upstream, with a grep for at least one ```console fence as the interim.

Separately, this test does not run on the PRs it exists to protect. `build-and-test` is gated on `needs.changes.outputs.code == 'true'` (`.github/workflows/test.yaml:178`), and the `changes` filter excludes `!docs/**` and `!**/*.md`. The guide and all of `.in/` live under `docs/`, so a PR that edits only them yields `code == 'false'`, the job skips, and `pr-checks` permits it (`test.yaml:314`). Exactly the drift this harness was built to catch. This PR itself only escapes because it also touches `Cargo.toml`, `Cargo.lock`, and Rust files.

Fix: mirror the existing `filter-devcontainer` pattern in `test.yaml` — a second `dorny/paths-filter` output scoped to the guide and its `.in`/`.out` dirs, OR'd into `build-and-test`'s condition and the `pr-checks` allowed-skip expression.

solx/tests/cli/docs_examples.rs
```diff
@@ -0,0 +1,28 @@
+//!
```
hedgar2017, line 1: Sibling files in this directory (`bin.rs:1-3`, `overwrite.rs:1-3`) use a terse three-line `//!`. This is a 17-line essay, and most of it restates what the code does rather than a why the code cannot express.

Concretely: "Checks that every example in the CLI user guide produces the output the guide shows" restates the function name; lines 4, 5 and 14 carry em dashes and a non-ASCII ellipsis; and the `TRYCMD=overwrite` caveat at lines 11-16 is written near-verbatim again in `.in/README.md:27-39`, so the two will drift.

Match the sibling shape:

```rust
//!
//! CLI tests for the user guide examples.
//!
```

and move the regeneration caveat to `docs/src/developer-guide/01-testing.md`, which currently says nothing about `docs_examples`. While there: that chapter and the root `CLAUDE.md` both give the entry point as `cargo test --test cli`, but the integration target is `mod` — as this PR's own regeneration command correctly uses. Both are already stale.

One more on the caveat's content: it claims `TRYCMD=overwrite` expands inline `[..]` wildcards in rewritten regions. trycmd only expands a wildcard-bearing line when that specific line fails to match, so the blanket phrasing overstates what happens.

solx/Cargo.toml
```diff
@@ -37,6 +37,7 @@ assert_cmd = "2.2"
 predicates = "3.1"
 tempfile = "3.27"
 test-case = "3.3"
+trycmd = "1.2"
```
hedgar2017, line 40: `trycmd` is correctly a dev-dependency and does not reach the shipped binary.

The `toml_edit`/`toml_writer` "pinned one patch back to clear the cooldown check" from the PR description has no expression here. It exists only as `Cargo.lock` entries, with no `Cargo.toml` constraint and no `renovate.json` rule, so the next `cargo update` silently reverts it and the cooldown check comes back with no record of why.

The claim also doesn't hold for both crates: `toml_writer 1.1.2` is the latest published version, not held back at all. `toml_edit` is genuinely one patch behind (0.25.12 vs 0.25.13, about seven weeks apart).

Either express the pin where it is durable — `toml_edit = "=0.25.12"` with a stated rationale and an expiry, or a cooldown-allowlist entry — or drop the "pinned back" framing from the description, since as written it is neither durable nor accurate.

.gitattributes
```diff
@@ -4,3 +4,8 @@
 # The fixtures get the same pin so the comparison is LF on both sides.
 solx-benchmark-converter/templates/*.md text eol=lf
 solx-benchmark-converter/src/tests/fixtures/*.md text eol=lf
+
+# The CLI-guide fixtures are hashed in documented compiler output (the
+# metadata example embeds Simple.sol's keccak256), so CRLF checkouts on
+# Windows must not rewrite their line endings.
+docs/src/user-guide/02-command-line-interface.in/** text eol=lf
```
hedgar2017, line 12: Correct and load-bearing: `git check-attr` confirms this pins `Simple.sol`, `Simple.yul`, `Simple.ll`, `README.md`, and `.keep` to LF, and the rationale is real — `Simple.sol` hashes to the `keccak256` documented at guide line 121 (`0x402fe0b3…`, verified byte-for-byte), which a CRLF checkout would break.

The rule stops at `.in/**`. `.out/` and the `.md` case file are uncovered. Harmless today, since `.out/` holds only an empty `.keep` and snapbox normalizes line endings on the `.md` comparison — but the moment `.out/` gains real fixtures the same Windows failure returns. Add the `.out/**` rule now rather than rediscovering it.

docs/src/user-guide/02-command-line-interface.in/README.md
```diff
@@ -0,0 +1,39 @@
+# CLI documentation examples
```
hedgar2017, line 1: `.in/.keep` should be deleted. It exists to make git track an otherwise-empty directory, and this directory now holds four real tracked files.

`.out/.keep` is load-bearing and should stay, but not for the reason the layout suggests: the mere existence of a sibling `.out/` is what switches trycmd's sandbox on, which is what keeps commands from writing artifacts into the checkout. An empty `.out/` does **not** assert that the sandbox ends empty — the comparison is `snapbox::dir::PathDiff::subset_matches_iter`, so unlisted files are ignored. Worth one sentence here, since the empty directory reads as an emptiness assertion to anyone who finds it.

docs/src/user-guide/02-command-line-interface.md
```diff
@@ -28,16 +28,12 @@ The rest of this section describes the available CLI options and their usage. Yo
 
 Emits the full bytecode.
 
-```bash
-solx 'Simple.sol' --bin
-```
+```console
```
hedgar2017, line 46: The trycmd dialect is never explained to the reader. `[..]` first appears at line 36, bare `...` at line 72, `? failed` at line 528 — all before any explanation exists, and the only explanation lives in `.in/README.md`, which nobody reading the published guide will ever see.

Add a short legend before this first console block covering `$`, `[..]`, `...`, and `? failed`.

docs/src/user-guide/02-command-line-interface.md
```diff
@@ -28,16 +28,12 @@ The rest of this section describes the available CLI options and their usage. Yo
...
+$ solx 'Simple.sol' --bin
 
-Output:
-
-```text
 ======= Simple.sol:Simple =======
 Binary:
-34601557630000008480630000001a...
+34601557630000008480630000001a[..]
```
hedgar2017: This pins the first ~15 bytes of bytecode; the dispatcher, both function bodies, and the CBOR trailer are all wildcarded. Same at line 50 for `--bin-runtime`.

The `--metadata-hash ipfs` block at line 666 shows the tighter assertion is achievable: it pins ~95 literal bytes and wildcards only the unavoidable 34-byte digest. Nothing about the bytecode here is hash- or path-dependent, so the same depth should apply.

The `--asm` block at lines 64-83 has the same problem for a different reason: both function bodies are elided by bare `...` although assembly text is fully deterministic. That truncation is doc brevity, not a constraint the harness imposes.

docs/src/user-guide/02-command-line-interface.md
```diff
@@ -143,16 +128,15 @@ Metadata:
...
+```console
+$ solx 'Simple.sol' --ast-json
 
-Output:
-
-```text
 ======= Simple.sol =======
 JSON AST:
-{"absolutePath":"Simple.sol","exportedSymbols":{"Simple":[24]},"id":25,"license":"MIT","nodeType":"SourceUnit","nodes":[ ... ],"src":"32:288:0"}
+{"absolutePath":"Simple.sol","exportedSymbols":{"Simple":[24]},"id":25,"license":"MIT","nodeType":"SourceUnit","nodes":[[..]],"src":"32:288:0"}
```
hedgar2017, line 153: `"nodes":[[..]]` wildcards the entire AST body — the one thing `--ast-json` exists to produce. What remains asserted is the envelope and the byte offsets.

If the full AST is too long for the page, pin a structural prefix rather than the whole node array, or move the real assertion into a CLI unit test and keep the doc block illustrative with a sentence saying so.

docs/src/user-guide/02-command-line-interface.md
```diff
@@ -272,19 +238,15 @@ Developer Documentation:
...
-Output:
+```console
+$ solx 'Simple.sol' --asm-solc-json
 
-```text
 ======= Simple.sol:Simple =======
 EVM assembly:
 000     PUSH                80
-001     MEMORYGUARD
+001     MEMORYGUARD
```
hedgar2017: This line and line 249 carry 9 and 14 trailing spaces respectively, and the test compares whitespace exactly, so any editor or pre-commit hook that strips trailing whitespace breaks the build.

The root cause is compiler-side: the mnemonic is padded to a fixed width whether or not a value follows (`solx-evm-assembly/src/assembly/instruction/mod.rs:389-391`):

```rust
match self.name {
    Name::Tag => write!(f, "{name:4}"),
    _ => write!(f, "{name:20}"),
}?;
```

Fix the emitter to pad only when a value follows, then re-bless. Load-bearing invisible whitespace in a committed doc is not a stable place to leave this.

docs/src/user-guide/02-command-line-interface.md
```diff
@@ -329,16 +292,12 @@ object "Simple_24" {
...
-
-Output:
+```console
+$ solx 'Simple.sol' --debug-info
 
-```text
 ======= Simple.sol:Simple =======
 Debug info:
-7f454c46010201ff...
+7f454c46010201ff[..]
```
hedgar2017, line 319: All that survives here is the 8-byte ELF magic `7f454c46010201ff`; the whole DWARF payload is wildcarded. Same at line 314 for `--debug-info-runtime`.

The embedded working-directory path genuinely cannot be pinned, so there may be no cheap fix — but then say so plainly in the prose rather than leaving a block that reads like an output assertion, and put the real debug-info verification in `solx/tests/cli/debug_info.rs` where it can decode structure.

docs/src/user-guide/02-command-line-interface.md
```diff
@@ -369,28 +324,20 @@ When used with `--output-dir`, writes `.evmla` files to the output directory. Wi
...
-```
+```console
+$ solx 'Simple.sol' --evmla --output-dir './build-evmla/'
+Compiler run successful. Artifact(s) can be found in directory "./build-evmla/".
 
-Output:
-
-```text
-Compiler run successful. Artifact(s) can be found in directory "./build/".
+$ ls './build-evmla/'
```
hedgar2017, line 354: Three things converge on this line.

The `$ ls` steps never execute — see the comment on `docs_examples.rs:22`. This listing is decorative.

`./build-evmla/` and the sibling `./build-ethir/`, `./build-llvm-ir/`, `./build-overwrite/` dirs exist only because all commands on the page share one sandbox and `./build/` listings would otherwise accumulate earlier sections' artifacts. That is a harness constraint presented to readers as CLI guidance, and the `--output-dir` and `--overwrite` sections still use plain `./build/`, so the page now teaches two conventions. Pick one, and if the per-section dirs stay, say in a sentence that they exist for test isolation.

Related: the `--overwrite` success example at line 516 only demonstrates overwriting because an earlier, unrelated section happened to populate `./build/` first. It does not stand on its own if read or run in isolation.

And the guide never shows `Simple.sol`, `Simple.yul`, or `Simple.ll` anywhere in its prose, though every example compiles them. The page is now reproducible for the harness but not for a human following along. Pull them in with `{{#include}}` so the sources are visible.

docs/src/user-guide/02-command-line-interface.md
```diff
@@ -727,16 +658,13 @@ Specifies the hash format used for contract metadata.
...
-
-Output with `ipfs`:
+```console
+$ solx 'Simple.sol' --bin --metadata-hash 'ipfs'
 
-```text
 ======= Simple.sol:Simple =======
 Binary:
-34601557630000008480630000001a6080396080f35b5f5ffdfe34600b57600336116016575b5f5ffd5b5060016031565b5f3560e01c633df4ddf48114600f57635a8ac02d03600b5760025b60805260206080f3fea2646970667358221220a1c9557d1ba161f8b10cbf71bbb227befd112b91b438365ac59448a2b9f0d74c64736f6c637816736f6c783a302e312e363b736f6c633a302e382e33340047
+34601557630000008480630000001a6080396080f35b5f5ffdfe34600b57600336116016575b5f5ffd5b5060016031565b5f3560e01c633df4ddf48114600f57635a8ac02d03600b5760025b60805260206080f3fea2646970667358221220[..]64736f6c637816736f6c783a302e312e363b736f6c633a302e382e33340047
```
hedgar2017, line 695: The solx and solc versions are baked in literally as hex (`736f6c783a302e312e363b736f6c633a302e382e3334` here and at line 809), so every version bump forces a re-bless of the doc tests — which is how this snapshot already drifted once. Wildcarding just the version bytes keeps the CBOR structure pinned without the churn.

hedgar2017 (changes requested):
Inline comments below.

solx/tests/cli/docs_examples.rs
nebasuke: Confirmed at the source level and fixed. `ls` is now resolved from `PATH` at test time and registered like `solx`, so all five listings execute. Went with registration over `.out/` fixtures because the listings are documentation: `.out/` fixtures are invisible to readers of the published guide, and committed artifact contents would re-churn on every release. The formatting exposure is smaller than it looks — `ls` writing to a pipe prints one filename per line on GNU, BSD, and MSYS2 alike, and `LC_ALL=C` pins the sort order.

The fix proved itself immediately: the first live comparison failed because the hand-written listings lacked the trailing newline `ls` actually emits. Content was otherwise accurate; all five listings now compare for real.

solx/tests/cli/docs_examples.rs
nebasuke: Both fixed. The gating gap: a dedicated `filter-docs-examples` paths filter (the guide plus its `.in/` and `.out/` dirs) is now OR'd into `build-and-test`'s condition, and `pr-checks` only allows its skip when neither code nor the guide changed — same shape as `filter-devcontainer`.

The zero-cases mode: the harness now asserts the guide contains at least one ```` ```console ```` fence before handing it to trycmd. Agreed that a recognized-case count from trycmd itself would be the real fix and is worth an upstream issue; the fence assertion covers the realistic failure (restructured or retagged fences) in the meantime.

solx/tests/cli/docs_examples.rs
nebasuke: Trimmed to near the sibling shape — two extra lines point at trycmd and the regeneration doc. The caveat moved to `docs/src/developer-guide/01-testing.md`, which now has a docs-examples section, and the stale `cargo test --test cli` entry points are fixed to `--test mod` there and in `CLAUDE.md` — good catch, both were broken. The overwrite caveat is reworded to say wildcards are expanded only on lines that no longer match.

solx/Cargo.toml
nebasuke: Right on both counts — `toml_writer` 1.1.2 is the latest, so the description's claim was simply wrong there. In the meantime `toml_edit` 0.25.13 cleared the 7-day cooldown, so the hold-back is gone entirely: bumped in the lockfile, verified locally with cargo-cooldown-check, and the pin story is dropped from the PR description.

.gitattributes
nebasuke: Added the `.out/**` rule.

docs/src/user-guide/02-command-line-interface.in/README.md
nebasuke: `.in/.keep` deleted. The README now states explicitly that an empty `.out/` does not assert the sandbox ends empty (the comparison is subset-matching) — its only job is switching the sandbox on.

docs/src/user-guide/02-command-line-interface.md
nebasuke: Added a legend before the first block covering `$`, `[..]`, `...`, and `? failed` — plus the `Simple.sol` source itself via `{{#include}}`, so the page is followable by a human now, not just by the harness.

docs/src/user-guide/02-command-line-interface.md
nebasuke: Pinned both blocks to the `--metadata-hash` shape: full literal bytecode with only the 34-byte digest wildcarded.

One correction to the premise, though it doesn't change the conclusion: the bytecode here *is* hash-dependent. The trailer's ipfs digest hashes metadata that embeds the solc commit hash, which differs between dev builds and release binaries, so the digest `[..]` has to stay — exactly as in the `--metadata-hash` block.

The `--asm` body elisions stay: that one is deliberate doc brevity — the section's point is the shape of the assembly output, not the full instruction stream.

docs/src/user-guide/02-command-line-interface.md
nebasuke: Kept illustrative, but the block now says so: a sentence after it notes the `nodes` array is abbreviated and that it runs ~4.5k characters even for this contract. Pinning a structural prefix inside a one-line JSON blob buys little here — the envelope (ids, byte offsets, source name, license) is the part that drifts in practice, and it is fully pinned.

docs/src/user-guide/02-command-line-interface.md
nebasuke: Done — both blocks now state in prose that only the ELF magic is pinned, because the DWARF payload embeds the absolute build directory and is not reproducible documentation material.

The structural verification you're asking for already exists outside the guide: `solx/tests/debug_info/` parses the emitted `.debug_line` section with the `object` crate and pins line attribution end-to-end, which is a stronger check than any doc block could carry.

docs/src/user-guide/02-command-line-interface.md
nebasuke: Fixed in this PR rather than separately: the mnemonic is padded only when a value follows, so no-value instructions no longer emit trailing spaces, and the two doc lines are re-blessed. Agreed that load-bearing invisible whitespace in a committed doc was a trap — it's gone.

docs/src/user-guide/02-command-line-interface.md
nebasuke: All four taken:

- The `ls` steps are real now — see the thread on the harness file.
- The per-section dirs stay, and the new intro legend explains why in one sentence (all commands share one sandbox, so per-section dirs keep each listing scoped).
- The `--overwrite` success example now states it reuses the `./build/` dir populated by the `--output-dir` section.
- `Simple.sol` and `Simple.yul` are pulled in with `{{#include}}` (rendering verified with mdBook). `Simple.ll` stays by reference — 57 lines of LLVM IR would swamp its section — with a provenance sentence instead: it is `Simple.sol`'s own `--emit-llvm-ir` output.

docs/src/user-guide/02-command-line-interface.md
nebasuke: Keeping these literal, deliberately. The version stamp is the one drift this guide has actually suffered — the 0.1.4 examples were regenerated from a pre-tag build (#558) — so the pinned trailer bytes are exactly the assertion that catches it, and they are what makes the `ci:release` dry-run prove the release binary stamps the version being tagged. Wildcarding them would let a mis-stamped release binary pass the docs test.

The churn is real but bounded: two lines, once per release, and release PRs already touch the version in the docs (#585 did five spots). At our cadence that is a cost worth carrying.

The part of the comment I did act on is the blessing footgun, which is worse than stated: the trailer shares its line with the `[..]` ipfs digest wildcard, so `TRYCMD=overwrite` on a version bump expands the digest to a literal that only breaks weeks later, on the next solx-solidity re-pin. The developer guide's new regeneration section now says to update the trailer hex directly for version bumps (the bytes are deterministic) instead of blessing via overwrite.

If the per-release bless ever becomes a burden, the escape hatch that keeps the assertion is `TestCases::insert_var` with the solx part derived from `CARGO_PKG_VERSION` — noting it here so we don't reach for `[..]` later.

hedgar2017 (approved):

PR #563, opened 2026-07-17
feat(slang): complete lvalues and assignment

hedgar2017:
Completes lvalues and assignment on the Slang frontend:

- A destructuring assignment flattens its left targets against the right operand's values and stores them pairwise, every value evaluated before any target is written so `(a, b) = (b, a)` swaps; a nested tuple flattens into the same list, and a blank target discards its paired value.
- A plain assignment through a place that is its own address — a storage or calldata reference — emits `sol.copy`, which bridges both the element type and the data location of its right operand.
- The ternary conditional yields a tuple by storing each arm into per-element stack slots, loaded after the merge.
- Prefix and postfix `++`/`--` step through any computed place.
- An uninitialized scalar declaration or named return zero-fills by folding to a typed constant.
- `delete` resets its operand in place, clearing a storage aggregate through `sol.delete` and storing the element type's zero into every other place.

### DIFF
--- a/solx-slang/src/contract/function/expression/assignment.rs
+++ b/solx-slang/src/contract/function/expression/assignment.rs
@@ -18,7 +18,16 @@ impl<'contract, 'source_unit, 'context> FunctionScope<'contract, 'source_unit, '
     pub fn assignment(&mut self, node: &AssignmentExpression) -> Option<Value<'context>> {
         let places = self.expression_places(&node.left_operand());
         if places.len() > 1 {
-            let values = self.expression_values(&node.right_operand());
+            let targets: Vec<_> = places
+                .iter()
+                .map(|&element| {
+                    let Some((place, element_type)) = element else {
+                        return None;
+                    };
+                    (place.r#type() != element_type).then_some(element_type)
+                })
+                .collect();
+            let values = self.coerced_values(&node.right_operand(), &targets);
             for (place, value) in places.into_iter().zip(values) {
                 let Some((place, element_type)) = place else {
                     continue;
@@ -26,7 +35,7 @@ impl<'contract, 'source_unit, 'context> FunctionScope<'contract, 'source_unit, '
                 if place.r#type() == element_type {
                     place.copy_from(value, self);
                 } else {
-                    place.store(value.coerce(element_type, self), self);
+                    place.store(value, self);
                 }
             }
             return None;
--- a/solx-slang/src/contract/function/expression/assignment.rs
+++ b/solx-slang/src/contract/function/expression/assignment.rs
@@ -10,17 +10,46 @@ use solx_mlir::Value;
 use crate::scope::function::FunctionScope;
 
 impl<'contract, 'source_unit, 'context> FunctionScope<'contract, 'source_unit, 'context> {
-    /// Plain `=` and the compound `op=` forms, lowered through the left operand's place. The right
-    /// operand is evaluated before the place's old value is read; `=` coerces and stores it without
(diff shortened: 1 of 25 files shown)


solx-slang/src/contract/function/expression/assignment.rs
```diff
@@ -4,23 +4,37 @@
...
+        if let Expression::TupleExpression(tuple) = &left
+            && tuple.items().len() > 1
+        {
+            let places = self.expression_places(&left);
+            let values = self.expression_values(&node.right_operand());
+            for (place, value) in places.into_iter().zip(values) {
+                if let Some((place, element_type)) = place {
+                    place.store(value.coerce(element_type, self), self);
+                }
+            }
```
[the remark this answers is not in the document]
hedgar2017: Thanks!

solx-slang/src/contract/function/expression/conditional.rs
```diff
@@ -3,23 +3,60 @@
...
 use crate::scope::function::FunctionScope;
 
 impl<'contract, 'source_unit, 'context> FunctionScope<'contract, 'source_unit, 'context> {
-    /// The ternary conditional operator: neither operand short-circuits, so there is no initializer
-    /// and both arms store their evaluated operand.
-    pub fn conditional(&mut self, node: &ConditionalExpression) -> Value<'context> {
-        let result_type = self.typing(node.get_type());
+    /// The ternary conditional operator. Neither operand short-circuits, so both arms evaluate their
+    /// element values and store them into per-element stack slots — one for a scalar result, one per
+    /// element for a tuple — loaded after the merge.
```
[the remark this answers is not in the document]
hedgar2017: Fixed!

solx-slang/src/contract/function/expression/assignment.rs
```diff
@@ -10,62 +10,75 @@ use solx_mlir::Value;
...
-        let (place, element_type) = self.expression_place(&node.left_operand());
+    /// `lhs = rhs` and the compound `lhs op= rhs`. A multi-element tuple left operand destructures and
+    /// yields nothing, tuple assignment being value-less; every other form yields the stored value. A
+    /// reference-typed place is its own address and takes a `sol.copy` of the right operand, which
+    /// bridges both its type and its data location. A shift compound keeps its right operand at its
+    /// own type, every other form coerces it to the place's type.
+    pub fn assignment(&mut self, node: &AssignmentExpression) -> Option<Value<'context>> {
+        let places = self.expression_places(&node.left_operand());
+        if places.len() > 1 {
+            let values = self.expression_values(&node.right_operand());
```
abinavpp: This misses the string-literal fold, so a literal in a destructuring assignment crashes (`sol.cast` on a `!sol.string` asserts):

```solidity
function f() public pure returns (bytes2, uint256) {
    bytes2 a;
    uint256 b;
    (a, b) = ("ab", 42);
    return (a, b);
}
```

The declaration form `(bytes2 a, uint256 b) = ("ab", 42);` works since it goes through `coerced_values` - this path should too.

abinavpp (commented):
claude and i found a potential bug, can you confirm?

solx-slang/src/contract/function/expression/assignment.rs
hedgar2017: Thanks, will check shortly!

hedgar2017: Fixed!

abinavpp (approved):
thank you!

PR #531, opened 2026-07-10
feat(test-harness): one machine-written integration PR summary

nebasuke:
PR adding the status of an integration test run as a comment. My goal is that it should be clear if an integration test was successful (with the goals in mind). In particular, the comment states the following:

* Whether it is output preserving -> good in case of a performance improvement that should not have impact on output
* Whether we have introduced new test failures -> generally should be a fail. 
* Numbers on compile time, gas reports along with sensible defaults to allow for jitter. 

The original reports are preserved, and are linked in the PR comment.

# Claude Summary
## Why

The integration workflow posted **three link-only PR comments**; all analysis of the XLSX reports happened offline (the local `solx-perf-review` skill's `parse_reports.py`). Nothing on the PR told you whether output changed, what failed, or how compile-time moved.

## What

A single machine-written comment (`integration-test-summary`), rendered **in-tree in Rust** from the converter's native `Benchmark` model — no XLSX round-trip, no Python. Designed so a reviewer can call pass/fail from the comment alone, using the same decision procedure as the offline review.

- **New `solx-benchmark-summary` binary** + a layered `solx-benchmark-converter/src/output/summary/` module: `toolchain` (declared per-matrix toolchain tables — the one place naming is interpreted), `stats` (per-suite numbers), `verdict` (typed ✅/⚠️/❌ decisions), `render` (Rust line-builders for every sentence and cell + a 46-line askama layout template, `templates/summary.md`).
- **`solx-dev`** now writes `{foundry,hardhat}-benchmark.json` next to each XLSX (one shared writer); the workflow emits `solx-tester-benchmark.json`. The summary reads those.
- The three link comments collapse into **one** comment (reusing `mshick/add-pr-comment` via `message-path`), also mirrored to `$GITHUB_STEP_SUMMARY`.
- **XLSX artifacts are unchanged** — still uploaded, still linked (by their real file names) from the summary table. The noisy-row blacklist (proxy fallbacks, brutalized multicalls) now filters in `Benchmark` itself instead of the XLSX writer, so the spreadsheet and the summary agree on the data set.

### The comment

Three independent verdict lines, each with its numbers:

[...]

PR #567, opened 2026-07-17
DO NOT MERGE: revert 6 perf PRs to test the summary on real regressions

hedgar2017:
Throwaway probe stacked on #531 to validate the machine-written integration summary against **real regressions**, by reverting the perf optimizations merged to `main` this week. Reverting a genuine optimization is exactly the class of change the summary exists to catch, far better than synthetic garbage (the earlier block-map probe showed nothing because it was a negligible slice of compile time).

## Reverted (newest first)
- #524 perf(core): reuse persistent worker subprocesses across translation units
- #502 perf(codegen): output-gating optimizations
- #503 perf(debug-info): debug info build optimizations
- #501 perf(evmla): redesign EVM assembly to offload the LLVM optimizer
- #504 perf(core): solc and subprocess payload optimizations
- #499 perf: optimize the EVM legacy assembly compilation pipeline

## Expected in the summary
The PR toolchain becomes the pre-optimization compiler, so it should show a clear **compile-time regression** (the process-pool and pipeline reverts are pipeline-wide) and quite possibly **size/gas** changes too (output-gating and the EVMLA-assembly reverts change generated bytecode). Multiple summary sections should light up.

## Compiles
`cargo check --all-targets` is clean on the whole default workspace. The only `cargo build` gap locally is the native `solc` static library, which this worktree does not build but CI does. One revert conflict (a debug-info CLI test touched by #532) was resolved to the pre-#503 version.

## Timeout note
Kept to `ci:integration` (standard set, not the full matrix): reverting six optimizations makes the PR toolchain compile slower, so the smaller contract set keeps the run inside the 300-minute job timeout. If it still runs long we dial back to fewer reverts.

## Not for merge
Pure test scaffolding. **Do not merge** — these are reverts of real, wanted optimizations.


hedgar2017:
Works as expected!

PR #571, opened 2026-07-18
perf(test-harness): clone and install each project once per run

nebasuke:
Stacked on #570, whose 3-line invariant-cap commit rides along in this diff and drops out when it merges.

## Claude Summary

The Foundry and Hardhat runners re-clone and re-install every project for each (compiler × codegen) pair — 4× per project on the PR matrix, 8× in full mode. For the Foundry suite that is dominated by repeated `git clone --recursive` and yarn installs; for Hardhat it is repeated npm/pnpm/bun installs. This PR hoists clone, package installs, and `forge config --fix` to once per project and reuses the checkout across toolchains.

The only toolchain-dependent inputs to the sources are the pragma sed and the config-version sed, and their patterns match their own output — so no checkout reset is needed; the seds simply re-run per toolchain with each compiler's version.

## What replaces the fresh clone's implicit guarantees

- **Dependency pragmas stay untouched.** Installs now precede the per-toolchain sed, so the `**/*.sol` glob skips `node_modules/` paths — same end state as before, where the install always followed the sed.
- **Hardhat config sed is made re-runnable.** It replaces the committed literal `0.8.30`, which is gone after the first pass, so the config file is restored from a pristine snapshot before each toolchain's sed (also correct under per-compiler `solidity_version` overrides).
- **Hardhat is forced to actually recompile.** `cache/` and `artifacts/` are deleted before every compile — Hardhat's cache cannot distinguish two solx binaries reporting the same base solc version, and would otherwise reuse the previous toolchain's artifacts. (Foundry already passes `--force` on the first build.)
- **Stale junit reports can't stand in for missing ones.** `junit-report.json` is deleted before each test run; previously the fresh clone guaranteed its absence, making a test run that produced no report fail loudly.

The compiler-identity shim from #529 remains the provenance backstop for any residual staleness.

## Measured effect

**Standard tier** — this PR's `ci:integration` run ([29658942544](https://github.com/NomicFoundation/solx/actions/runs/29658942544)) vs #570's run ([29657293509](https://github.com/NomicFoundation/solx/actions/runs/29657293509)) — the exact stacked baseline: same day, same standard tier, invariant cap in both. Foundry buckets from the harness status-line timestamps:

| Foundry bucket | #570 run | this PR | Δ |
|---|---|---|---|
| clones | 5.0 min | 1.3 min | **−3.8** |
| installs | 2.6 min | 0.9 min | **−1.7** |
| builds / tests / gas / size | 19.5 min | 19.2 min | ~0 (jitter) |

Foundry step 27.3 → 21.5 min, Hardhat 7.4 → 6.7 min, **job total 38.7 → 32.1 min (−6.6 min)**. The residual 1.3 min of cloning is the irreducible once-per-project fetch.

**Full tier** (`ci:integration-benchmark-full`) — this PR's full-matrix run ([29660324611](https://github.com/NomicFoundation/solx/actions/runs/29660324611)) vs the pre-#570 full-matrix baseline ([29643434263](https://github.com/NomicFoundation/solx/actions/runs/29643434263)), so the deltas include both this PR and #570's invariant cap:

| Foundry bucket | baseline | this PR | Δ |
|---|---|---|---|
| builds | 60.1 min | 60.1 min | 0 (per-toolchain identical to ±0.1 min) |
| clones | 10.2 min | 1.3 min | **−8.9** |
| gas passes | 10.2 min | 2.6 min | −7.6 (#570) |
| test passes | 7.7 min | 1.4 min | −6.3 (#570) |
| installs | ~4.7 min | ~1.0 min | **−3.7** |

Foundry step 93.6 → 66.9 min, Hardhat 17.3 → 15.5 min, **job total 114.8 → 86.4 min (−25%)**.

**Checkout reuse perturbed nothing measurable, on either tier** — every deterministic signal is identical to the fresh-clone baseline, including on the full matrix where one clone per project is shared across all four compiler lineages (solc, solx-latest, solx-main, solx-PR): 685 Foundry / 40 Hardhat pre-existing failures, 0 new; bytecode size identical across all 4,972 rows; gas 5,193 rows; solx-tester identical. Combined with #570: default `ci:integration` ~45 → ~32 min (−29%), full matrix ~115 → ~86 min.

## Local verification

End-to-end runs with two toolchains (both solc 0.8.34, reference + candidate, 2 codegens each), debug `solx-dev`:

- **Foundry** (`lil-web3`, `maple-erc20`): one clone per project, all 4 toolchains built/tested, 25 contract entries with size rows, exit 0. `lil-web3` viaIR shows the same 2 test failures on both identical toolchains — deterministic parity, gate correctly silent.
- **Foundry yarn path** (`diamonds`, `requires_yarn`): yarn install ran exactly once, all 4 toolchains built.
- **Hardhat** (`hardhat-v2` template, bun): one clone, `Compiled 1 Solidity file successfully` on all four passes (proving the cache deletion forces recompilation — without it passes 2–4 report "Nothing to compile"), config restore+sed ran 4×, 0 build/test failures, exit 0.
- `cargo check` / `clippy` / `fmt` clean on `solx-dev`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4477,7 +4477,6 @@ dependencies = [
  "clap",
  "colored",
  "fs_extra",
- "glob",
  "hex",
  "itertools 0.15.0",
  "path-slash",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4800,7 +4800,6 @@ dependencies = [
  "clap",
  "colored",
  "fs_extra",
- "glob",
  "hex",
  "itertools 0.15.0",
  "path-slash",
--- a/solx-dev/Cargo.toml
+++ b/solx-dev/Cargo.toml
@@ -21,7 +21,6 @@ serde.workspace = true
 sha2.workspace = true
 serde_json.workspace = true
 regex.workspace = true
-glob.workspace = true
 hex.workspace = true
 itertools.workspace = true
 toml = "1.0"
--- a/solx-dev/src/test/foundry/mod.rs
+++ b/solx-dev/src/test/foundry/mod.rs
@@ -89,30 +89,87 @@ pub fn test(
         let mut project_directory = crate::utils::absolute_path(projects_directory.as_path())?;
         project_directory.push(project_name.as_str());
 
+        crate::utils::remove(project_directory.as_path(), project_name.as_str())?;
+
+        let project_directory_str = project_directory.to_string_lossy();
+        crate::utils::clone_repository(
+            project.url.as_str(),
+            &project_directory_str,
+            project.commit.as_deref(),
+            &format!(
+                "{} Foundry project {}",
+                solx_utils::cargo_status_ok("Cloning"),
+                project_name.bright_white().bold()
+            ),
+        )?;
+
+        if project.requires_yarn {
+            crate::utils::exists("npm")?;
+
+            let build_system = "yarn";
+            let yarn_version = config.build_systems.get(build_system).ok_or_else(|| {
+                anyhow::anyhow!("Foundry test configuration missing `build_systems.{build_system}`")
+            })?;
+            let npm_spec = format!("{build_system}@{yarn_version}");
+            let mut npm_install_yarn = Command::new("npm");
+            npm_install_yarn.current_dir(project_directory.as_path());
+            npm_install_yarn.arg("install");
+            npm_install_yarn.args(["--loglevel", "error"]);
+            npm_install_yarn.arg("--force");
+            npm_install_yarn.arg("--yes");
+            npm_install_yarn.arg("--global");
+            npm_install_yarn.arg(&npm_spec);
+            crate::utils::command_with_retries(
+                &mut npm_install_yarn,
+                format!(
+                    "{} build system {} for Foundry project {project_name}",
+                    solx_utils::cargo_status_ok("Installing"),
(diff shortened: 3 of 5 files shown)


hedgar2017:
Need to be careful with these!

At first, I used to clone everything just once, but then realized that those projects were finding ways to keep artifacts where I could not find them and it was caching stuff ruining benchmark results in very subtle ways.

I'm sure it can be done, but I didn't risk back then as the network was probably faster so nuking the whole state every time was safer.

nebasuke:
> Need to be careful with these!
> 
> At first, I used to clone everything just once, but then realized that those projects were finding ways to keep artifacts where I could not find them and it was caching stuff ruining benchmark results in very subtle ways.
> 
> I'm sure it can be done, but I didn't risk back then as the network was probably faster so nuking the whole state every time was safer.

Fully agreed. 

I think I already discovered Hardhat doesn't know the difference between solx/solc when they're both using the same Solidity version, and there's other subtleties. This one will likely be draft for a little while until I either have a clear story, or a very obvious optimisation that doesn't implicitly break stuff.

hedgar2017:

Force-pushed `integration-clone-once`: `6f725270` → `142aaa83`. No conflicts, and `git range-diff` reports both commits identical (`=`) — the patches came through unchanged.

PR #587, opened 2026-07-22
refactor(codegen): de-globalize the LLVM stack-too-deep and option channels

abinavpp:
https://github.com/NomicFoundation/solx-llvm/pull/123


abinavpp:
This PR carries a benchmark comparing multithreading against the multiprocess worker pool (`tests/benchmark/pool-modes/`, run by the `ci:mode-bench` label). Results below are from https://github.com/NomicFoundation/solx/actions/runs/31893500518 and were reproduced across two runs. It is a pure mode comparison: both sides run identical code, the only difference is whether a contract compiles on a thread or in a worker process.

The input is generated: many tiny contracts, a few big ones (~3s of codegen each), each in a clean variant and a variant that triggers stack-too-deep, plus a mix of all four.

ratio = multiprocess/multithreading (>1 = threads faster):

| corpus | Linux x86 | Linux arm64 | macOS arm64 | macOS x86 |
|---|---|---|---|---|
| tiny-clean (200) | 1.03 | 0.99 | 0.96 | 0.98 |
| tiny-overflow (100) | 1.10 | 1.06 | 1.20 | **2.14** |
| big-clean (12) | 1.02 | 0.98 | 0.94 | 1.04 |
| big-overflow (6) | 1.02 | 1.00 | 0.98 | 0.97 |
| mixed (159) | 1.02 | 0.99 | 1.00 | 1.07 |

Parity everywhere except stack-too-deep on cheap contracts: multiprocess kills and respawns the worker on every overflow, and that cost scales with how expensive process spawning is on the platform (2.1x on macOS x86).

Windows: multithreading is broken there (LLVM's PrettyStackTrace gets corrupted under concurrent compilation - needs its own investigation), so Windows keeps the multiprocess default; all tests are green with that arrangement.

Not measured yet: real contracts (limited by current slang coverage), and the optimizations only multithreading can unlock later, like compiling each contract fully in memory instead of shipping it to the job as text (~6-12% of a big contract's compile time today) and in-memory stack-too-deep retries. There is no urgency to force the switch; the question is worth revisiting with real contracts and a working Windows build once slang matures.

hedgar2017:
Left this on `main-solc` after the trunk swap (`main-slang` is now `main`). Unlike the other two, nothing here is superseded, so it needs a real rebase rather than a shortcut. It stops on `44fd30b4` with conflicts in `solx-codegen-evm/src/lib.rs` and `solx-core/src/process/child.rs`. One heads-up: `51f0ef2e` looks like it's already done on the trunk, so it will probably drop out of the rebase.

PR #605, opened 2026-07-24
fix(debug-info): make standard JSON output reproducible

### LINKED ISSUE #594: Debug info embeds the build directory: identical standard JSON inputs yield different debugInfo across working dirs
## Summary

In standard JSON mode, the emitted DWARF (`evm.*.debugInfo`) absolutizes source unit names against the compiler's working directory: `DW_AT_comp_dir` is set to the cwd, and `DW_AT_decl_file` / `DW_AT_call_file` become `<cwd>/<source unit name>`. As a result, the same standard JSON input compiled from two different directories produces byte-identical bytecode (metadata hash included) but different `debugInfo` — the debug info is not reproducible across build locations.

## Reproduction (solx 0.1.6, linux-amd64; also reproduced on 0.1.4)

```sh
mkdir a b
cat > input.json <<'EOF'
{
 "language": "Solidity",
 "sources": {
  "C.sol": { "content": "// SPDX-License-Identifier: MIT\npragma solidity ^0.8.34;\ncontract C {\n    function f() public pure returns (uint256) {\n        return 1;\n    }\n}\n" }
 },
 "settings": {
  "outputSelection": { "*": { "*": ["evm.deployedBytecode.object", "evm.deployedBytecode.debugInfo"] } }
 }
}
EOF
(cd a && solx --standard-json < ../input.json > out.json)
(cd b && solx --standard-json < ../input.json > out.json)
```

`evm.deployedBytecode.object` is identical in both outputs; `evm.deployedBytecode.debugInfo` is not. Decoding the hex to ELF and diffing `llvm-dwarfdump --debug-info` shows the only difference:

```
<               DW_AT_comp_dir	(".../a")
---
>               DW_AT_comp_dir	(".../b")
<                 DW_AT_decl_file	(".../a/C.sol")
---
>                 DW_AT_decl_file	(".../b/C.sol")
```

`--debug-line` file tables shift identically; there are no non-path differences.

## Why it matters

- **Artifact reproducibility**: two machines (or two checkouts) building the same input cannot produce the same artifact once `debugInfo` is part of it — this breaks content-addressed build caching and byte-for-byte verification of debug-enabled builds.
- **Leaks local paths**: absolute build paths (user names, CI workspace layouts) end up embedded in artifacts that may be shared or published.
- **Downstream fixtures churn**: EDR commits solx standard JSON outputs as test fixtures and regenerates them per solx release; every regeneration from a different directory rewrites every `debugInfo` blob even when nothing semantic changed. (Found while verifying that a committed fixture could be reproduced: bytecode round-tripped byte-identically, `debugInfo` differed only by these paths.)

## Expected behavior

In standard JSON mode, source unit names (the `sources` keys) are the canonical identifiers — the file table should carry them verbatim, with `DW_AT_comp_dir` empty or a fixed constant. Consumers already treat the names as opaque identifiers (EDR's DWARF decoder matches them back to the input's source names by suffix precisely because of this absolutization). For reference, no other artifact solc or solx emits in standard JSON mode (source maps, metadata) embeds the build directory.

hedgar2017:
Makes standard JSON debug info reproducible across working directories:

- Emits source unit names with an empty DWARF compilation directory instead of the compiler cwd.
- Covers EVMLA and via-IR with an end-to-end regression.

Closes #594.

### DIFF
--- a/solx-codegen-evm/src/context/debug_info.rs
+++ b/solx-codegen-evm/src/context/debug_info.rs
@@ -4,7 +4,6 @@
 
 use std::collections::BTreeMap;
 use std::collections::HashMap;
-use std::path::PathBuf;
 
 use inkwell::debug_info::AsDIScope;
 use inkwell::debug_info::DIFlagsConstants;
@@ -19,9 +18,6 @@ use crate::context::IContext;
 pub struct DebugInfo<'ctx> {
     /// The debug info builder.
     builder: inkwell::debug_info::DebugInfoBuilder<'ctx>,
-    /// The main compile unit.
-    /// Directory of the current translation unit. Stored to prevent memory freeing.
-    _directory: PathBuf,
     /// The files used for the current translation unit.
     files: HashMap<usize, inkwell::debug_info::DIFile<'ctx>>,
 }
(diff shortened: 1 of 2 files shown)


nebasuke (approved):
Verified end-to-end in EDR, and it indeed fixes the reproducibility issue. Thanks for tackling this!

Issue #598, opened 2026-07-23
`memory-safe` assembly restoring `0x40` to `0x80` hits the spill area


jubnzv:
solx reserves its stack-spill region at `[0x80, 0x80+N)`, so an `assembly ("memory-safe")` block that restores the free-memory pointer to the documented initial value `0x80` redirects the allocator into that region. solc certifies the block, so no `UNSAFEASM` marker is emitted and nothing warns.

### MRE
`mre.sol` -- ordinary Solidity plus a three-token memory-safe block. `deep()` exists only to force a spill; it never touches `arr`:

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract Test {
    uint256 public s;

    function deep(uint256 a, uint256 b) internal returns (uint256) {
        unchecked {
            uint256 v0 = a * 1 + b;   uint256 v1 = a * 2 + b;
            uint256 v2 = a * 3 + b;   uint256 v3 = a * 4 + b;
            uint256 v4 = a * 5 + b;   uint256 v5 = a * 6 + b;
            uint256 v6 = a * 7 + b;   uint256 v7 = a * 8 + b;
            uint256 v8 = a * 9 + b;   uint256 v9 = a * 10 + b;
            uint256 v10 = a * 11 + b; uint256 v11 = a * 12 + b;
            uint256 v12 = a * 13 + b; uint256 v13 = a * 14 + b;
            uint256 v14 = a * 15 + b; uint256 v15 = a * 16 + b;
            v0 ^= uint256(keccak256(abi.encode(v7, v13)));
            v1 ^= uint256(keccak256(abi.encode(v8, v14)));
            v2 ^= uint256(keccak256(abi.encode(v9, v15)));
            v3 ^= uint256(keccak256(abi.encode(v10, v0)));
            v4 ^= uint256(keccak256(abi.encode(v11, v1)));
            v5 ^= uint256(keccak256(abi.encode(v12, v2)));
            v6 ^= uint256(keccak256(abi.encode(v13, v3)));
            v7 ^= uint256(keccak256(abi.encode(v14, v4)));
            v8 ^= uint256(keccak256(abi.encode(v15, v5)));
            v9 ^= uint256(keccak256(abi.encode(v0, v6)));
            v10 ^= uint256(keccak256(abi.encode(v1, v7)));
            v11 ^= uint256(keccak256(abi.encode(v2, v8)));
            v12 ^= uint256(keccak256(abi.encode(v3, v9)));
[...]

Issue #596, opened 2026-07-23
`mcopy(x,x,n)` is deleted, so `msize()` and expansion gas are wrong


jubnzv:
A self-copy is erased as a no-op, but `MCOPY` on the EVM performs and charges for memory expansion even when source equals destination, and that expansion is observable through `MSIZE` and gas.

### MRE

`tests/solidity/simple/mre.sol`:

```solidity
//! { "cases": [
//!  {"name":"selfcopy_msize","inputs":[{"method":"f","calldata":[]}],"expected":["0x0000000000000000000000000000000000000000000000000000000000002020"]}
//! ] }
// SPDX-License-Identifier: MIT
pragma solidity >=0.8.28;
contract Test {
    function f() public pure returns (uint256 m) {
        assembly {
            mcopy(0x2000, 0x2000, 32)
            m := msize()
        }
    }
}
```

Same thing straight from Yul (`solx --yul`), no Solidity involved:

```yul
object "T" {
  code {
    mcopy(0x2000, 0x2000, 32)
    let m := msize()
    mstore(0, m)
    return(0, 32)
  }
}
```

```
./target/release/solx-tester --solidity-compiler ./target/release/solx --path tests/solidity/simple/mre.sol
```

Execution results — all 10 calls fail (2 pipelines x 5 optimization modes):

```
expected: 0x2020   (MCOPY touches [0x2000,0x2020), so msize = 0x2020)
got:      0x0060   (msize unchanged; the copy is gone)
```

```bash
./target/release/solx --asm -O1 mre.sol | grep -c MCOPY   # 0
./target/release/solx --asm -O3 mre.sol | grep -c MCOPY   # 0
```

Reference: `solc 0.8.35 --strict-assembly --evm-version cancun --optimize` keeps the `mcopy`.

solx `ea317f4f32f8d1647a666a1ff212c138c24ec3ec` · solx-llvm `c6876bba8dc44ddcdf7757475e9b6bbef5410930`

akiramenai:
I'm not sure this should be classified as a bug. As you mentioned, gas is also an observable effect, yet compilers routinely optimize for gas, and writing programs that depend on precise gas arithmetic is generally considered bad practice. Likewise, while `msize` is observable, from a purely pragmatic point of view I'd eliminate this `MCOPY` anyway.

By the way, `solc` simply rejects the program with `--via-ir --optimize`:
https://godbolt.org/z/MfGq87os1

[...]

PR #606, opened 2026-07-25
docs: clarify optimizer and assembly semantics

hedgar2017:
Documents the optimizer and assembly compatibility boundary discussed in #596 and #598:

- Yul and inline assembly are compiler input rather than verbatim EVM bytecode; exact gas, memory expansion, and `msize()` are not stable functional dependencies.
- The `memory-safe` annotation is an author guarantee; the free memory pointer must be read at block entry, and bypassing the spill-safety check can corrupt memory.
- The CLI, Standard JSON, and Yul builtin references now point to the canonical limitations sections.


nebasuke (approved):
Thanks, looks good!

PR #612, opened 2026-07-27
ci(slang-tests): skip doc-only changes via paths filter

nebasuke:
# ci(slang-tests): skip doc-only changes via paths filter

Mirrors `test.yaml`'s `changes` job in `slang-tests.yaml`: PRs and pushes that only touch `docs/`, `*.md`, `*.mdx`, or `LICENSE*` no longer run the 5-platform toolchain matrix.

Wiring:

- New `changes` job (same `dorny/paths-filter` pin and filter list as `test.yaml`) runs in parallel with `label-check`.
- `cooldown-check` gains `needs: changes` + `if: needs.changes.outputs.code == 'true'`; the skip propagates to `slang-tests` through the existing `needs` chain.
- `pr-checks` (alls-green) gets a second `allowed-skips` branch so the required check passes when the matrix was skipped for a doc-only change, while a failed `changes` job still fails it.

Behavior is unchanged for the existing `ci:slang` label gating and `push`/`merge_group` events — this only adds the doc-only skip, as groundwork for eventually enabling Slang tests by default.

`actionlint` clean.

### DIFF
--- a/.github/workflows/slang-tests.yaml
+++ b/.github/workflows/slang-tests.yaml
@@ -19,44 +19,40 @@ concurrency:
 jobs:
 
   label-check:
+    # Checks the PR's current label set, not the triggering action: keying on
+    # github.event.label.name would skip this run (and cancel a real one in
+    # flight) when an unrelated label is added to a PR carrying ci:slang.
     if: >-
       github.event_name == 'push'
-      || ((github.event.action == 'labeled' && github.event.label.name == 'ci:slang'
-      || github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'ci:slang'))
+      || (contains(github.event.pull_request.labels.*.name, 'ci:slang')
       && !github.event.pull_request.head.repo.fork)
     runs-on: ubuntu-24.04
     steps:
       - run: 'true'
 
-  # Detect whether code (non-doc) files changed; mirrors the filter in test.yaml.
   changes:
+    needs: label-check
     runs-on: ubuntu-24.04
     permissions:
       contents: read
-      pull-requests: read
     outputs:
       code: ${{ steps.filter.outputs.code }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           persist-credentials: false
-      - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
+          # Full history for the filter's git-diff mode.
+          fetch-depth: 0
+      - uses: ./.github/actions/code-changes
         id: filter
-        with:
-          predicate-quantifier: 'every'
-          filters: |
-            code:
-              - '**'
-              - '!docs/**'
-              - '!**/*.md'
-              - '!**/*.mdx'
-              - '!**/LICENSE*'
 
   cooldown-check:
     name: Cargo cooldown check
     runs-on: ubuntu-24.04
     needs: [label-check, changes]
-    if: needs.changes.outputs.code == 'true'
+    # A custom if: overrides the implicit skip cascade from needs
+    # (actions/runner#491), so the label gate must be restated here.
+    if: needs.label-check.result == 'success' && needs.changes.outputs.code == 'true'
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
@@ -206,6 +202,6 @@ jobs:
         with:
           jobs: ${{ toJSON(needs) }}
           allowed-skips: >-
-            ${{ needs.label-check.result == 'skipped' && 'label-check, cooldown-check, slang-tests'
+            ${{ needs.label-check.result == 'skipped' && 'label-check, changes, cooldown-check, slang-tests'
             || needs.changes.outputs.code != 'true' && 'cooldown-check, slang-tests'
             || '' }}
--- a/.github/workflows/slang-tests.yaml
+++ b/.github/workflows/slang-tests.yaml
@@ -28,10 +28,35 @@ jobs:
     steps:
       - run: 'true'
 
+  # Detect whether code (non-doc) files changed; mirrors the filter in test.yaml.
+  changes:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      code: ${{ steps.filter.outputs.code }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
+        id: filter
+        with:
+          predicate-quantifier: 'every'
+          filters: |
+            code:
+              - '**'
+              - '!docs/**'
+              - '!**/*.md'
+              - '!**/*.mdx'
+              - '!**/LICENSE*'
+
   cooldown-check:
     name: Cargo cooldown check
     runs-on: ubuntu-24.04
-    needs: label-check
+    needs: [label-check, changes]
+    if: needs.changes.outputs.code == 'true'
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
@@ -172,6 +197,7 @@ jobs:
     if: always()
     needs:
       - label-check
+      - changes
       - cooldown-check
       - slang-tests
     steps:
@@ -180,4 +206,6 @@ jobs:
         with:
           jobs: ${{ toJSON(needs) }}
           allowed-skips: >-
-            ${{ needs.label-check.result == 'skipped' && 'label-check, cooldown-check, slang-tests' || '' }}
+            ${{ needs.label-check.result == 'skipped' && 'label-check, cooldown-check, slang-tests'
+            || needs.changes.outputs.code != 'true' && 'cooldown-check, slang-tests'
+            || '' }}
--- a/.github/actions/code-changes/action.yml
+++ b/.github/actions/code-changes/action.yml
@@ -1,0 +1,33 @@
+name: Code changes filter
+description: >-
+  Detects whether any changed file is code rather than documentation.
+  The repository must already be checked out with full history
+  (fetch-depth 0) before calling this action.
+
+outputs:
+  code:
+    description: Whether any non-documentation file changed (true/false)
+    value: ${{ steps.filter.outputs.code }}
+
+runs:
+  using: composite
+  steps:
+    # token '' forces git-diff mode: the PR files API caps at 3,000 files, and
+    # a truncated list can fail open (matrix skipped despite a code change).
+    # 'every': a file counts as code only if no negation pattern excludes it.
+    # Not every .md is documentation — solx-benchmark-converter compiles
+    # templates/summary.md into the binary and asserts golden .md fixtures.
+    - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
+      id: filter
+      with:
+        token: ''
+        predicate-quantifier: 'every'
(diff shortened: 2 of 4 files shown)


.github/workflows/slang-tests.yaml
```diff
@@ -28,10 +28,35 @@ jobs:
     steps:
       - run: 'true'
 
+  # Detect whether code (non-doc) files changed; mirrors the filter in test.yaml.
```
hedgar2017: The first clause restates the job name and the filter three lines below it. The second asserts an invariant nothing enforces — see the comment on line 32. Delete the line.

.github/workflows/slang-tests.yaml
```diff
@@ -28,10 +28,35 @@ jobs:
...
+  changes:
```
hedgar2017, line 33: Two things about this job.

It has no `needs` and no `if:`, so it runs on every event the workflow accepts. On `merge_group` and on unlabeled PRs nothing here ran before; now we pay a checkout and a diff every time, and if the job errors the required check fails outright — `allowed-skips` never excuses a failure. Gate it:

```yaml
changes:
  needs: label-check
```

and widen the first `allowed-skips` branch to `'label-check, changes, cooldown-check, slang-tests'`.

Lines 32-53 are also a verbatim copy of `test.yaml:24-46`, held together by a one-way prose comment that `test.yaml` does not reciprocate. Both filter defects below have to be fixed in two places today, which is the argument. Extract the shared `code:` filter into a composite action under `.github/actions/` — the convention `cooldown-check/action.yml` already uses — and reference it from both workflows. `test.yaml`'s `devcontainer` filter stays inline; it is genuinely local.

.github/workflows/slang-tests.yaml
```diff
@@ -28,10 +28,35 @@ jobs:
...
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      code: ${{ steps.filter.outputs.code }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
```
hedgar2017: On `pull_request` this resolves changed files through GitHub's "List pull requests files" API, which returns at most 3,000 files regardless of pagination — v4.0.2 does paginate, but it cannot fetch what the API will not return. A PR above that threshold yields a truncated list, `code` can come back `false` on a real code change, and the required check passes with the matrix skipped.

Force the git-diff path, here and in `test.yaml`: add `token: ''` to this step and `fetch-depth: 0` to the checkout above. With `token: ''` the `pull-requests: read` permission on line 36 is no longer needed and should come off.

.github/workflows/slang-tests.yaml
```diff
@@ -28,10 +28,35 @@ jobs:
...
+        id: filter
+        with:
+          predicate-quantifier: 'every'
+          filters: |
+            code:
+              - '**'
+              - '!docs/**'
+              - '!**/*.md'
```
hedgar2017: Copied from `test.yaml:44,46`, where they are already wrong: not every `.md` in this repo is documentation.

`solx-benchmark-converter/templates/summary.md` is compiled into the binary via `#[template(path = "summary.md", escape = "none")]` (`solx-benchmark-converter/src/output/summary/summary_template/mod.rs:39`), and `src/tests/fixtures/*.md` are golden fixtures asserted against rendered output. `solx-benchmark-converter` is in `default-members`. A PR changing only the template and its fixtures evaluates `code == 'false'` and skips the matrix while showing green — and under `predicate-quantifier: 'every'` no positive pattern can re-include them, because `every` requires all patterns to match.

`'!**/LICENSE*'` on line 53 matches any basename starting with `LICENSE` anywhere, including `tests/solidity/complex/defi/UniswapV2/LICENSE.txt`.

Narrow to the paths that are actually documentation, here and in `test.yaml`:

```yaml
code:
  - '**'
  - '!docs/**'
  - '!*.md'
  - '!**/README.md'
  - '!.github/**/*.md'
  - '!**/LICENSE-APACHE'
  - '!**/LICENSE-MIT'
```

.github/workflows/slang-tests.yaml
```diff
@@ -28,10 +28,35 @@ jobs:
...
+              - '!**/*.mdx'
+              - '!**/LICENSE*'
+
   cooldown-check:
     name: Cargo cooldown check
     runs-on: ubuntu-24.04
-    needs: label-check
+    needs: [label-check, changes]
+    if: needs.changes.outputs.code == 'true'
```
hedgar2017: A job-level `if:` without a status-check function suppresses the implicit `success()` over `needs`. Before this PR `cooldown-check` had `needs: label-check` and no `if:`, so a skipped `label-check` skipped it. Now the job runs on the written expression alone: every PR with a code change — unlabeled, fork — and every `merge_group`, where `label-check` is always skipped. That contradicts "behavior is unchanged for the existing `ci:slang` label gating", and it means a second `cooldown-check` per unlabeled PR on top of `test.yaml`'s.

`slang-tests` is unaffected — it keeps a plain `needs`, so the skip still cascades and nothing gets past the gate.

Make the dependency explicit, since the `if:` no longer implies it:

```yaml
if: needs.label-check.result == 'success' && needs.changes.outputs.code == 'true'
```

Separately, `label-check` also runs on `push`, so this gate applies there too: a doc-only merge to `main` now skips the whole Slang matrix on trunk. `test.yaml` already behaves that way, so it may be intended — but it is a real reduction in post-merge coverage and the PR description claims the opposite. Either fix the description or scope the gate:

```yaml
if: >-
  github.event_name != 'pull_request'
  || needs.changes.outputs.code == 'true'
```

.github/workflows/slang-tests.yaml
```diff
@@ -180,4 +206,6 @@ jobs:
         with:
           jobs: ${{ toJSON(needs) }}
           allowed-skips: >-
-            ${{ needs.label-check.result == 'skipped' && 'label-check, cooldown-check, slang-tests' || '' }}
+            ${{ needs.label-check.result == 'skipped' && 'label-check, cooldown-check, slang-tests'
```
hedgar2017: This branch widens what a skipped `label-check` excuses, which makes an existing hole worse.

`label-check`'s condition (lines 21-26) keys the `labeled` case on `github.event.label.name`, so adding **any other** label to a PR that already carries `ci:slang` skips `label-check` for that run — the `action == 'labeled'` branch short-circuits the "already has the label" branch. `cooldown-check` and `slang-tests` then skip via `needs`, this branch excuses all three, and `PR Checks (Slang)` reports green with nothing tested. `merge_group` hits the same path: no `pull_request` object in that payload, so both disjuncts are false — and `merge_group` is the event GitHub added to validate the merge-group head. `cancel-in-progress` on line 17 compounds it: the unrelated-label event cancels an in-flight real run and then reports green.

Check the PR's current label set rather than the triggering action, and let non-PR events through:

```yaml
if: >-
  github.event_name != 'pull_request'
  || (contains(github.event.pull_request.labels.*.name, 'ci:slang')
  && !github.event.pull_request.head.repo.fork)
```

While this is open: `sanitizer.yaml` has the identical `label-check` → `cooldown-check` → matrix → `alls-green` shape with a single-branch `allowed-skips` and no doc-skip filter, so a doc-only `ci:sanitizer` PR still pays a full ASan build. Once the filter is a composite action, wiring it there is three lines. `integration-tests.yaml` and `compile-benchmark.yaml` have no `alls-green` job, so they need nothing.

hedgar2017 (changes requested):
Inline comments below.

.github/workflows/slang-tests.yaml
nebasuke: Removed (the whole inline filter moved into the shared action, so the line is gone either way).

.github/workflows/slang-tests.yaml
nebasuke: Done in 7ae7f8b3:

- `changes` now has `needs: label-check`, so unlabeled PRs and merge groups pay nothing and a failed filter can't fail `PR Checks (Slang)` on a PR that never opted into Slang CI. First `allowed-skips` branch widened to `'label-check, changes, cooldown-check, slang-tests'`.
- Filter extracted to `.github/actions/code-changes` and referenced from both `slang-tests.yaml` and `test.yaml`; the one-way "mirrors test.yaml" comment is gone with it. `test.yaml`'s `devcontainer` filter stays inline as you suggested.

.github/workflows/slang-tests.yaml
nebasuke: Done in 7ae7f8b3: `token: ''` on the filter (inside the shared action) and `fetch-depth: 0` on the checkout, in both workflows; `pull-requests: read` dropped from both. I also put `token: ''` on `test.yaml`'s devcontainer filter so that job has no API dependence at all — it shares the same full-history checkout.

Checked the pinned v4.0.2 source while at it: tokenless mode git-diffs against `pull_request.base.sha` on PRs, `event.before` on push, and `merge_group.base_sha..head_sha` on merge groups, so all three triggers stay on the git path.

.github/workflows/slang-tests.yaml
nebasuke: Done in 7ae7f8b3 — your list verbatim in the shared action, applied to both workflows. Confirmed the failure you describe: `summary.md` is compiled in via the askama `#[template]` and the six fixture `.md`s are golden files, all in a `default-members` crate.

One deliberate omission: I dropped `!**/*.mdx` entirely — there are no `.mdx` files in the tree, and any future ones under `docs/` are covered by `!docs/**`.

.github/workflows/slang-tests.yaml
nebasuke: Done in 7ae7f8b3: `if: needs.label-check.result == 'success' && needs.changes.outputs.code == 'true'`, with a comment pointing at actions/runner#491 since the pattern is easy to regress.

For push: kept the trunk doc-skip (it matches what `test.yaml` already does on `main`) and fixed the PR description instead, which was wrong as you said.

.github/workflows/slang-tests.yaml
nebasuke: The hole is fixed in 7ae7f8b3: `label-check` now keys on the PR's current label set, so an unrelated label added to a `ci:slang` PR runs the suite instead of cancelling it and reporting green.

I deviated from your suggested condition in one place: `github.event_name == 'push' || ...` rather than `!= 'pull_request'`, which keeps `merge_group` skipped as today. Letting merge groups through would run the 5-platform Slang matrix on every code merge-queue entry while the suite is still opt-in — that felt like a separate cost decision from this fix. If you want the matrix validating the queue head now, I'll flip it (it's the natural end state once Slang tests are on by default).

`sanitizer.yaml` is done in this PR (429f5699): same wiring as `slang-tests.yaml` — `changes` job on the shared filter gating `cooldown-check`, widened `allowed-skips` — plus the same `label-check` fix, since it had the identical unrelated-label hole. `merge_group` unchanged there too.

.github/workflows/slang-tests.yaml
```diff
@@ -28,10 +28,35 @@ jobs:
...
+    outputs:
+      code: ${{ steps.filter.outputs.code }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
+        id: filter
+        with:
+          predicate-quantifier: 'every'
```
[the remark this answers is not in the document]
nebasuke: The quantifier applies per file, not across the change set: each changed file matches the `code` filter only if it satisfies every pattern (`'**'` plus all negations), and the output is `true` when *any* file matches. A mixed code+docs PR still yields `code == true` via the code file — only a PR where every file is documentation skips. This is the standard dorny negation pattern, and `test.yaml` has been running the identical filter on every PR without skipping mixed changes.

hedgar2017 (approved):
Thanks! Let me think how to fix the CI tomorrow.
For now a simple bump of solx-llvm should help.

Issue #560, opened 2026-07-17
Function symbol mangling in the Slang frontend collides under inheritance and function-pointer overloads


abinavpp:
Symbols come from slang's `compute_internal_signature()`, which isn't unique. Two collision classes:

- Function pointers, today: slang renders any function type as the bare string `function`, so overloads differing only in a function-typed parameter - `apply(function(uint256) returns (uint256))` vs `apply(function(bool) returns (bool))` - both mangle to `apply(function)`.
- Inheritance, once it lands: the name isn't qualified with the defining contract, so same-signature functions coexisting in one contract's code (`super.f()` next to its override, shadowed private base functions) collide.

The node id could make symbols unique, like the C++ frontend's `getMangledName` (`name_<ast id>`); provenance is source locations' job: #559.

hedgar2017:
A third class, reachable as soon as free functions are emitted — no inheritance and no function-typed parameters needed: a free function and a contract method with the same signature both lower into one `sol.contract`.

`semanticTests/freeFunctions/free_namesake_contract_function.sol`:

```solidity
function f() pure returns (uint) { return 1337; }
contract C {
  function f() public pure returns (uint) { return f(); }
}
```

Both mangle to `f()`, and the module is rejected:

```
error: redefinition of symbol named 'f()'
```

Hit while emitting every free function into the contract body, since MLIR has no file scope for them to land in.

Worth recording alongside it: the node-id scheme above fixes the symbol clash, but solc does not emit this free function at all. It resolves the inner `f()` to the contract's own method — the free one is shadowed — and emits a single `@f_17` calling itself. So symbol uniqueness and emitting only reachable free functions are separate requirements; this test needs both.

PR #614, opened 2026-07-28
feat(slang): complete enums and user-defined value types

hedgar2017:
Completes enums and user-defined value types on the Slang frontend:

- An enum member `E.Variant`, its qualified form `C.E.Variant`, and `type(E).min` / `type(E).max` materialize as the member's ordinal: a `sol.constant` at the EVM word, cast by `sol.enum_cast` to `!sol.enum<N>`; the ordinal is a member's position among the enum's declarations, and the maximum is read back off the dialect type that already bounds it.
- `E(v)` and `uint8(e)` lower to `sol.enum_cast` in either direction, a wider target chaining `sol.cast` behind it; the range check on an integer narrowed into an enum belongs to the lowering pass, not the frontend.
- An enum comparison emits `sol.cmp` at `!sol.enum<N>` with no widening in front of it, the dialect admitting an enum wherever it admits an integer.
- An enum-typed state variable, mapping key, struct field, or array element carries `!sol.enum<N>` through `sol.state_var`, `sol.mapping`, `sol.gep`, and `sol.load`; an uninitialized local, a named return, and `delete` all default-initialize to ordinal zero through the same constant-and-cast pair solc emits.
- `T.wrap(x)` and `T.unwrap(v)` emit nothing, a user-defined value type lowering to the type it wraps, so both are the identity case of an explicit conversion.
- A user-defined operator — the fourteen binary forms, `~`, and unary `-` — lowers to `sol.call` of the function a `using {f as op} for T global;` directive binds it to, which the binder resolves and the AST reads back; being a call, its operands evaluate left-first rather than in a built-in operator's right-first order.
- The operator-bound free functions emit into the contract body alongside the contract's own, MLIR having no file scope for them to land in.

### DIFF
--- a/solx-mlir/tests/lit/type_casts.sol
+++ b/solx-mlir/tests/lit/type_casts.sol
@@ -37,7 +37,19 @@
 // CHECK: sol.func @{{.*array_to_memory.*}}
 // CHECK:   sol.data_loc_cast %{{.*}} : !sol.array<? x ui256, CallData>, !sol.array<? x ui256, Memory>
 
+// CHECK: sol.func @{{.*uint8_to_enum.*}}
+// CHECK:   sol.enum_cast %{{.*}} : ui8 to !sol.enum<2>
+
+// CHECK: sol.func @{{.*enum_to_uint8.*}}
+// CHECK:   sol.enum_cast %{{.*}} : !sol.enum<2> to ui8
+
+// CHECK: sol.func @{{.*enum_to_uint256.*}}
+// CHECK:   sol.enum_cast %{{.*}} : !sol.enum<2> to ui8
+// CHECK:   sol.cast %{{.*}} : ui8 to ui256
+
 contract C {
+    enum E { First, Second, Third }
+
     function uint8_to_uint256(uint8 x) public pure returns (uint256) {
         return uint256(x);
     }
@@ -85,5 +97,17 @@ contract C {
 
     function array_to_memory(uint256[] calldata source) external pure returns (uint256[] memory) {
         return uint256[](source);
+    }
+
+    function uint8_to_enum(uint8 x) public pure returns (E) {
+        return E(x);
+    }
+
+    function enum_to_uint8(E x) public pure returns (uint8) {
+        return uint8(x);
+    }
+
+    function enum_to_uint256(E x) public pure returns (uint256) {
+        return uint256(uint8(x));
     }
 }
--- a/solx-slang/src/source_unit.rs
+++ b/solx-slang/src/source_unit.rs
@@ -4,7 +4,12 @@
 
 use std::collections::BTreeMap;
 
+use itertools::Itertools;
+use slang_solidity_v2::ast::Definition;
+use slang_solidity_v2::ast::FunctionDefinition;
 use slang_solidity_v2::ast::SourceUnit;
+use slang_solidity_v2::ast::SourceUnitMember;
+use slang_solidity_v2::ast::UsingClause;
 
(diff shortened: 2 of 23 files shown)


solx-slang/src/source_unit.rs
```diff
@@ -44,4 +50,29 @@ impl<'context> SourceUnitScope<'context> {
...
+                        UsingClause::IdentifierPath(_) => None,
+                    }
+                }
+                _ => None,
+            })
+            .flat_map(|deconstruction| deconstruction.symbols().iter().collect::<Vec<_>>())
+            .filter(|symbol| symbol.alias().is_some())
+            .filter_map(|symbol| match symbol.name().resolve_to_definition()? {
+                Definition::Function(function) => Some(function),
+                _ => None,
```
PavelKopyl, line 74: I wonder is there a chance to get a symbol collision between an operator function and a same signature contract member, as we do not mangle operator functions: 

```
using {add as +} for T global;

// File-level free function, bound as the `+` operator for T.
function add(T a, T b) pure returns (T) {
    return T.wrap(T.unwrap(a) + T.unwrap(b));
}

contract C {
    // Legal Solidity: a member function with the same name and parameter
    // types as the free function. Inside C it shadows the free `add` for
    // direct calls, but `+` still binds the free one.
    function add(T a, T b) public pure returns (T) {
        return a + b; // lowers to sol.call of the *free* add
    }
}
```
?

solx-mlir/tests/lit/type_casts.sol
```diff
@@ -86,4 +98,16 @@ contract C {
     function array_to_memory(uint256[] calldata source) external pure returns (uint256[] memory) {
         return uint256[](source);
     }
+
+    function uint8_to_enum(uint8 x) public pure returns (E) {
+        return E(x);
+    }
```
PavelKopyl, line 104: It's reasonable to add a runtime test (if such does't exist yet) to check when 'x' is out of range for the enum 'E'. This should cause a runtime Panic.  On the other hand we have such a test in era-solidity.

solx-slang/src/source_unit.rs
```diff
@@ -44,4 +50,29 @@ impl<'context> SourceUnitScope<'context> {
             Contract::new_mlir(mlir, method_identifiers),
         )]))
     }
+
+    /// The functions `unit`'s `using {f as op} for T global;` directives bind operators to, in
+    /// source order and without repeats. A binding reaches the whole compilation unit, so a
+    /// function bound from another file is missing here.
```
PavelKopyl, line 56: Let's add a TODO here: we don't yet support cross-file binding of operator functions. Since `global` is specifically what makes a `using ... for` attachment visible to importing files, that's behaviour the keyword's semantics require, and we'll need to handle it.

solx-mlir/tests/lit/type_casts.sol
hedgar2017: If solx-solidity already has one, I don't see any reason to add it in this repo.

solx-slang/src/source_unit.rs
hedgar2017: Yes, tracked in https://github.com/NomicFoundation/solx/issues/560 and fixed in #621 🚀

solx-slang/src/source_unit.rs
hedgar2017: Good idea, adding!

PavelKopyl (approved):
LGTM, thanks!

PR #617, opened 2026-07-29
ci: switch main-slang to the slang pipeline

hedgar2017:
- Pin `solx-solidity` via `.gitmodules` (`branch = main`) and the recorded pointer (`ebeac7c2` → `e8565c9d`, its main tip), replacing the slang-tests step that synced it to main unconditionally at CI time.
- Run slang tests on every push and PR — no `ci:slang` label needed on this branch.
- `build-rust` builds `solx` with the `build-slang` alias (MLIR-enabled LLVM, no solc); pre-releases publish without the `solx-release` approval gate, which now applies to tag releases only.
- Compile benchmark compares the PR against a `main-slang` baseline and the latest production release binary.
- Coverage runs only for PRs to `main`: its solc-pipeline test steps cannot use the slang binary this branch builds.


.github/actions/build-rust/action.yml
```diff
@@ -82,11 +82,11 @@ runs:
         if [[ "${RUNNER_OS}" == "Windows" ]] || [[ "${RUNNER_OS}" == "Linux" ]]; then
           export RUSTFLAGS="${RUSTFLAGS} -C link-arg=-fuse-ld=lld"
         fi
-        ${SFW_PREFIX:-} cargo ${BUILD_STD_LIB} build --frozen ${RELEASE} ${{ steps.build-target.outputs.target }}
+        ${SFW_PREFIX:-} cargo ${BUILD_STD_LIB} build-slang --frozen ${RELEASE} ${{ steps.build-target.outputs.target }}
```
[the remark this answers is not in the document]
hedgar2017: Addressed in 7b96a053. The action no longer accepts exec_name; it now builds and reports the solx binary explicitly, and both callers were updated.

.github/actions/build-rust/action.yml
```diff
@@ -98,8 +98,8 @@ runs:
       run: |
         mkdir -p ./releases/${{ inputs.release-suffix }}
         [ "$RUNNER_OS" = "Windows" ] && WIN_SUFFIX=".exe"
-        strip "./target/${{ inputs.target }}/${{ inputs.build-type }}/${{ inputs.exec_name }}${WIN_SUFFIX}"
-        mv "./target/${{ inputs.target }}/${{ inputs.build-type }}/${{ inputs.exec_name }}${WIN_SUFFIX}" \
+        strip "./target-slang/${{ inputs.target }}/${{ inputs.build-type }}/${{ inputs.exec_name }}${WIN_SUFFIX}"
+        mv "./target-slang/${{ inputs.target }}/${{ inputs.build-type }}/${{ inputs.exec_name }}${WIN_SUFFIX}" \
           "./releases/${{ inputs.release-suffix }}/${{ inputs.exec_name }}-${{ inputs.release-suffix }}${WIN_SUFFIX}"
```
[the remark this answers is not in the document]
hedgar2017: Addressed in 7b96a053. Prepare binary now consumes steps.build.outputs.binary-path through BINARY_PATH, so packaging uses the same computed path with or without an explicit target.

.github/workflows/test.yaml
```diff
@@ -136,7 +134,9 @@ jobs:
     name: Cargo cooldown check
     runs-on: ubuntu-24.04
     needs: changes
-    if: needs.changes.outputs.code == 'true'
+    if: >-
+      needs.changes.outputs.code == 'true'
+      && (github.event_name != 'pull_request' || !github.event.pull_request.head.repo.fork)
```
nebasuke: Nit: I found this a bit hard to read, maybe !(=pr && ==fork) instead? (Logically equivalent). I'm okay either way.

.github/workflows/test.yaml
```diff
@@ -224,36 +224,86 @@ jobs:
...
         uses: ./.github/actions/build-toolchain
+        with:
+          solc: 'false'
 
-      # Positioned after build-toolchain: prepare-msys switches the rustup
-      # host, which is part of the cache key, so only the post-msys position
-      # matches the keys already saved on main.
-      - name: Cache cargo artifacts
-        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+      # TODO: remove this step once the lit tests stop running against solc.
```
nebasuke, line 234: Question: what is actually needed here?

nebasuke (approved):
compile-benchmark failing as expected, but it's failing at the right stage (OpenZeppelin build), so this looks good.

.github/workflows/test.yaml
hedgar2017: It means we won't have to build solc in this workflow anymore when our LIT tests stop comparing solx's MLIR output against solc's. It's copied verbatim from `slang-tests.yaml`.

.github/workflows/test.yaml
hedgar2017: Sure, fixed!

PR #618, opened 2026-07-29
ci: fold benchmark mirror into build-{llvm,solc} composite actions

### LINKED ISSUE #448: ci: fold benchmark mirror into build-{llvm,solc} composite actions
## Context

In #375 we added a SHA-keyed mirror of LLVM and solc artifacts between the PR checkout and the `temp-solx-main` checkout in `integration-tests.yaml`, to skip the second build when the submodule SHAs and `solx-dev/src/{llvm,solc}` scripts match. The mirror predicate lives in workflow-level bash, in parallel to the cache-key logic the `build-llvm` and `build-solc` composite actions already own.

This issue tracks moving the mirror into the composite actions, as proposed by @hedgar2017 in [#375 (review)](https://github.com/NomicFoundation/solx/pull/375#pullrequestreview-4244821906).

## Proposed shape

A new `mirror-from` input on each composite action, gated on `inputs.mirror-from != ''`. When set, a single new step at the top of the action tries to mirror from the sibling checkout and short-circuits the cache-restore + build chain on success. When the mirror predicate doesn't match, falls through to the existing chain.

The workflow callsite collapses to two ordinary `uses:` calls per artifact:

```yaml
- name: Build LLVM (PR)
  uses: ./.github/actions/build-llvm
  with: { build-type: Release, enable-assertions: 'false' }

- name: Build LLVM (main)
  uses: ./.github/actions/build-llvm
  continue-on-error: true
  with:
    working-dir: temp-solx-main
    mirror-from: '.'
    build-type: Release
    enable-assertions: 'false'

- name: Build solc (PR)
  uses: ./.github/actions/build-solc

- name: Build solc (main)
  uses: ./.github/actions/build-solc
  continue-on-error: true
  with:
    working-dir: temp-solx-main/solx-solidity
    mirror-from: solx-solidity
```

The `share-{llvm,solc}` bash steps in `integration-tests.yaml` go away.

## Why this is better than the surgical fix in #375

- **Cleanup is unconditional and adjacent to the write.** `rm -rf "$DST"` sits one line above the mirror copy, inside the action — it can't drift out of the SHA-match branch the way the share-step `rm -rf` did before the #375 fixes.
- **Mirror predicate inherits the action's full cache-key dimensions for free.** Submodule SHA, `solx-dev/src/{llvm,solc}` script hash, boost version, MLIR flag, sanitizer — whatever the cache key composes from, the mirror decision composes from. The two can't drift as keys grow.
- **`cp -al` (hardlinks) instead of `ln -snf`.** Under cmake/cargo's rename-on-write conventions, hardlinks don't have the through-write surface symlinks do, and the consumers here are read-only.

## Scope

- `build-llvm/action.yml` — add `mirror-from` input + mirror step
- `build-solc/action.yml` — same
- `integration-tests.yaml` — collapse `share-{llvm,solc}` + `Build {LLVM,solc} (main)` step pairs to single `uses:` calls
- Verify the mirror predicate matches the existing cache-key composition exactly (not a subset)

hedgar2017:
Folds the SHA-keyed artifact mirror from workflow bash into the composite actions, as proposed in #375's review:

- `build-{llvm,solc}` gain a `cache-key` output and a `mirror-key` input: the main-side call receives the PR-side invocation's key and symlinks the workspace-root artifacts when its own key is equal, skipping restore and build. The predicate is the full cache key of both invocations, so it cannot drift into a subset as key dimensions grow and cannot false-match when the paired callsites diverge in build parameters.
- `integration-tests.yaml` and `compile-benchmark.yaml` (which cloned the share steps after the issue was filed) collapse to plain `uses:` calls per artifact.
- On a mirror miss, the main-side LLVM build now goes through the composite chain — artifact-cache restore (cache-warmup saves the exact Release/no-assertions key on main) and ccache — instead of the raw cold rebuild.
- The dev-script hashes widen to the modules the builds read from outside `src/llvm`/`src/solc`: `arguments/{llvm,solc}` (flag defaults), `build_type`, `ccache_variant`, and `utils` (cmake-flag composition). This re-keys every artifact cache once; cache-warmup gets dispatched right after merge so open PRs stop cold-building.

Closes #448.

### DIFF
--- a/.github/actions/build-llvm/action.yml
+++ b/.github/actions/build-llvm/action.yml
@@ -5,8 +5,11 @@ name: 'Build LLVM'
 description: 'Builds backend LLVM framework'
 outputs:
   cache-hit:
-    description: 'Whether the LLVM cache was hit'
-    value: ${{ steps.llvm-artifact-cache.outputs.cache-hit }}
+    description: 'Whether the LLVM artifacts were reused, from the cache or a sibling mirror, and no build ran'
+    value: ${{ steps.llvm-artifact-cache.outputs.cache-hit || steps.llvm-mirror.outputs.mirrored }}
+  cache-key:
+    description: 'Artifact cache key; a sibling invocation''s `mirror-key`'
+    value: ${{ steps.llvm-cache.outputs.key }}
   llvm-prefix:
     description: 'Path to the LLVM installation'
     value: 'target-llvm/target-final'
@@ -15,6 +18,13 @@ inputs:
     description: 'Working directory.'
     required: false
     default: '.'
+  mirror-key:
+    description: >-
+      Sibling invocation's `cache-key`: when equal to this invocation's own
+      key, the workspace-root artifacts are symlinked, skipping restore and
+      build. Empty disables mirroring.
+    required: false
+    default: ''
   enable-utils:
     description: "Build and install the LLVM utility binaries (FileCheck, llvm-lit, etc.). Implied by enable-tests."
     required: false
@@ -57,7 +67,7 @@ inputs:
       Disable it when the solx-llvm tree changes every run (e.g. solx-llvm's own
       regression CI) or when later steps need the build directory: an artifact hit
       skips the build entirely, so target-llvm/build-final would not exist.
-      When disabled, the `cache-hit` output is always empty.
+      When disabled, `cache-hit` reports a mirror hit or nothing at all.
     required: false
     default: 'true'
 
@@ -74,29 +84,23 @@ runs:
       run: echo "hash=$(pacman -Q | sha256sum | head -c 8)" >> "${GITHUB_OUTPUT}"
 
     - name: Compute LLVM cache key
-      if: inputs.enable-artifact-cache == 'true'
+      if: inputs.enable-artifact-cache == 'true' || inputs.mirror-key != ''
       id: llvm-cache
       shell: bash
       working-directory: ${{ inputs.working-dir }}
       run: |
         SHA=$(git -C solx-llvm rev-parse HEAD)
-        # solx-dev owns LLVM-build cmake args (shared.rs + platforms/*.rs).
-        # Hash that tree into the key so any Rust-side flag change invalidates
-        # stale artifact-cache entries automatically. Truncate for readability.
-        DEV_HASH=$(printf %s '${{ hashFiles(format('{0}/solx-dev/src/llvm/**', inputs.working-dir)) }}' | head -c 8)
+        # solx-dev composes cmake args and flag defaults outside `llvm/**` too;
+        # hash every module the LLVM build reads so flag changes re-key.
+        DEV_HASH=$(printf %s '${{ hashFiles(format('{0}/solx-dev/src/llvm/**', inputs.working-dir), format('{0}/solx-dev/src/arguments/llvm/**', inputs.working-dir), format('{0}/solx-dev/src/build_type.rs', inputs.working-dir), format('{0}/solx-dev/src/ccache_variant.rs', inputs.working-dir), format('{0}/solx-dev/src/utils.rs', inputs.working-dir)) }}' | head -c 8)
         KEY="v1-llvm-${{ runner.os }}-${{ runner.arch }}"
         [ '${{ inputs.build-type }}' != 'Release' ] && KEY="${KEY}-${{ inputs.build-type }}"
         [ '${{ inputs.enable-mlir }}' = 'true' ] && KEY="${KEY}-mlir"
         [ -n '${{ inputs.sanitizer }}' ] && KEY="${KEY}-${{ inputs.sanitizer }}"
         [ '${{ inputs.enable-coverage }}' = 'true' ] && KEY="${KEY}-coverage"
         [ '${{ inputs.enable-assertions }}' != 'true' ] && KEY="${KEY}-no-assertions"
-        # `enable-utils` extends the install-distribution whitelist with
-        # FileCheck (see solx-dev's `build_opts_distribution`), so the install
-        # prefix actually differs by this flag. `enable-tests` is kept
-        # defensively even though no current workflow sets it; if one later
-        # does, LLVM_BUILD_TESTS=On could leak test artifacts into the prefix
-        # or change cmake-exports content.
-        # Normalize tests→utils because `--enable-tests` implies `--enable-utils`.
+        # `enable-utils` adds FileCheck to the install prefix; `enable-tests`
+        # implies it, so normalize before keying.
         UTILS='${{ inputs.enable-utils }}'
         [ '${{ inputs.enable-tests }}' = 'true' ] && UTILS='true'
         [ "$UTILS" = 'true' ] && KEY="${KEY}-utils"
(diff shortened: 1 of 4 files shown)


nebasuke (changes requested):
Thanks for picking this up! 

Some Claude assisted comments (after some back and forth). One I severely downgraded as I think it's optional, see what you want to do with it. I did spend some time digging deeper in what files actually change the cache key as you (likely inadvertently) made a change there. You might want to do a second pass.

nebasuke:
The fold itself looks right, and it's validated by this PR's own compile-benchmark run: both LLVM and solc mirrored on the main side in ~0.1s with the whole build chain skipped. Nice.

One substantive issue: the PR body says "Key strings are unchanged: old and new compositions verified byte-identical … so no cached artifact goes cold", but the hash *inputs* were broadened — `solx-dev/src/*.rs` was added to LLVM's `DEV_HASH` and solc's `SCRIPT_HASH` — so the key *values* changed. This PR's own run shows it: LLVM computes `…-dev26b75e96-…` while main's cache holds `…-devd05cb5e1-…`, and the log says `Cache not found for input keys` for both LLVM and solc (`scripta3548091` vs cached `script622e2341`). So every platform's artifact cache goes cold until cache-warmup re-saves on main — PR runs can't save. The macOS/Windows legs on this PR are paying full LLVM rebuilds right now.

The broadening is defensible — `BuildType` and `CcacheVariant::cmake_args` really do live outside `src/llvm/` — but `src/*.rs` over-captures. I traced all six files it matches against the actual `llvm build` / `solc build` code paths:

- `build_type.rs`, `ccache_variant.rs` — used by both, compose cmake flags. Belong.
- `utils.rs` — used by both, and not just plumbing: `get_xcode_version` gates cmake flags on macOS, `path_windows_to_unix` on Windows, `ninja()` picks `install` vs `install-distribution`. Belongs.
- `lib.rs` — module wiring only. Harmless either way.
- `shim.rs`, `workflow.rs` — test-harness only (the compiler-invocation shim from #529, the tester `Workflow` enum). Zero references from `src/llvm/` or `src/solc/`. Both were touched this year as part of harness work, and each future touch would cold-rebuild LLVM on every platform plus invalidate the solc caches for nothing.

This is the change I'm requesting: replace the `src/*.rs` glob with the three files the builds actually depend on — extra patterns to `hashFiles` in build-llvm, and named files instead of `-maxdepth 1 -name '*.rs'` in build-solc's `script_hash` (keeps the two symmetric):

```
format('{0}/solx-dev/src/build_type.rs', …), format('{0}/solx-dev/src/ccache_variant.rs', …), format('{0}/solx-dev/src/utils.rs', …)
```

Related gap, take it or leave it for a follow-up: the real invocation chain is `src/solx_dev/main.rs` → `src/arguments/{llvm,solc}/build.rs` (clap) → the build functions, and neither the old nor the new key hashes that layer — `default_value_t = BuildType::Release` lives in `arguments/llvm/build.rs`, and that layer had 14 commits this year. A default change there alters the artifact without invalidating anything. If catching flag changes outside `src/{llvm,solc}/` is the goal, `src/arguments/{llvm,solc}/**` + `main.rs` is the set that completes it.

Either way the body's cache-continuity claim needs correcting, and it'd be worth triggering cache-warmup right after merge so open PRs stop cold-building.

.github/actions/build-llvm/action.yml
```diff
@@ -74,39 +81,76 @@ runs:
...
-        [ "$UTILS" = 'true' ] && KEY="${KEY}-utils"
-        [ '${{ inputs.enable-tests }}' = 'true' ] && KEY="${KEY}-tests"
-        KEY="${KEY}-dev${DEV_HASH}-${SHA}"
-        [ -n '${{ steps.msys-toolchain.outputs.hash }}' ] && KEY="${KEY}-msys${{ steps.msys-toolchain.outputs.hash }}"
-        echo "key=${KEY}" | tee -a "${GITHUB_OUTPUT}"
+        # solx-dev owns LLVM-build cmake args: `llvm/**` plus the crate's
+        # top-level modules, which compose flags from outside that tree
+        # (`CcacheVariant::cmake_args`, `BuildType`). Hash both so any
+        # Rust-side flag change invalidates stale artifact-cache entries
+        # automatically. Truncate for readability.
```
nebasuke: Non-blocking observation, and I don't think we should guard against it — noting it so the limitation is on record. The mirror key is composed from the sibling's SHA + dev-hash but *this instance's* inputs for every other dimension, so input-derived dimensions cancel out of the comparison and the predicate reduces to SHA + dev-hash equality. The body's "cannot drift into a subset as key dimensions grow" therefore holds for content-derived dimensions, not input-derived ones — those rely on the paired callsites staying symmetric, exactly as this comment says. If we ever wanted the exact predicate, the shape would be exposing the composed key as an output and passing the PR-side step's key in as a `mirror-key` input. I considered asking for that and decided against: the two calls of a pair sit adjacent in the same workflow file, so an asymmetric edit is exactly what review catches, and the wiring cost isn't worth insuring a rare, review-visible mistake. Fine as is.

.github/actions/build-llvm/action.yml
```diff
@@ -57,7 +64,7 @@ inputs:
       Disable it when the solx-llvm tree changes every run (e.g. solx-llvm's own
       regression CI) or when later steps need the build directory: an artifact hit
       skips the build entirely, so target-llvm/build-final would not exist.
-      When disabled, the `cache-hit` output is always empty.
+      When disabled, `cache-hit` reports a mirror hit or nothing at all.
```
nebasuke, line 70: Nit: with `cache-hit` now being the `==`/`||` expression, it always evaluates to `'true'`/`'false'` and can't be empty anymore — "or nothing at all" is stale. Suggest: "When disabled, `cache-hit` reports whether the mirror hit." No consumers today (checked workflows and composite actions), so purely doc accuracy.

hedgar2017:
@nebasuke fixed in b61523b5, second pass done: dropped the glob and hashed the named set — `build_type`, `ccache_variant`, `utils`, plus `arguments/{llvm,solc}` from your follow-up, since that is where the flag defaults live and re-keying once beats re-keying twice. `main.rs` and `lib.rs` stay out: no `BuildType`/cmake references, dispatch and module wiring only — the same standard that excluded `shim`/`workflow`. Body corrected. cache-warmup's push trigger is path-filtered to submodule bumps, so it will not fire on this merge by itself — I will dispatch it manually right after merging.

.github/actions/build-llvm/action.yml
hedgar2017: Went with the exact shape after all: `cache-key` output, `mirror-key` input, the PR-side step passes its key. The wiring cost inverted once the action stopped recomputing the sibling's key — the dual key composition and its caveat comment fold away, net fewer lines than the approximate predicate.

.github/actions/build-llvm/action.yml
hedgar2017: Reworked the expression instead: it is a plain pass-through again (`cache-hit || mirrored`), which can be empty, so the sentence holds as written.

nebasuke (approved):
Thanks!

.github/actions/build-llvm/action.yml
```diff
@@ -105,8 +109,22 @@ runs:
...
+    - name: Mirror LLVM from sibling checkout
+      if: inputs.mirror-key != ''
+      id: llvm-mirror
+      shell: bash
+      working-directory: ${{ inputs.working-dir }}
+      run: |
+        if [ '${{ steps.llvm-cache.outputs.key }}' = '${{ inputs.mirror-key }}' ]; then
+          mkdir -p target-llvm
+          ln -snf "${GITHUB_WORKSPACE}/target-llvm/target-final" target-llvm/target-final
+          echo "mirrored=true" | tee -a "${GITHUB_OUTPUT}"
```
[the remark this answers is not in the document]
hedgar2017: The mechanism is real — `-n` only guards a destination that is already a symlink (keeping the step idempotent and keeping `-f` from dereferencing into the PR's real artifacts); against a real directory, both GNU and BSD `ln` silently link inside it. But the precondition cannot arise: this step only runs in the sibling invocation (`mirror-key` is only passed for the main-side build), and both callsites wipe the sibling checkout wholesale before it exists — the `rm -rf temp-solx-main` step in compile-benchmark.yaml and integration-tests.yaml, added precisely because `git clean` does not recurse into submodules on the persistent workspace. An `rm -rf` here would re-check what the callers guarantee.

.github/actions/build-solc/action.yml
```diff
@@ -31,14 +42,31 @@ runs:
...
+    - name: Mirror solc from sibling checkout
+      if: inputs.mirror-key != ''
+      id: solc-mirror
+      shell: bash
+      working-directory: ${{ inputs.working-dir }}
+      run: |
+        if [ '${{ steps.solc-cache.outputs.key }}' = '${{ inputs.mirror-key }}' ]; then
+          ln -snf "${GITHUB_WORKSPACE}/solx-solidity/build" build
+          ln -snf "${GITHUB_WORKSPACE}/solx-solidity/boost" boost
+          echo "mirrored=true" | tee -a "${GITHUB_OUTPUT}"
```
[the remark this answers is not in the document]
hedgar2017: Same as the build-llvm thread: the `ln` behavior is real, but the destination cannot pre-exist — the mirror only runs in the sibling invocation, and both callsites `rm -rf temp-solx-main` before checking it out (added in integration-tests.yaml precisely for stale `solx-solidity/{build,boost}` on the persistent workspace). An `rm -rf` here would re-check what the callers guarantee.

PR #619, opened 2026-07-29
ci: fold benchmark mirror into the build-llvm composite action

hedgar2017:
Mirrors #618 onto the main-slang line and makes the coverage and solc plumbing #617 left dormant actually fire:

- `build-llvm` matches #618: a `cache-key` output and a `mirror-key` input replace the share-llvm bash in `integration-tests.yaml` and `compile-benchmark.yaml` — the main-slang side receives the PR invocation's key and symlinks the workspace-root artifacts when its own key is equal. Both dev-script hashes widen to the modules the builds read outside `src/{llvm,solc}`: `arguments/{llvm,solc}`, `build_type`, `ccache_variant`, and `utils`; solc keeps only the hash fix, since no job on this line mirrors it.
- `coverage.yaml` builds and tests the slang pipeline on `main-slang*` triggers, dropping the solc it needed only to link the default pipeline.
- `build-toolchain` builds the lit oracle's Release solc, so `warm-solc` warms the key `build-and-test` restores on all five platforms rather than a RelWithDebInfo key nothing reads; `warm-llvm-integration` drops its solc half, which no integration job consumes.
- `target-slang` collapses into `target` across the cargo aliases, CI paths, devcontainer volume, and lit config, the isolation being pointless once no job on this line builds the linking default pipeline.

### DIFF
--- a/.github/actions/build-toolchain/action.yml
+++ b/.github/actions/build-toolchain/action.yml
@@ -6,8 +6,8 @@ description: >-
   define the standard-config cache keys shared by test.yaml and
   cache-warmup.yaml — changing them here re-keys both workflows together.
   Workflows needing a different LLVM configuration (sanitizer, coverage,
-  integration, release) call build-llvm/build-solc directly and pair with
-  their own cache-warmup job.
+  integration, release) call build-llvm and build-solc directly as needed
+  and pair with their own cache-warmup job.
 
 inputs:
   llvm:
--- a/.github/actions/build-toolchain/action.yml
+++ b/.github/actions/build-toolchain/action.yml
@@ -5,9 +5,9 @@ description: >-
   on macOS), then the LLVM and solc builds. The build parameters below
   define the standard-config cache keys shared by test.yaml and
   cache-warmup.yaml — changing them here re-keys both workflows together.
-  Workflows needing a different LLVM
-  configuration (sanitizer, coverage, integration, release) call
-  build-llvm/build-solc directly and pair with their own cache-warmup job.
+  Workflows needing a different LLVM configuration (sanitizer, coverage,
+  integration, release) call build-llvm/build-solc directly and pair with
+  their own cache-warmup job.
 
 inputs:
   llvm:
@@ -73,10 +73,15 @@ runs:
       shell: bash
       run: rm -fv "$(rustc --print sysroot)/bin/libwinpthread-1.dll"
 
+    # TODO: remove this step once the lit tests stop running against solc.
+    #
+    # Release: RelWithDebInfo bakes -g3 debug info into the statically
+    # linked solc.exe, producing a ~2 GB PE that the Windows loader
+    # rejects ("Exec format error"), and the lit tests execute solc.
     - name: Build solc
       if: inputs.solc == 'true'
       uses: ./.github/actions/build-solc
       with:
-        cmake-build-type: RelWithDebInfo
+        cmake-build-type: Release
         working-dir: 'solx-solidity'
         enable-mlir: 'true'
--- a/.github/workflows/compile-benchmark.yaml
+++ b/.github/workflows/compile-benchmark.yaml
@@ -97,47 +97,21 @@ jobs:
       # Release build without assertions: matches the released binary's
       # configuration so the three-way comparison is apples-to-apples.
       - name: Build LLVM
+        id: build-llvm-pr
         uses: ./.github/actions/build-llvm
         with:
           build-type: Release
           enable-assertions: 'false'
           enable-mlir: 'true'
 
-      # Submodule SHA alone is not sufficient: `solx-dev llvm build` is
-      # driven by `solx-dev/src/llvm/*`, so a PR that edits those scripts
-      # without bumping the solx-llvm submodule would silently share its
-      # own LLVM as the "main" baseline. Mirror the SHA + script-hash
-      # check used by integration-tests.yaml.
-      - name: Share LLVM with main checkout
-        id: share-llvm
-        run: |
-          # Content hash of a directory tree (order-independent).
-          hash_tree() {
-            find "$1" -type f -print0 | sort -z | xargs -0 cat | sha256sum | cut -d' ' -f1
-          }
-          MAIN_SHA=$(git -C temp-solx-main/solx-llvm rev-parse HEAD)
-          PR_SHA=$(git -C solx-llvm rev-parse HEAD)
-          SCRIPT_PR=$(hash_tree solx-dev/src/llvm)
-          SCRIPT_MAIN=$(hash_tree temp-solx-main/solx-dev/src/llvm)
-          if [ "$MAIN_SHA" = "$PR_SHA" ] && [ "$SCRIPT_MAIN" = "$SCRIPT_PR" ]; then
-            echo "LLVM SHAs and build scripts match ($PR_SHA / ${SCRIPT_PR:0:8}), sharing via symlink"
-            mkdir -p temp-solx-main/target-llvm
-            rm -rf temp-solx-main/target-llvm/target-final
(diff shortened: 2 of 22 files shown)


nebasuke:
Verified the fold against #618: `build-llvm/action.yml` is byte-identical between this branch and #618's current head (`b61523b5`), so the two lines won't drift, and the narrowed hash set from the #618 review (`llvm/**` + `arguments/llvm/**` + `build_type/ccache_variant/utils`) is in on both sides, symmetric with solc's `SCRIPT_HASH`. The mirror predicate also behaved correctly live in this PR's own runs: with the submodule bumped on the PR side only, both integration and compile-benchmark correctly declined the mirror (`…-7d0702e1` vs `…-acacc0c8`) and built separately. I also checked the collapse for stragglers — no `target-slang` left anywhere at head (release.yaml, run.sh, lit, devcontainer, docs all consistent), and `solx-tester` doesn't depend on the `solx` crate, so the shared `target/` doesn't reintroduce the inkwell llvm-linking poisoning that motivated the isolation. The `rust-unit-tests` input drops are no-ops (they were passing the action's defaults), and warm-solc's `Release-end` restore key now matches what build-toolchain actually builds — the remaining `RelWithDebInfo` keys in cache-warmup are all LLVM configs, not solc.

Three things from this PR's own CI runs, in descending order of importance (slang-pipeline compile failures excluded throughout — those are expected WIP):

**1. The main-side binary path is broken until this merges — and the old `cp` bridge was the fix.** `temp-solx-main` is checked out from `main-slang`, whose own `.cargo/config.toml` still has `--target-dir target-slang` in the aliases, so "Build solx (main)" writes `temp-solx-main/target-slang/…/solx` while the PR-side tooling now looks in `temp-solx-main/target/…`. Compile-benchmark fails hard (`run.sh: ./temp-solx-main/target/x86_64-unknown-linux-gnu/release/solx: No such file or directory`), and integration's "Run solx-tester benchmarks (main-slang)" dies the same way (`continue-on-error` masks it, but the report loses the main column). The removed `cp target-slang/release/solx target/release/solx` was exactly this bridge. Suggest a transitional guard in the two main-side build steps — `[ -x target-slang/…/solx ] && mkdir -p target/… && cp …` with a TODO to drop after merge — so this PR's runs validate end-to-end and stale PRs branched pre-merge don't hit the reverse skew either. (Inline suggestion below.)

**2. Coverage exits before uploading anything.** The integration-test leg runs `complex/default`, which the slang pipeline can't compile yet, so the job dies at that step — before the profdata merge and Codecov upload. The expected-failure state is fine, but as wired, activating slang coverage produces *no* coverage at all (and no Codecov base on `main-slang` pushes, which the workflow's own comment says is the point of the push trigger). Until the pipeline matures, either make the integration leg tolerant (`|| true` on the solx-dev invocation — profraw from the passing portion still merges) or scope it to slang-supported paths. Otherwise the unit-test coverage that does work never gets published.

**3. Sanitizer: this PR's runs are the first time the job has ever executed on the main-slang line — it's bring-up, not regression.** Every earlier "green" sanitizer run on this branch was a label-gated skip (`sanitizer=skipped`, aggregator green), and `main-slang` itself has zero sanitizer runs — so the melior E0463 wasn't caused by the submodule bump; it was there all along, waiting for the first real execution. The `3cc09dea` fix for it is right and empirically confirmed: melior now compiles, and the shape is sound (proc macro is compile-time tooling, so no sanitizer coverage is lost; the extra build's inputs exactly match build-toolchain's so it shares the standard artifact key — a ~30s restore once cache-warmup has run post-merge; the `mv`+`rm -rf build-final` hygiene correctly keeps the sanitized build from reusing the standard cmake tree; un-forcing `TABLEGEN_210_PREFIX` is required since cargo `[env] force` beats process env, at the documented cost that a stale env var now wins locally).

The run on `3cc09dea` then surfaced the next pre-existing layer: linking the sanitized test binaries fails with `undefined reference to LLVMAssembleEVM / LLVMGetImmutablesEVM / LLVMDisposeImmutablesEVM` — the lld-c EVM API from `lld/lld-c/LLDAsLibraryC.cpp`, referenced by the inkwell fork and linked via solx-mlir's build.rs (`-llldC -llldCommon -llldELF`). Both LLVM installs in that run contain `liblldC.a`, and the identical link works in the non-sanitized slang-tests on all five platforms, so it's specific to the ASan link. Given the job never ran on this line, I'd treat sanitizer bring-up as its own track: either keep iterating here if the lld-c fix is close, or drop `ci:sanitizer` from this PR and file the bring-up separately so the CI-plumbing fold isn't hostage to it.

Cache continuity is the same story as #618 — everything re-keys, PR runs can't save — but here the merge itself touches `solx-llvm`, so cache-warmup fires on push without manual help, and the daily cron backstops the solc re-key.

.github/workflows/compile-benchmark.yaml
```diff
@@ -190,8 +164,8 @@ jobs:
       - name: Run benchmark
         run: |
           ./tests/benchmark/run.sh \
-            --bin pr=./target-slang/x86_64-unknown-linux-gnu/release/solx \
-            --bin main=./temp-solx-main/target-slang/x86_64-unknown-linux-gnu/release/solx \
+            --bin pr=./target/x86_64-unknown-linux-gnu/release/solx \
+            --bin main=./temp-solx-main/target/x86_64-unknown-linux-gnu/release/solx \
```
nebasuke, line 168: Until this merges, this path doesn't exist: `temp-solx-main` is checked out from `main-slang`, whose own `.cargo/config.toml` still has `--target-dir target-slang` in the `build-slang` alias, so "Build solx (main)" above writes the binary to `temp-solx-main/target-slang/x86_64-unknown-linux-gnu/release/solx` — this run's failure. The removed `cp` was the bridge. Transitional guard in "Build solx (main)":

```yaml
run: |
  ${SFW_PREFIX:-} cargo fetch --locked
  ${SFW_PREFIX:-} cargo build-slang --frozen --release --target x86_64-unknown-linux-gnu
  # TODO(remove after merge): pre-merge main-slang still writes to target-slang
  if [ -x target-slang/x86_64-unknown-linux-gnu/release/solx ]; then
    mkdir -p target/x86_64-unknown-linux-gnu/release
    cp target-slang/x86_64-unknown-linux-gnu/release/solx target/x86_64-unknown-linux-gnu/release/solx
  fi
```

(integration-tests' main-slang variant needs the same, with `target-slang/release/solx` → `target/release/solx`)

.github/actions/build-toolchain/action.yml
```diff
@@ -5,9 +5,9 @@ description: >-
   on macOS), then the LLVM and solc builds. The build parameters below
   define the standard-config cache keys shared by test.yaml and
   cache-warmup.yaml — changing them here re-keys both workflows together.
-  Workflows needing a different LLVM
-  configuration (sanitizer, coverage, integration, release) call
-  build-llvm/build-solc directly and pair with their own cache-warmup job.
+  Workflows needing a different LLVM configuration (sanitizer, coverage,
+  integration, release) call build-llvm/build-solc directly and pair with
+  their own cache-warmup job.
```
nebasuke: Nit: the description still lists coverage among the workflows that "call build-llvm/build-solc directly", but coverage no longer builds solc at all on this line. Purely doc accuracy.

nebasuke (approved):
Thank you!

.github/actions/build-toolchain/action.yml
hedgar2017: Fixed!

.github/workflows/compile-benchmark.yaml
hedgar2017: Self-healing on merge! Leaving as is!

PR #621, opened 2026-07-30
fix(slang): qualify function symbols with the node id

### LINKED ISSUE #560: Function symbol mangling in the Slang frontend collides under inheritance and function-pointer overloads

hedgar2017:
Fixes #560 by qualifying every registered function symbol with its node id:

- Overloads differing only in a function-typed parameter, `apply_pointer(function(uint256) ...)` next to `apply_pointer(function(bool) ...)`, both mangled to `apply_pointer(function)` and died on symbol redefinition; they now emit distinct `{signature}_{node id}` symbols, the qualification the C++ frontend's mangling applies, which also covers same-signature overrides once inheritance lands.

### DIFF
--- a/solx-mlir/tests/lit/function_overloads.sol
+++ b/solx-mlir/tests/lit/function_overloads.sol
@@ -1,0 +1,77 @@
+// RUN: solx --emit-mlir=sol %s | FileCheck %s
+// RUN: solc --mlir-action=print-init %s 2>/dev/null | FileCheck %s
+
+// CHECK: sol.func @{{.*pick.*}}(%{{.*}}: ui256) -> ui256
+// CHECK: sol.func @{{.*pick.*}}(%{{.*}}: i1) -> i1
+
+// CHECK: sol.func @{{.*sum.*}}(%{{.*}}: ui256) -> ui256
+// CHECK: sol.func @{{.*sum.*}}(%{{.*}}: ui256, %{{.*}}: ui256) -> ui256
+
+// CHECK: sol.func @{{.*run_nested.*}}
+// CHECK:   %[[INNER:.*]] = sol.call @{{.*sum.*}}(%{{.*}}) : (ui256) -> ui256
+// CHECK:   sol.call @{{.*sum.*}}(%[[INNER]], %{{.*}}) : (ui256, ui256) -> ui256
+
+// CHECK: sol.func @{{.*get.*}}(%{{.*}}: ui256) -> ui256 attributes {{.*}}selector = -1794649190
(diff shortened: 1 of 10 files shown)


solx-mlir/tests/lit/function_overloads.sol
```diff
@@ -0,0 +1,77 @@
...
+// CHECK: sol.func @{{.*apply_pointer.*}}(%{{.*}}: !sol.func_ref<(ui256) -> ui256>) -> ui256
+// CHECK:   sol.icall %{{[0-9]+}}(%{{.*}}) : !sol.func_ref<(ui256) -> ui256>, (ui256) -> ui256
+// CHECK: sol.func @{{.*apply_pointer.*}}(%{{.*}}: !sol.func_ref<(i1) -> i1>) -> i1
+// CHECK:   sol.icall %{{[0-9]+}}(%{{.*}}) : !sol.func_ref<(i1) -> i1>, (i1) -> i1
+
+// CHECK: sol.func @{{.*run.*}}
+// CHECK:   sol.call @{{.*apply_pointer.*}}(%{{.*}}) : (!sol.func_ref<(i1) -> i1>) -> i1
+// CHECK:     sol.call @{{.*pick.*}}(%{{.*}}) : (i1) -> i1
+
+contract C {
```
PavelKopyl, line 26: Let's also check (in another file) that a file-level free `add(T,T)` function bound as an operator does not conflict with a contract member `add(T,T)`?

Also makes sense to check the example form the issue:
```
function f() pure returns (uint) { return 1337; }
contract C {
  function f() public pure returns (uint) { return f(); }
}
```

PavelKopyl (approved):
LGTM, thanks!

solx-mlir/tests/lit/function_overloads.sol
hedgar2017: Thanks! Adding as `function_shadowing.sol`.

PR #625, opened 2026-07-30
feat(slang): named arguments

hedgar2017:
Completes named-argument emission for every callee kind the frontend lowers today:

- `f({y: .., x: ..})` on a direct call and `S({b: .., a: ..})` on a struct constructor evaluate their arguments in the callee’s parameter and struct-member declaration order, the order legacy solc evaluates them in, resolved through the node id each label binds to so an overloaded callee is selected by its binding and never by name.
- `emit E({..})` and `revert E({..})` order through that same binding instead of matching parameter names as strings.
- The empty `f({})` braces order against no declarations, so every callee kind that declares no parameters lowers them as its no-argument form: `revert({})` to `sol.revert`, `array.push({})` and `array.pop({})` to `sol.push` and `sol.pop`, and a call through an internal function pointer to `sol.icall`.

Also optimizes and de-duplicates some macro:

- `sol_ops!` takes a `checked(CheckedOp)` marker on the declaration itself, so the checked/unchecked arithmetic pair is stamped by the same terminal rule as every other method rather than by a dedicated one, the `checked: bool` selector threaded through the operation slot.
- `mlir_op!` absorbs `mlir_op_void!`, a `; ()` tail marking the value-less form.

Suppresses warnings from the dialect that have been polluting our build and CI logs:

- `sol_attr_stubs.cpp` takes the MLIR headers through `-isystem`, so the `-Wunused-parameter` wall their tblgen output raises no longer lands on every build.

### DIFF
--- a/solx-mlir/src/macros.rs
+++ b/solx-mlir/src/macros.rs
@@ -1,12 +1,7 @@
 //!
 //! ODS op-construction macros.
 //!
-//! `mlir_op_build!` / `mlir_op!` / `mlir_op_void!` / `mlir_region_op!` collapse the ceremony of an
-//! ODS-generated op builder (the `(context, unknown_location)` head and `.build().into()` tail) so a
-//! site states only the op name and its setters.
-//!
 
-/// Builds an inlined dialect op and yields it as an `Operation`, without appending.
 macro_rules! mlir_op_build {
     ($context:expr, $operation:ident $(.$method:ident($($argument:expr),* $(,)?))*) => {
         $operation::builder($context.melior, $context.location())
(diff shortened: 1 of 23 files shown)


solx-mlir/src/macros.rs
```diff
@@ -16,13 +11,13 @@ macro_rules! mlir_op_build {
     };
 }
 
-/// Builds an inlined dialect op ([`mlir_op_build!`]), appends it to `$block`, and
-/// returns its single result value. The `expect` message is derived from the op.
-/// Omitting `$block` appends at the `current_block()` cursor.
```
PavelKopyl, line 21: Let's restore the macro documentation.

solx-slang/src/contract/function/expression/call/mod.rs
```diff
@@ -124,27 +120,56 @@ impl Call {
...
+                    &named,
+                    function_definition
+                        .parameters()
+                        .iter()
+                        .map(|parameter| parameter.node_id()),
+                ),
+                Self::TypeConversion
+                | Self::Builtin(_)
+                | Self::Member(_)
+                | Self::FunctionPointer => Vec::new(),
```
PavelKopyl: Will it make sense to add an assert that we really do not have labeled parameters for these kind of calls (just it case of a binder regression)?

PavelKopyl (approved):
LGTM.

PavelKopyl:
I suggest to add  a side-effect evaluation order test for named args (for instance here: `tests/solidity/simple/evaluation_order/named_arguments.sol` as a runtime, or to lit tests), with something like:
`f({b: t(2), a: t(1)})`, where asserting `t(1)` runs first.

solx-mlir/src/macros.rs
hedgar2017: Restored!

solx-slang/src/contract/function/expression/call/mod.rs
hedgar2017: Thanks!
Indeed this catch-all is too permissive. `assert`s are not done in the codegen to avoid pretending to be the semantic validator, but I managed to tighten it so it looks good.
Also found 2 missing unsupported cases in the process!

hedgar2017:
> I suggest to add a side-effect evaluation order test for named args (for instance here: `tests/solidity/simple/evaluation_order/named_arguments.sol` as a runtime, or to lit tests), with something like: `f({b: t(2), a: t(1)})`, where asserting `t(1)` runs first.

Added!

PR #627, opened 2026-07-30
chore(deps): source tblgen from NomicFoundation

hedgar2017:
Moves `tblgen` off the personal `hedgar2017` fork and onto `NomicFoundation/tblgen-rs`, so every git source we build from is org-owned.

It is not a direct dependency — the fork URL is hard-coded in melior, so it was changed there (NomicFoundation/melior@d188691b) rather than overridden here with a `[patch]`. This PR bumps the melior rev to pick that up and swaps the `deny.toml` allow-list entry to match. The tblgen rev is unchanged, so the compiled code is identical; only the fetch URL differs.


nebasuke (approved):
Thanks for the clear explanation!

PR #636, opened 2026-08-05
refactor: hold the runtime child in a Dependencies field

### LINKED ISSUE #633: Dependencies: hold the runtime child in a field instead of matching its name
`Dependencies::push` has to put the runtime child at index 0, because that is where `LLVMAssembleEVM` looks for it. Right now it finds the child by stripping `_deployed` or `.runtime` off the dependency name and comparing to `self.identifier`, even though all seven construction sites have the identifier on hand. solc's MLIR path just writes `evmObj.runtimePart` into slot 1.

So make it a field:

```rust
pub struct Dependencies {
    pub identifier: String,
    /// The runtime object the deploy code returns. `None` in a runtime segment.
    pub runtime: Option<String>,
    pub inner: Vec<String>,
}
```

`new(identifier, runtime)`, `push` loses its branch, and an `iter()` yields the runtime child then `inner` so callers stop reassembling the ordering. `is_runtime_child` and `RUNTIME_SEGMENT_SUFFIX` go away.

Two things to watch:

- The name-sniffing tests (`.runtime` suffix, foreign object named like a runtime child) become meaningless and should go; the encounter-order one gets rewritten against `iter()`.
- `Dependencies` is serialized into standard JSON (`contracts[path][name].mlir.deploy_dependencies`) and over the worker CBOR frame, so the new field shows up in the output.

hedgar2017:
Fixes #633

### DIFF
--- a/solx-core/src/build/contract/object.rs
+++ b/solx-core/src/build/contract/object.rs
@@ -146,10 +146,12 @@ impl Object {
     ) -> anyhow::Result<inkwell::memory_buffer::MemoryBuffer> {
         let memory_buffer = self.to_memory_buffer()?;
 
-        let mut memory_buffers = Vec::with_capacity(1 + self.dependencies.inner.len());
+        let mut memory_buffers = Vec::with_capacity(
+            1 + usize::from(self.dependencies.runtime.is_some()) + self.dependencies.inner.len(),
+        );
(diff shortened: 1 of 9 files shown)


abinavpp (approved):
nice, thank you!

PR #637, opened 2026-08-07
fix(slang): type a pushed reference element as its own place

hedgar2017:
Fixes the array push lowering fabricating `!sol.ptr<element, Storage>` slots for reference-typed elements, against the `Sol_GepOp` rule that a reference in storage is its own place.

- **`s.push().a = 1`** no longer crashes the Sol pass pipeline: the slot type comes from `SourceUnitScope::pointer`, the rule's owner, so a struct element's slot is the struct itself and the member `gep` lowers.
- **`nested.push().push(2)`** no longer crashes: the inner push sees the array place instead of a pointer to it.
- **`s.push(Y(1, 2))`** was silently mis-lowered — a single-word store of the memory pointer into a two-slot struct; it now assigns through `Place::assign` and takes the `sol.copy` branch.
- The place path (`push() = v`) reads the element back off the slot alone — the pointee where the slot points at one, the slot itself otherwise — inverting `SourceUnitScope::pointer` rather than consulting the binder a second time. `Type::is_pointer` is the `sol::PointerType` predicate this needs.

Three new dual-oracle cases join `array_string_ops.sol` (`pushStructMember`, `pushStructValue`, `pushNested`). Tester: 8792 → 8825 passing against the same base, zero regressed files. Pre-existing and untouched: push in value position (`uint y = arr.push();`, `push_no_args_1d/2d`).

### DIFF
--- a/solx-slang/src/contract/function/expression/call/mod.rs
+++ b/solx-slang/src/contract/function/expression/call/mod.rs
@@ -806,32 +806,50 @@ impl Call {
                     return Vec::new();
                 }
 
-                let (element_type, slang_location) = match &base_slang_type {
-                    Type::Array(array_type) => (
-                        scope.resolve_type(&array_type.element_type(), None),
-                        array_type.location(),
-                    ),
+                let (element_type, slot_type) = match &base_slang_type {
+                    Type::Array(array_type) => {
+                        let element_slang_type = array_type.element_type();
+                        let element_type = scope.resolve_type(&element_slang_type, None);
+                        (
+                            element_type,
+                            scope.pointer_type(
+                                &element_slang_type,
+                                element_type,
+                                solx_utils::DataLocation::from_slang(array_type.location(), None),
+                            ),
+                        )
+                    }
                     Type::Bytes(bytes_type) => {
-                        (MlirType::byte(scope.melior), bytes_type.location())
+                        let element_type = MlirType::byte(scope.melior);
+                        (
+                            element_type,
+                            MlirType::pointer(
+                                scope.melior,
+                                element_type,
+                                solx_utils::DataLocation::from_slang(bytes_type.location(), None),
+                            ),
+                        )
                     }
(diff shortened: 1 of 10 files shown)


solx-slang/src/contract/function/expression/call/mod.rs
```diff
@@ -925,7 +943,13 @@ impl<'contract, 'source_unit, 'context> FunctionScope<'contract, 'source_unit, '
             .into_iter()
             .next()
             .expect("an array push in place position yields the new element's slot");
-        (Place::from(slot), slot.r#type().element_type(0))
+        let slang_type = node.get_type().expect("the binder types every expression");
+        let element_type = if slang_type.is_reference_type() {
```
hedgar2017: I'm not sure about having 2 sources of truths here.
@abinavpp @ggiraldez do you think we should unify it somehow?

hedgar2017: @abinavpp I've tried to drop `sol.byte` in #638, but it looks like it's deeply woven into LLVM.
Please let me know if we can rework the dialect and drop it.

For the context, it looks like `byte` is an alias that is not valid in Solidity 0.8, and Slang doesn't emit it anywhere.

abinavpp: Re the two sources: agree - derive it from the push result's type alone (ptr -> pointee, else the type itself is the element); needs only an `is_pointer` FFI predicate.

Re `sol.byte`: keep it - it's `!sol.string`'s element type and carries packed-byte place semantics `bytes1` doesn't (mstore8 stores, masked loads, never-reverting cleanup); value-level casts between the two are no-ops.

hedgar2017: @abinavpp thanks! Fixed and closed #638. Please check.

abinavpp (approved):
lgtm, thank you!

PR #644, opened 2026-08-13
fix(solc): preserve standard JSON input on errors

### LINKED ISSUE #643: Solc::standard_json mutates its input when an error occurs
#### 📝 Description

`Solc::standard_json` temporarily modifies the caller-provided standard JSON input before invoking solc.

It normalizes and extends `settings.output_selection`, and clears the solx-specific optimizer fields `mode` and `size_fallback`. These fields are restored after a successful solc invocation, but an earlier error return leaves the caller's input modified.

`Solc` is publicly exported as part of the solc client library, so callers can observe this mutation after an unsuccessful invocation.

#### 🔄 Reproduction Steps

Call `Solc::standard_json` with a valid input and a base path containing an interior NUL byte:

```rust
let result = solc.standard_json(
    &mut input,
    true,
    Some("abc\0def"),
    &[],
    None,
);
```

CString::new rejects the base path before solc is invoked. After the error, the input's output_selection, optimizer.mode, and optimizer.size_fallback no longer match their original values.

#### 🤔 Expected Behavior

Solc::standard_json should leave the caller-provided input unchanged regardless of whether the operation succeeds or returns an error.

#### 😯 Current Behavior

The original output_selection and optimizer settings are restored only after the fallible solc invocation and output parsing complete successfully.

#### 🖥️ Environment

Current main.

#### 📋 Additional Context

The restoration should happen before propagating the result of the fallible solc invocation, rather than being duplicated at every individual error site.

Add any other context about the problem here. If applicable, add screenshots to help explain.

#### 📎 Log Output

```
Paste any relevant log output here.
```

questfever:
Closes https://github.com/NomicFoundation/solx/issues/643

## Description

`Solc::standard_json` temporarily adjusts the output selection and optimizer settings before passing the input to solc.

Previously, these fields were restored only after the complete invocation and output parsing succeeded. Errors produced while preparing paths, invoking solc, or parsing its output therefore left the caller-provided input modified.

This change captures the result of the fallible solc invocation, restores the original input settings, and only then propagates the result.

## Testing

Added a regression test that:

- invokes `Solc::standard_json` with a base path containing an interior NUL byte;
- verifies that the call returns an error;
- verifies that the complete standard JSON input remains unchanged.

### DIFF
--- a/solx/src/solc.rs
+++ b/solx/src/solc.rs
@@ -101,7 +101,10 @@ impl solx_core::Frontend for Solc {
         input_json.settings.optimizer.mode = None;
         input_json.settings.optimizer.size_fallback = None;
 
-        let input_string = serde_json::to_string(input_json)
+        let input_string = serde_json::to_string(input_json);
+        input_json.settings.output_selection = original_output_selection;
+        input_json.settings.optimizer = original_optimizer;
+        let input_string = input_string
             .map_err(|error| anyhow::anyhow!("solc standard JSON input serialization: {error}"))?;
         let input_c_string = CString::new(input_string)
             .map_err(|error| anyhow::anyhow!("solc standard JSON input CString: {error}"))?;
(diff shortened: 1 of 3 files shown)


hedgar2017 (approved):
Thanks! Adjusted the code to the repo standards.

PR #648, opened 2026-08-14
ci: cross-OS wall-clock benchmark of the pool dispatch modes

abinavpp:
Opt-in CI benchmark (label ci:mode-bench) comparing in-process threads vs the SOLX_SUBPROCESS worker pool on all slang-tests platforms. Numbers feed the threads-by-default discussion on #587.


abinavpp:
We compared multithreading against the multiprocess worker pool. The
benchmark scripts and the CI workflow that runs them are part of this
PR (`tests/benchmark/pool-modes/`); label a PR with `ci:mode-bench` to
run it. Results below are from
https://github.com/NomicFoundation/solx/actions/runs/31893500518 and
were reproduced across two runs. It is a pure mode comparison: both
sides run identical code, the only difference is whether a contract
compiles on a thread or in a worker process.

The input is generated: many tiny contracts, a few big ones (~3s of
codegen each), each in a clean variant and a variant that triggers
stack-too-deep, plus a mix of all four.

ratio = multiprocess/multithreading (>1 = threads faster):

| corpus | Linux x86 | Linux arm64 | macOS arm64 | macOS x86 |
|---|---|---|---|---|
| tiny-clean (200) | 1.03 | 0.99 | 0.96 | 0.98 |
| tiny-overflow (100) | 1.10 | 1.06 | 1.20 | **2.14** |
| big-clean (12) | 1.02 | 0.98 | 0.94 | 1.04 |
| big-overflow (6) | 1.02 | 1.00 | 0.98 | 0.97 |
| mixed (159) | 1.02 | 0.99 | 1.00 | 1.07 |

Parity everywhere except stack-too-deep on cheap contracts:
multiprocess kills and respawns the worker on every overflow, and
that cost scales with how expensive process spawning is on the
platform (2.1x on macOS x86).

Windows: multithreading is broken there (LLVM's PrettyStackTrace gets
corrupted under concurrent compilation - needs its own
investigation), so Windows keeps the multiprocess default; all tests
are green with that arrangement.

Not measured yet: real contracts (limited by current slang coverage),
and the optimizations only multithreading can unlock later, like
compiling each contract fully in memory instead of shipping it to the
job as text (~6-12% of a big contract's compile time today) and
in-memory stack-too-deep retries. No urgency to force the switch;
revisit with real contracts and a working Windows build once slang
matures.

abinavpp:
Folded into #587 (the benchmark harness, workflow, and report now live there).

PR #649, opened 2026-08-14
feat(slang): libraries and using for

hedgar2017:
Lowers libraries end to end: external delegatecall dispatch, internal calls by copy, the deployable library object, library-as-value expressions, and `--libraries` linking.

`cargo run-tester-slang`: 9505 → 10022 passed.

### DIFF
--- a/solx-slang/src/contract/function/expression/call/mod.rs
+++ b/solx-slang/src/contract/function/expression/call/mod.rs
@@ -30,6 +30,7 @@ use solx_utils::FunctionReferenceKind;
 
 use crate::contract::function::expression::call::external_callee::ExternalCallee;
 use crate::contract::function::expression::call::options::Options;
+use crate::contract::object::Object;
 use crate::scope::function::FunctionScope;
 use crate::scope::source_unit::SourceUnitScope;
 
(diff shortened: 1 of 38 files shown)


solx-slang/src/contract/function/expression/call/mod.rs
```diff
@@ -619,22 +687,149 @@ impl Call {
...
+        );
+        let converted = scope.external_arguments(arguments, &function.function_type.parameters);
+        let is_static = StateMutability::from(definition.attributes().mutability()).is_static();
+        Self::external_dispatch(
+            &function,
+            &converted,
+            &function.function_type.results,
+            address,
+            selector,
+            &Options::default(),
```
abinavpp: I think the gas option is skipped here - `L.f{gas: g}(x)` would forward all gas. Can you check?

abinavpp (approved):
lgtm, modulo one 1 question

solx-slang/src/contract/function/expression/call/mod.rs
hedgar2017: Thanks, checking now.

hedgar2017: Good ketch, fixed!

PR #651, opened 2026-08-16
feat(slang): transient storage

hedgar2017:
Wires transient storage end to end: layout, declarations, accesses, and getters.

`cargo run-tester-slang`: 10129 → 10217 passed.

### DIFF
--- a/solx-mlir/tests/lit/storage_layout_at.sol
+++ b/solx-mlir/tests/lit/storage_layout_at.sol
@@ -1,0 +1,25 @@
+// RUN: solx --emit-mlir=sol %s | FileCheck %s
+
+// solc emits a getter before the functions and solx after them, so this test
+// is solx-only.
+
+// CHECK: sol.state_var @{{.*low.*}} slot 100 offset 0 : ui64
+// CHECK: sol.state_var @{{.*high.*}} slot 100 offset 8 : ui64
+
+// CHECK: sol.func @{{.*write.*}}
+// CHECK:   %[[PTR:.*]] = sol.addr_of @{{.*high.*}} : !sol.ptr<ui64, Storage>
+// CHECK:   sol.store %{{.*}}, %[[PTR]] : ui64, !sol.ptr<ui64, Storage>
+
+// CHECK: sol.func @{{.*high.*}}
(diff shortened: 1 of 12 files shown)


solx-mlir/tests/lit/storage_layout_at.sol
```diff
@@ -0,0 +1,18 @@
...
+// CHECK: sol.state_var @{{.*low.*}} slot 100 offset 0 : ui64
+// CHECK: sol.state_var @{{.*high.*}} slot 100 offset 8 : ui64
+
+// CHECK: sol.func @{{.*write.*}}
+// CHECK:   %[[PTR:.*]] = sol.addr_of @{{.*high.*}} : !sol.ptr<ui64, Storage>
+// CHECK:   sol.store %{{.*}}, %[[PTR]] : ui64, !sol.ptr<ui64, Storage>
+
+contract Based layout at 100 {
+    uint64 low;
+    uint64 high;
```
PavelKopyl: You could declare, for instance, `high` as public to check the generated getter also reads from the slot 100.

PavelKopyl (approved):
LGTM, thanks!

If we do not have a runtime test that checks transient variables get cleared between transactions, it would be reasonable to add it.

solx-mlir/tests/lit/storage_layout_at.sol
hedgar2017: Done — `high` is public now and the fixture pins its getter:

```
// CHECK: sol.func @{{.*high.*}}
// CHECK:   %[[GETTER:.*]] = sol.addr_of @{{.*high.*}} : !sol.ptr<ui64, Storage>
// CHECK:   %[[VALUE:.*]] = sol.load %[[GETTER]] : !sol.ptr<ui64, Storage>, ui64
// CHECK:   sol.return %[[VALUE]] : ui64
```

so the getter is shown resolving the same symbol that carries `slot 100 offset 8`, rather than deriving an address of its own — the case no other fixture covered, since `getter.sol` only has getters at the default base.

One side effect: the file had to drop its `solc --mlir-action=print-init` run. solc emits a getter before the other functions and solx after them, so ordered CHECKs covering both `write` and `high()` cannot satisfy both frontends; the file now says so in a comment.

PR #652, opened 2026-08-16
feat(slang): immutables

hedgar2017:
Supports user-declared immutables: declarations, constructor writes and reads, runtime loads, getters, and declaration-site initializers. Every state variable's symbol now comes from one owner function.

`cargo run-tester-slang`: 10217 → 10323 passed.

### DIFF
--- a/solx-slang/src/contract/function/mod.rs
+++ b/solx-slang/src/contract/function/mod.rs
@@ -37,10 +37,11 @@ impl<'source_unit, 'context> ContractScope<'source_unit, 'context> {
             _ => None,
         };
         let signature = self.source_unit.function_signature(function);
+        let dispatch = FunctionDispatch::from(function);
         let state_mutability = StateMutability::from(function.attributes().mutability());
         let entry = signature.define(
             selector,
-            FunctionDispatch::from(function),
+            dispatch,
             state_mutability,
             self,
             self.contract.body,
@@ -49,7 +50,7 @@ impl<'source_unit, 'context> ContractScope<'source_unit, 'context> {
             parameters,
             results,
         } = signature.function_type;
-        self.function(entry, results, |scope| {
+        self.function(entry, dispatch, results, |scope| {
             for (index, parameter) in function.parameters().iter().enumerate() {
                 let Some(identifier) = parameter.name() else {
                     continue;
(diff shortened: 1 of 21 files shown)


solx-slang/src/contract/mod.rs
```diff
@@ -46,26 +46,45 @@ impl<'source_unit, 'context> ContractScope<'source_unit, 'context> {
...
+                    let element_type = self.source_unit.resolve(
+                        &state_variable
+                            .get_type()
+                            .expect("binder types every state variable"),
+                        None,
+                    );
+                    self.contract.declare_immutable(
+                        &SourceUnitScope::state_variable_symbol(state_variable),
+                        element_type,
+                        self,
```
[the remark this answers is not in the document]
hedgar2017: The erase-when-absent behavior is the documented design, and the fallback path cannot miss errors through it: `mlirEvmLowerSetImmutables` has no failure path — both arms (store-and-erase, erase-only) always succeed and emit no diagnostics — and the constructor's computation of the stored value is fully retained; only the terminal store is dropped, in a diagnostic-only artifact whose compilation is already reporting the runtime failure. Erasure is also what lets the deploy fallback compile at all: an unlowered `llvm.setimmutable` fails LLVM translation, which is the problem the pass exists to solve (its own comment in `SolImmutables.cpp`: an immutable the runtime never loads has no reserved offsets and its set is dead). The suggested analogy does not hold either: the non-MLIR pipeline allocates no offsets — its `build.immutables` is the `immutableReferences` output metadata on the runtime object (the deploy arm passes `None`), with offsets embedded by solc's own codegen. Allocating fabricated offsets in solx-core would add a second offset authority beside the LLD extraction for no diagnostic gain.

ggiraldez (approved):
Looks good!

solx-slang/src/contract/function/mod.rs
```diff
@@ -37,10 +37,11 @@ impl<'source_unit, 'context> ContractScope<'source_unit, 'context> {
...
         let signature = self.source_unit.function_signature(function);
+        let dispatch = FunctionDispatch::from(function);
         let state_mutability = StateMutability::from(function.attributes().mutability());
         let entry = signature.define(
             selector,
-            FunctionDispatch::from(function),
+            dispatch,
             state_mutability,
             self,
             self.contract.body,
```
PavelKopyl, line 47: Probably is reasonable to introduce a structure describing  sol.func declaration, like:
```
pub struct FunctionHeader {
    pub selector: Option<u32>,
    pub dispatch: FunctionDispatch,
    pub state_mutability: StateMutability,
}

and a structure describing a result of creation (the entry block) with the header it was declared with:

pub struct FunctionEntry<'context> {
    pub block: Block<'context>,
    pub header: FunctionHeader,
}
```
This will reduce the number of argument in several call states and slightly improve readability. I guess with this we will not need `#![allow]... `

hedgar2017: Thanks, nice refactor, pushed!

PR #653, opened 2026-08-16
feat(mlir): enable the spill-retry protocol for the MLIR pipeline

abinavpp:
https://github.com/NomicFoundation/solx-llvm/pull/142

### DIFF
--- a/solx-codegen-evm/src/codegen/context/mod.rs
+++ b/solx-codegen-evm/src/codegen/context/mod.rs
@@ -178,11 +178,9 @@ impl<'ctx> Context<'ctx> {
             self.optimizer.settings(),
         );
         let spill_area_size = self.optimizer.settings().spill_area_size();
-        let target_machine = TargetMachine::new(
-            self.optimizer.settings(),
-            self.llvm_options.as_slice(),
-            spill_area_size.map(|size| (self.memory_guard, size)),
-        )?;
+        self.describe_stack_region(spill_area_size.unwrap_or_default());
+        let target_machine =
+            TargetMachine::new(self.optimizer.settings(), self.llvm_options.as_slice())?;
         target_machine.set_target_data(self.module());
         target_machine.set_asm_verbosity(true);
 
@@ -419,6 +417,51 @@ impl<'ctx> Context<'ctx> {
     ///
     pub fn set_memory_guard(&mut self, offset: u64) {
         self.memory_guard = offset;
+    }
+
+    ///
+    /// Sets the spill region module flags read by the EVM backend.
+    ///
+    fn describe_stack_region(&self, size: u64) {
+        if self.module().get_flag("evm-memory-guard").is_none() {
+            self.set_module_flag("evm-memory-guard", self.memory_guard);
+        }
+        self.set_module_flag("evm-stack-region-size", size);
+    }
+
+    ///
+    /// Whether the `llvm.module.flags` triplet `flag` is named `key`.
+    ///
+    fn is_flag_named(flag: &inkwell::values::MetadataValue<'ctx>, key: &str) -> bool {
+        flag.get_node_values().get(1).is_some_and(|name| {
+            name.into_metadata_value()
+                .get_string_value()
+                .is_some_and(|name| name.to_bytes() == key.as_bytes())
+        })
+    }
+
+    ///
+    /// Adds the `key` module flag, or replaces its value if it is already there.
+    ///
+    fn set_module_flag(&self, key: &str, value: u64) {
+        use inkwell::values::AsValueRef;
+
+        let value = self.llvm().i64_type().const_int(value, false);
+        for flag in self.module().get_global_metadata("llvm.module.flags") {
+            if !Self::is_flag_named(&flag, key) {
+                continue;
+            }
+            unsafe {
+                inkwell::llvm_sys::core::LLVMReplaceMDNodeOperandWith(
+                    flag.as_value_ref(),
+                    2,
+                    inkwell::llvm_sys::core::LLVMValueAsMetadata(value.as_value_ref()),
+                );
+            }
+            return;
+        }
+        self.module()
+            .add_basic_value_flag(key, inkwell::module::FlagBehavior::Error, value);
     }
 
     ///
--- a/solx-core/src/project/contract/mod.rs
+++ b/solx-core/src/project/contract/mod.rs
@@ -533,9 +533,6 @@ impl Contract {
                 let context = unsafe { inkwell::context::Context::new(raw_llvm.context) };
                 let module = unsafe { inkwell::module::Module::new(raw_llvm.module) };
                 module.set_name(code_identifier.as_str());
-                if let Some(size) = optimizer.settings().spill_area_size() {
-                    Self::set_spill_area_module_flag(&context, &module, size)?;
-                }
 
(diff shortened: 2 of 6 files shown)


solx-mlir/sol_attr_stubs.cpp
```diff
@@ -107,7 +107,9 @@ MlirType solxCreateArrayType(MlirContext ctx, int64_t size, MlirType elementType
     if (dataLocation > 5) abort();
     auto *context = unwrap(ctx);
     auto location = static_cast<mlir::sol::DataLocation>(dataLocation);
-    return wrap(mlir::sol::ArrayType::get(context, size, unwrap(elementType), location));
+    std::optional<llvm::APInt> arraySize;
+    if (size >= 0) arraySize = llvm::APInt(256, static_cast<uint64_t>(size));
+    return wrap(mlir::sol::ArrayType::get(context, arraySize, unwrap(elementType), location));
```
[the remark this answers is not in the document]
abinavpp: Added the `size < -1` guard.

solx-mlir/src/context/mod.rs
```diff
@@ -147,7 +147,9 @@ impl<'context> Context<'context> {
     /// 1. `canonicalize`
     /// 2. `sol-inline-modifiers`
     /// 3. `convert-sol-to-yul`: Sol → Yul
-    /// 4. `convert-yul-to-std`: Yul → func/arith/scf/cf/LLVM
+    /// 4. `convert-yul-to-std`: Yul → func/arith/scf/cf/LLVM, keeping the
+    ///    memoryguard symbolic so [`Self::translate_source_to_llvm`] can fold
+    ///    it per spill-retry pass
```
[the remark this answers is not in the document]
abinavpp: Updated the comment to point at the driver-side fold.

hedgar2017 (changes requested):
Thanks! Looks good, but could you run `cargo run-tester-slang` and post the test tally? AFAIU, it should have unblocked some stack-too-deep tests?

hedgar2017 (changes requested):
LGTM, but please get merged LLVM first.

solx-core/src/project/contract/mod.rs
```diff
@@ -610,4 +613,45 @@ impl Contract {
...
+    #[cfg(feature = "mlir")]
+    fn set_spill_area_module_flag(
+        context: &inkwell::context::Context,
+        module: &inkwell::module::Module,
+        size: u64,
+    ) -> anyhow::Result<()> {
+        use inkwell::values::AsValueRef;
+
+        let size = context.i64_type().const_int(size, false);
+        for flag in module.get_global_metadata("llvm.module.flags") {
```
vladimirradosavljevic: Can we use `module.get_flag` instead?

solx-codegen-evm/src/codegen/context/mod.rs
```diff
@@ -421,6 +419,51 @@ impl<'ctx> Context<'ctx> {
...
+        if self.module().get_flag("evm-memory-guard").is_none() {
+            self.set_module_flag("evm-memory-guard", self.memory_guard);
+        }
+        self.set_module_flag("evm-stack-region-size", size);
+    }
+
+    ///
+    /// Whether the `llvm.module.flags` triplet `flag` is named `key`.
+    ///
+    fn is_flag_named(flag: &inkwell::values::MetadataValue<'ctx>, key: &str) -> bool {
```
hedgar2017, line 435: Single call: please inline.

solx-codegen-evm/src/codegen/context/mod.rs
```diff
@@ -421,6 +419,51 @@ impl<'ctx> Context<'ctx> {
...
+        flag.get_node_values().get(1).is_some_and(|name| {
```
hedgar2017, line 436: Worth explaining why is this index == 1, using a named constant, and/or using a stricter / more typed accessor.

solx-codegen-evm/src/codegen/context/mod.rs
```diff
@@ -421,6 +419,51 @@ impl<'ctx> Context<'ctx> {
         self.memory_guard = offset;
     }
 
+    ///
+    /// Sets the spill region module flags read by the EVM backend.
+    ///
+    fn describe_stack_region(&self, size: u64) {
+        if self.module().get_flag("evm-memory-guard").is_none() {
```
hedgar2017, line 426: Is this guard needed, or is the setter idempotent?
If it's needed, worth spelling the magic literal once, or even refactoring these keys by moving them to some enum as their number grows.

solx-codegen-evm/src/codegen/context/mod.rs
```diff
@@ -421,6 +419,51 @@ impl<'ctx> Context<'ctx> {
...
+            name.into_metadata_value()
+                .get_string_value()
+                .is_some_and(|name| name.to_bytes() == key.as_bytes())
+        })
+    }
+
+    ///
+    /// Adds the `key` module flag, or replaces its value if it is already there.
+    ///
+    fn set_module_flag(&self, key: &str, value: u64) {
```
hedgar2017, line 446: This is more low-level so belongs in inkwell. Maybe it already contains some flag setters in the module.

hedgar2017 (changes requested):

solx-codegen-evm/src/codegen/context/mod.rs
```diff
@@ -421,6 +419,51 @@ impl<'ctx> Context<'ctx> {
         self.memory_guard = offset;
     }
 
+    ///
+    /// Sets the spill region module flags read by the EVM backend.
+    ///
+    fn describe_stack_region(&self, size: u64) {
```
hedgar2017, line 425: I think can also be inlined once `set_module_flag` is moved to inkwell.

hedgar2017 (changes requested):

solx-codegen-evm/src/codegen/context/mod.rs
vladimirradosavljevic, line 426: In which cases this won't be set? I assume for pure Yul, but in that case, we don't want to add these flags, right? I would say, if there are no flags, then don't set them here, otherwise do the update for 2nd run.
Also, how are we handling case when there is unsafe assembly?

solx-codegen-evm/src/codegen/context/mod.rs
abinavpp: Moved it here: https://github.com/NomicFoundation/inkwell/pull/5. Done.

solx-codegen-evm/src/codegen/context/mod.rs
abinavpp: done

solx-codegen-evm/src/codegen/context/mod.rs
abinavpp: done

solx-codegen-evm/src/codegen/context/mod.rs
abinavpp: done

solx-codegen-evm/src/codegen/context/mod.rs
abinavpp: The setter moved to inkwell. The check stays so we don't override the guard MLIR already wrote.

PR #661, opened 2026-08-18
chore: bump solx-llvm and solx-solidity

hedgar2017:
Updates the lit expectations to the output the newer solc produces, and adapts the array type stub and the parenthesized callee to match it.

### DIFF
--- a/solx-slang/src/contract/function/expression/call/mod.rs
+++ b/solx-slang/src/contract/function/expression/call/mod.rs
@@ -363,18 +363,14 @@ impl Call {
                     options = Some(decorated.options());
                     callee = decorated.operand();
                 }
-                Expression::TupleExpression(inner) => match inner
-                    .items()
-                    .iter()
-                    .next()
-                    .and_then(|item| item.expression())
-                    .expect("a parenthesized callee wraps a single operand")
-                {
-                    Expression::Identifier(_) | Expression::MemberAccessExpression(_) => {
-                        return (Expression::TupleExpression(inner), options);
-                    }
-                    wrapped => callee = wrapped,
-                },
+                Expression::TupleExpression(inner) => {
+                    callee = inner
+                        .items()
+                        .iter()
+                        .next()
+                        .and_then(|item| item.expression())
+                        .expect("a parenthesized callee wraps a single operand");
+                }
                 resolved => return (resolved, options),
             }
         }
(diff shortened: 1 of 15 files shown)


solx-slang/src/contract/function/expression/call/mod.rs
```diff
@@ -363,18 +363,14 @@ impl Call {
                     options = Some(decorated.options());
                     callee = decorated.operand();
                 }
```
PavelKopyl, line 365: Just a note not related to the change itself, but which is reasonable to address.
It looks like the `options = Some(decorated.options());` will overwrite all the options, but the the first one:

```
interface ISink {
    function absorb(uint256 x) external payable returns (uint256);
}

contract C {
    function chained(ISink sink) public payable returns (uint256) {
        return sink.absorb{value: 1}{gas: 800000}(42);
    }

    function bare(address target) public payable returns (bool) {
        (bool ok, ) = target.call{value: 1}{gas: 800000}("");
        return ok;
    }
}
```
On the MLIR level we loose `{gas: 800000}` option at both calls.
Actually, this code is invalid: solc rejects it with: "Function call options have already been set, you have to combine them into a single {...}-option", whereas Slang accepts it.

I suggest to change the option extraction assuming there's only one {...} clause:
```
Expression::CallOptionsExpression(decorated) => {
    if options.is_some() {
        unimplemented!("chained call options: combine them into a single {{...}}-option");
    }
    options = Some(decorated.options());
    callee = decorated.operand();
}
```

PavelKopyl (approved):
LGTM, thanks!

It looks reasonable to have index_access.sol also as a runtime test (if we do not have test where a mapping is declared with one type, but is accessed with another one).

solx-slang/src/contract/function/expression/call/mod.rs
PavelKopyl: Or at least comment that we expect here only one {...} and that we relay on Slang's semantic analysis that should reject invalid code.

hedgar2017: We're not leaving such comments, just because our whole codegen relies on this kind of analysis by default. But let me check this case first.

hedgar2017: Confirmed, pure Slang validator gap. Thanks for the heads-up anyway!

PR #662, opened 2026-08-18
feat: parse EVM legacy assemblies emitted as pre-serialized strings

nebasuke:
Companion PR to https://github.com/NomicFoundation/solx-solidity/pull/174

# Claude Summary
## What

Companion to NomicFoundation/solx-solidity#174 — together they cut the legacy-pipeline main-process peak RSS on Graph Horizon from **9.56 GB to 3.94 GB** (and wall time from ~48 s to ~33 s), with byte-identical output. Three commits:

1. `refactor(core)`: translate EVM legacy assemblies without redundant copies — `try_from_solidity_output` moves each parsed assembly into translation by value (cloning only when `evm.legacyAssembly` is selected in the output), and `Assembly::take_runtime_code` replaces the clone-then-strip pair in `from_contract`. Output-invariant cleanup; on its own it does not move the RSS peak (which sits inside `solidity_compile`).
2. `feat(core)`: parse `evm.legacyAssembly` arriving as a pre-serialized JSON string. This is the consumer side of the fork-side fix: libsolc stops accumulating every contract's assembly JSON tree in its response object and hands solx compact strings instead. `materialize` unwraps the string form and still accepts the object form.
3. `chore`: pin the `solx-solidity` submodule to the tip of NomicFoundation/solx-solidity#174. **To be re-pinned to the merged `0.8.34` commit before this lands.**

## Why

Compiling Graph Horizon (240 sources, 91 Foundry test units) in legacy mode peaked at ~9.8 GB main-process RSS — first observed in the hardhat solx compile benchmarks (NomicFoundation/hardhat#8521). Per-phase RSS instrumentation showed the peak lives entirely inside the linked libsolc while it builds the standard JSON response: one nlohmann tree per contract, each recursively embedding full copies of its dependencies' assemblies, all alive until the single final dump. Details, timelines, and the vanilla-solc control runs are in solx-solidity#174.

| Horizon replay | before | after |
|---|---|---|
| legacy dump | 9.56 GB / ~48 s | **3.94 GB / ~33 s** |
| via-IR dump | 2.43 GB | 2.44 GB |

## Validation

- **Output invariance**: while the fork change was still behind a setting, a flag-toggle control (same binaries, off vs on) produced `cmp`-byte-identical standard JSON responses on both the legacy and via-IR Horizon replays — full response compared, not sampled. The final unconditional revision replays with zero structural differences against that build; the only changed bytes are CBOR metadata-hash tails embedding the fork commit hash, which change on any submodule bump by construction.
- `cargo test`: 538 passed, 0 failed, against the pinned libsolc.
- The CLI standard-JSON test with `select_all.json` covers the user-facing `evm.legacyAssembly` output (still emitted as an object — the string form is purely internal solc↔solx transport); solx-tester's EVMLA modes exercise the whole legacy pipeline.

## Notes for review

- Fork-side memory numbers, the `null`-preservation subtlety, and follow-up candidates are documented in solx-solidity#174.

### DIFF
--- a/solx-standard-json/src/output/contract/evm/legacy_assembly.rs
+++ b/solx-standard-json/src/output/contract/evm/legacy_assembly.rs
@@ -22,11 +22,20 @@ impl LegacyAssembly {
     ///
     /// Parses the [`LegacyAssembly::Raw`] variant into [`LegacyAssembly::Parsed`] in place.
     ///
+    /// `solc` emits the assembly as a pre-serialized JSON string, so the raw value is unwrapped
+    /// before the assembly itself is parsed; the plain object form is accepted as well.
+    ///
     pub fn materialize(&mut self) -> anyhow::Result<()> {
         if let Self::Raw(raw) = self {
-            *self = Self::Parsed(solx_utils::deserialize_from_slice::<
-                solx_evm_assembly::Assembly,
-            >(raw.get().as_bytes())?);
+            let assembly = if raw.get().starts_with('"') {
+                let json = solx_utils::deserialize_from_slice::<String>(raw.get().as_bytes())?;
+                solx_utils::deserialize_from_str::<solx_evm_assembly::Assembly>(json.as_str())?
+            } else {
+                solx_utils::deserialize_from_slice::<solx_evm_assembly::Assembly>(
+                    raw.get().as_bytes(),
+                )?
+            };
+            *self = Self::Parsed(assembly);
         }
         Ok(())
     }
--- a/solx-standard-json/src/output/contract/evm/legacy_assembly.rs
+++ b/solx-standard-json/src/output/contract/evm/legacy_assembly.rs
@@ -23,18 +23,13 @@ impl LegacyAssembly {
     /// Parses the [`LegacyAssembly::Raw`] variant into [`LegacyAssembly::Parsed`] in place.
     ///
     /// `solc` emits the assembly as a pre-serialized JSON string, so the raw value is unwrapped
-    /// before the assembly itself is parsed; the plain object form is accepted as well.
+    /// before the assembly itself is parsed.
     ///
     pub fn materialize(&mut self) -> anyhow::Result<()> {
         if let Self::Raw(raw) = self {
-            let assembly = if raw.get().starts_with('"') {
-                let json = solx_utils::deserialize_from_slice::<String>(raw.get().as_bytes())?;
-                solx_utils::deserialize_from_str::<solx_evm_assembly::Assembly>(json.as_str())?
-            } else {
-                solx_utils::deserialize_from_slice::<solx_evm_assembly::Assembly>(
-                    raw.get().as_bytes(),
-                )?
-            };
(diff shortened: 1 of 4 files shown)


solx-standard-json/src/output/contract/evm/legacy_assembly.rs
```diff
@@ -22,11 +22,20 @@ impl LegacyAssembly {
...
+    /// before the assembly itself is parsed; the plain object form is accepted as well.
+    ///
     pub fn materialize(&mut self) -> anyhow::Result<()> {
         if let Self::Raw(raw) = self {
-            *self = Self::Parsed(solx_utils::deserialize_from_slice::<
-                solx_evm_assembly::Assembly,
-            >(raw.get().as_bytes())?);
+            let assembly = if raw.get().starts_with('"') {
+                let json = solx_utils::deserialize_from_slice::<String>(raw.get().as_bytes())?;
+                solx_utils::deserialize_from_str::<solx_evm_assembly::Assembly>(json.as_str())?
```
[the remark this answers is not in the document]
nebasuke: The gitlink update is part of this PR: the changed-file set includes `solx-solidity` (commit b8ea2851, `ebeac7c25` → `792f20090`), which pins the submodule to the string-emitting tip of NomicFoundation/solx-solidity#174. The pending step called out in the description is only the *re*-pin to the merged `0.8.34` commit once #174 lands.

solx-standard-json/src/output/contract/evm/legacy_assembly.rs
```diff
@@ -22,11 +22,20 @@ impl LegacyAssembly {
...
     ///
+    /// `solc` emits the assembly as a pre-serialized JSON string, so the raw value is unwrapped
+    /// before the assembly itself is parsed; the plain object form is accepted as well.
+    ///
     pub fn materialize(&mut self) -> anyhow::Result<()> {
         if let Self::Raw(raw) = self {
-            *self = Self::Parsed(solx_utils::deserialize_from_slice::<
-                solx_evm_assembly::Assembly,
-            >(raw.get().as_bytes())?);
+            let assembly = if raw.get().starts_with('"') {
```
hedgar2017: This branch must be redundant so let's remove it to make sure we fail loud on unexpected input.

hedgar2017 (changes requested):
Otherwise LGTM!

solx-standard-json/src/output/contract/evm/legacy_assembly.rs
nebasuke: Good idea.

Removed in a2a16d07. `materialize` now unconditionally unwraps the string form, so an object-form assembly aborts compilation with `invalid type: map, expected a string`. Verified both ways: the CLI standard JSON tests fail with exactly that error against a stale `libsolc` built before the string emission, and pass against the pinned `0.8.34` commit.

hedgar2017 (approved):

hedgar2017 (approved):

PR #666, opened 2026-08-19
fix: report worker fatal errors per contract instead of dropping the contract

nebasuke:
12 of the Sourcify corpus were discovered (https://github.com/NomicFoundation/hardhat/actions/runs/32248335986/job/96067781572?pr=8538) to drop a contract due to a recursive function hitting a stack too deep, which would kill the worker, thus silently not producing a compiled contract. 

Added a generic fix for LLVM crashing with a fatal error, so these get reported properly upstream.

## Claude Summary

A worker child dying on an LLVM fatal error — e.g. failed stackification of a recursive function with stack-too-deep — previously left the affected contract **silently missing from the standard JSON output with exit code 0** and no `errors` entry, while the CLI path panicked on the same condition (`unwrap_standard_json_ref`). Downstream consumers (Hardhat, Foundry) recorded a successful build with a missing artifact.

Found by the Sourcify 0.8.34 corpus sweep (NomicFoundation/hardhat#8538) in `--compare` mode: 12 of 112,587 contracts (4 distinct projects, all with recursive functions hitting stack-too-deep) lost their deployed target contract this way while every other contract in the job was emitted normally. Reproduction: the LLVM fatal kills the codegen child; the parent maps the dead worker to `Error::Generic` (`worker.rs`), and `write_to_standard_json` only forwarded the `Error::StandardJson` variant into `errors`, dropping everything else with a `continue`.

## Changes

- The worker child installs an LLVM fatal error handler (via `install_fatal_error_handler`, already exposed by our inkwell fork) alongside the existing stack-too-deep handler: it relays the fatal message over the reply channel as a generic error and exits, so the parent receives the diagnostic instead of only observing a dead process. This also converts other backend fatals — e.g. `cannot evaluate undefined symbol` from the linker — from crash-style aborts with LLVM stack dumps into clean error reports.
- `write_to_standard_json` wraps every non-`StandardJson` per-object error into a standard JSON error naming the contract (same pattern the child already applies to in-process generic errors) instead of dropping it.
- The CLI error path stringifies all error variants instead of panicking on non-`StandardJson` ones; `unwrap_standard_json_ref` is removed together with its last caller.
- Pool doc comments updated: a worker is now also retired by an LLVM-fatal reply.

## Testing

- New fixture `RecursiveStackTooDeep.sol` (18 values live across a recursive call) with two tests:
  - `stack_too_deep_recursive_standard_json` — **fails on the previous code** (no error entry, contract silently missing) and passes now with the error naming the contract and carrying the "refactor to a non-recursive approach" guidance;
  - `stack_too_deep_recursive` — CLI mode exits nonzero with the same message.
- `cli::llvm_ir::linker_error` expectation updated: the linker fatal is now reported as `Error: LLVM error: ...` instead of a raw `LLVM ERROR:` crash dump.
- Full CLI suite green (409 tests), clippy clean.
- Verified end to end against the sweep repro inputs: the previously dropped contracts now produce a standard JSON `error` with the contract `sourceLocation`, and stderr stays free of LLVM crash diagnostics.

## Follow-up

Once released, the sweep will classify these as expected rejections (`solx-unsupported`) by matching the new diagnostic, instead of surfacing them as output mismatches. A structured error for the recursion case (instead of riding the LLVM message string) is possible later polish in solx-llvm, not required for correctness.

### DIFF
--- a/solx-core/src/build/mod.rs
+++ b/solx-core/src/build/mod.rs
@@ -16,8 +16,6 @@ use solx_utils::SyncLock;
 
 use solx_standard_json::CollectableError;
 
-use crate::error::Error;
-
 use self::contract::Contract;
 use self::contract::object::Object as ContractObject;
 
@@ -333,21 +331,14 @@ impl Build {
                         contract.deploy_object_result.take(),
                         contract.runtime_object_result.take(),
                     ] {
-                        match result {
-                            Some(Err(Error::StandardJson(error))) => {
-                                standard_json.errors.push(error);
-                            }
-                            // Worker deaths (e.g. LLVM fatal errors) surface as
-                            // generic errors; dropping them would emit a
-                            // successful output with the contract silently
-                            // missing.
-                            Some(Err(error)) => standard_json.errors.push(
-                                solx_standard_json::OutputError::new_error_contract(
-                                    Some(contract.name.path.as_str()),
-                                    error,
-                                ),
-                            ),
-                            Some(Ok(_)) | None => {}
+                        // Worker deaths (e.g. LLVM fatal errors) surface as
+                        // generic errors; dropping them would emit a
+                        // successful output with the contract silently
+                        // missing.
+                        if let Some(Err(error)) = result {
+                            standard_json
+                                .errors
+                                .push(error.into_standard_json(Some(contract.name.path.as_str())));
                         }
                     }
                     continue;
--- a/solx-core/src/build/mod.rs
+++ b/solx-core/src/build/mod.rs
@@ -329,15 +329,26 @@ impl Build {
                         .unwrap_or_default(),
                 );
                 if deploy_object_result.is_err() || runtime_object_result.is_err() {
-                    if let Some(Err(Error::StandardJson(error))) =
-                        contract.deploy_object_result.take()
-                    {
-                        standard_json.errors.push(error);
-                    }
-                    if let Some(Err(Error::StandardJson(error))) =
-                        contract.runtime_object_result.take()
-                    {
-                        standard_json.errors.push(error);
+                    for result in [
+                        contract.deploy_object_result.take(),
+                        contract.runtime_object_result.take(),
+                    ] {
+                        match result {
+                            Some(Err(Error::StandardJson(error))) => {
+                                standard_json.errors.push(error);
+                            }
+                            // Worker deaths (e.g. LLVM fatal errors) surface as
+                            // generic errors; dropping them would emit a
+                            // successful output with the contract silently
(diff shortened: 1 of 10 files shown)


solx-core/src/process/child.rs
```diff
@@ -26,6 +26,11 @@ pub fn run() -> anyhow::Result<()> {
                 .ok_or_else(|| anyhow::anyhow!("The worker received no session"))?;
 
             inkwell::support::error_handling::install_stack_error_handler(evm_stack_error_handler);
+            unsafe {
```
hedgar2017: Can this `unsafe` be moved to declaration to be consistent with the other handler?

solx-core/src/build/mod.rs
```diff
@@ -329,15 +329,26 @@ impl Build {
...
+                    ] {
+                        match result {
+                            Some(Err(Error::StandardJson(error))) => {
+                                standard_json.errors.push(error);
+                            }
+                            // Worker deaths (e.g. LLVM fatal errors) surface as
+                            // generic errors; dropping them would emit a
+                            // successful output with the contract silently
+                            // missing.
+                            Some(Err(error)) => standard_json.errors.push(
```
hedgar2017: Can we keep the removed `unwrap_standard_json_ref` but rename it and make it always return?

hedgar2017 (changes requested):

solx-core/src/process/child.rs
nebasuke: Done. The `unsafe` came from the upstream inkwell signature — unlike our fork's `install_stack_error_handler`, which already wraps the FFI call internally. NomicFoundation/inkwell#4 makes `install_fatal_error_handler` safe the same way; c1b130bd bumps the pin and drops the `unsafe` block here. Once #4 is merged I'll re-pin to the final SHA.

solx-core/src/build/mod.rs
nebasuke: Done — reinstated as `Error::into_standard_json(self, path: Option<&str>)`: returns the inner error for `StandardJson` and wraps the other variants into a contract-attributed standard JSON error instead of panicking. The match block here collapses to a single call. The other former call site (`error_strings`) works on `&Error` and only needs the message, so it uses `Display` directly.

solx-core/src/process/child.rs
hedgar2017: Ah, I actually just meant moving it to the function declaration down here, not to inkwell. Let me check if it's possible.

hedgar2017 (approved):

PR #677, opened 2026-08-30
ci: port main's CI and dependency changes onto main-slang

nebasuke:
Ports main's CI and dependency changes onto `main-slang`, which has been diverging since `a85ef238` (2026-08-08). Two of them fix CI that is currently red on **every** PR in the main-slang stack.

Cherry-picks only, `-x` so each commit records the main SHA it came from. No feature work, no submodule change.

## Why now

`main-slang` is failing two gates on every PR, neither caused by anything in this repo:

- **zizmor** — `re-actors/alls-green` moved its rolling `release/v1` branch onto v1.3.0 on 2026-08-21, so our SHA pin's `# release/v1` comment stopped resolving to the pinned commit and `ref-version-mismatch` began failing. The SHA never moved and is what actually runs; only the comment went stale. Fixed by naming the immutable tag (`# v1.2.2`) instead of a branch that keeps moving.
- **cargo-deny** — `chacha20` 0.10.0/0.10.1 were yanked on 2026-08-27 (SSE4.1 intrinsic on an SSE2-guarded path). 0.10.2 fixes it but was inside our 7-day cooldown, so this holds the yanked version with a documented, time-boxed ignore rather than allowlisting a three-day-old crate past the gate. The defect is unreachable from solx — see the `deny.toml` comment.

Also picks up the Foundry pin from #673. That one matters here: the digest-pinned `foundry-toolchain` action still defaults its `version:` input to `stable`, so without it the foundry leg floats onto 1.8.x and hits the lint-on-build ICE.

## Included

| commit | from |
|---|---|
| ubuntu:26.04 digest | #655 |
| pnpm 11.21.0 | #658 |
| github-actions SHA bumps | #657 |
| rust_xlsxwriter 0.97 | #660 |
| object 0.40 | #659 |
| cargo minor/patch | #656 |
| arrayref 0.3.9 hold | #669 |
| cc 1.4.4 + the two CI fixes above | #672 |
| pnpm 11.22.0 + Foundry v1.7.1 pin | #673 |

#669 and #672 both land even though the second removes the first's arrayref block: crates.io reverted those yanks and deleted the compromised 0.3.10 outright, so the hold is dead config. Taking only #672 would conflict against a block that never existed here, and this keeps the history honest.

## Deliberately not included

`#662` (touches the `solx-solidity` submodule), `#664`/`#665` (BLOBHASH, SELFDESTRUCT), `#663`/`#666` (source behaviour), `#668` (0.1.8 version bump — a release chore).

A real `main` → `main-slang` merge is still blocked on the submodule: the `solx-solidity` pointers genuinely diverge, with main-slang's commit 1014 ahead of main's and main's 6 ahead of main-slang's. That needs its own merge in the fork before the feature commits can follow.

## Two conflicts worth reviewing

**`test.yaml` on #657.** main-slang inserts a MinGW bindgen step where main has its `rust-cache` step, so git offered main's version of that step wholesale — including `prefix-key: build-and-test-v2` and `save-if: refs/heads/main`. Accepting that would have silently repointed main-slang's cache keys and stopped it saving on its own branch. Resolved by keeping main-slang's step and bumping the SHA in place on both of its cache steps, preserving `cargo-checks-v1` / `slang-tests-v2` and `refs/heads/main-slang`. Verified no stale #657 SHAs remain anywhere under `.github`.

**`Cargo.lock` on #660.** Not hand-merged. Took main-slang's lock and ran `cargo update -p rust_xlsxwriter --precise 0.97.1`, which pulled `zip 7.2.0 → 8.6.0` exactly as main resolved it. Both are months outside the cooldown window.

## Verification

Run locally against this branch, using CI's exact invocations:

| check | result |
|---|---|
| `cargo deny --all-features check --allow unmaintained --hide-inclusion-graph` | `advisories ok, bans ok, licenses ok, sources ok`, exit 0 |
| `zizmor` 1.26.1, online, whole `.github` | no findings, exit 0 |
| `actionlint` with `SHELLCHECK_OPTS=--severity=warning` | exit 0 |
| `cargo metadata --locked` | resolves |

`renovate.json` and `solx-dev/hardhat-tests.toml` came out byte-identical to main. The Foundry pin was confirmed to land on the repo's only `foundry-toolchain` step, not merely to apply cleanly.

## Follow-up

Drop the `chacha20@0.10.1` entry and take 0.10.2 once its cooldown expires on **2026-09-03 17:51 UTC**. The `deny.toml` comment states that condition inline.

### DIFF
--- a/.github/actions/code-changes/action.yml
+++ b/.github/actions/code-changes/action.yml
@@ -17,7 +17,7 @@ runs:
     # 'every': a file counts as code only if no negation pattern excludes it.
     # Not every .md is documentation — solx-benchmark-converter compiles
     # templates/summary.md into the binary and asserts golden .md fixtures.
-    - uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
+    - uses: dorny/paths-filter@ceb8a2b8f2d89434be7ff52d3de7ec3738c5cc9d # v4.0.3
       id: filter
       with:
         token: ''
--- a/.github/docker/Dockerfile
+++ b/.github/docker/Dockerfile
@@ -15,7 +15,7 @@
 #          regression tests, cargo coverage/report tools). Must stay the final
 #          stage: docker-ci-image.yaml publishes the default target.
 
-FROM ubuntu:26.04@sha256:3131b4cc82a783df6c9df078f86e01819a13594b865c2cad47bd1bca2b7063bb AS base
+FROM ubuntu:26.04@sha256:678c6550cc43645e08669028bc177f50be4e7c5b8cca677067b1914d4afc7a03 AS base
 
 LABEL org.opencontainers.image.source="https://github.com/NomicFoundation/solx"
 LABEL org.opencontainers.image.description="CI runner image for solx"
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -79,7 +79,7 @@ jobs:
           optional: 'true'
 
       - name: Cache cargo artifacts
-        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+        uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
         with:
           prefix-key: coverage-v1
           cache-on-failure: true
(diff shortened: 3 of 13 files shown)


hedgar2017 (approved):

hedgar2017 (changes requested):
I'd say we need to do a full rebase. It should not have any conflicts.

PR #684, opened 2026-09-02
feat(slang): apply import remappings

nebasuke:
With the switch to the Slang front-end (away from solc), we need to actually implement import remapping support within the solx repo. 

Locally tested on Sourcify corpus of all contracts for 0.8.34 with all remappings now passing.

# Claude summary

The Slang frontend resolves imports as solc does (`CompilerStack::resolveImports`: `absolutePath`, then `ImportRemapper::apply`), against the sources it is given. `settings.remappings` is an order-preserving `Vec<Remapping>` parsed once at the settings boundary through the shared `solx_utils::Remapping`, which fixes the tie-break for both frontends; the CLI's `path=path` positionals feed it. Relative resolution is a literal port of boost 1.83's POSIX `parent_path`, so URL identifiers, leading `..`, and root-directory identifiers behave as under solc. Slang reports imports that resolve outside the source set.

Verified against solc 0.8.34 with the resolver driven directly: all 811,236 imports of the Sourcify 0.8.34 corpus (112,587 contracts, 49,874 remappings) resolve to the solc model's identifier; 60k fuzzed identifier/import pairs match a boost-linked `absolutePath`; 5,500 fuzzed remapping cases match solc's standard-JSON resolution. The `lib//` sanitize comment was answered inline: `sanitizePath` is the identity on POSIX.

Removed with respect to `main`: the parent-dir-join fallback for non-relative imports and the verbatim-import fallback, both of which accepted imports solc rejects.

PR #671, opened 2026-08-20
feat(slang): inline assembly

abinavpp:
Supports Solidity inline assembly through the Yul dialect.

PR #676, opened 2026-08-27
fix(slang): convert operands to the types their dialect ops admit

PavelKopyl:
This includes:
 - aligning 'sol.byte' with 'bytes1' at conversion boundaries
 - converting arguments for 'sol.ecrecover', 'sol.send', and 'sol.transfer'
 - folding constant identifiers at their declared type
 
Fixed semantic tests:
```
array/indexAccess/bytes_index_access.sol
functionCall/send_zero_ether.sol
isoltestTesting/balance_other_contract.sol
revertStrings/transfer.sol
```

PR #686, opened 2026-09-02
fix(slang): constants at their declared type, compound assignment on bytes elements

nebasuke:
Fixes bucket 2 of the Sourcify sweep (#685, run 1): 10,457 contracts failing with

```
error: 'sol.fixed_bytes_index' op operand #0 must be <fixed bytes>, but got '!sol.string<Memory>'
```

The sweep row called this `bytes memory` indexing; it is not. `bytes memory` indexing already works (`readBytes` in `index_access.sol`). The trigger is the OpenZeppelin `Strings` idiom:

```solidity
bytes16 private constant HEX_DIGITS = "0123456789abcdef";
buffer[i] = HEX_DIGITS[value & 0xf];
```

A constant folded to its initializer at the initializer's own type, so the string literal became a `!sol.string<Memory>` and `HEX_DIGITS[i]` reached `sol.fixed_bytes_index` with the wrong operand. The constant getter and the inline-assembly reference already fold through `converted` to the declared type; the identifier path now does the same (first commit).

9,907 of those contracts also reported

```
error: 'sol.xor' op requires the same type for all operands and results
```

from OpenZeppelin 5.1's `toChecksumHexString`, `buffer[i] ^= 0x20`. Compound assignment loaded the element as `!sol.byte` but converted the right-hand side to the binder's `bytes1`. It now operates at the binder's type and casts back for the store, as value-position indexing already does after its load (second commit). No record in the sweep had the xor error alone, so both fixes are needed to clear the bucket.

Two lit cases, one per bug, in `index_access.sol` and `compound_assignment.sol`. Both fail on the pre-fix build with exactly the two errors above and pass on this branch. The solc fork lowers `b[i] ^= 0x20` at `!sol.byte` where Slang uses `!sol.fixedbytes<1>` with casts around the operator; the file's existing `CHECK-SOLX`/`CHECK-SOLC` split records both.

Verified locally with an MLIR-enabled toolchain: `llvm-lit solx-mlir/tests/lit/` 134/134 on both frontends, `cargo test-slang` 201 passed. Corpus-level confirmation waits for sweep run 3 on the sweep branch: the OZ 5.x bundles also need inline assembly (#671) and the klaytn bundles remappings (#684), so they cannot compile on `main` alone.

### DIFF
--- a/solx-mlir/tests/lit/compound_assignment.sol
+++ b/solx-mlir/tests/lit/compound_assignment.sol
@@ -101,6 +101,18 @@
 // CHECK-SOLC:   sol.shr %[[OLD]], %[[SHR_AMOUNT]] : ui256, ui256
 // CHECK:   sol.store %{{.*}}, %[[XPTR]]
 
+// CHECK: sol.func @{{.*xor_assign_byte.*}}
+// CHECK-SOLX:   %[[RHS:.*]] = sol.bytes_cast %{{.*}} : ui8 to !sol.fixedbytes<1>
+// CHECK-SOLX:   %[[OLD:.*]] = sol.load %[[BPTR:.*]] : !sol.ptr<!sol.byte, Memory>, !sol.byte
+// CHECK-SOLX:   %[[OLD_BYTES:.*]] = sol.bytes_cast %[[OLD]] : !sol.byte to !sol.fixedbytes<1>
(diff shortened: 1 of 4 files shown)


hedgar2017 (changes requested):
Thanks!

Let me adjust it towards the repo conventions and update CLAUDE.md that everyone can follow.

nebasuke:
Closed in favour of #676 that incorporated any relevant work from this PR.

PR #688, opened 2026-09-03
feat: bump slang_solidity_v2 and update API

hedgar2017:
Bumps `slang_solidity_v2` from `8948967` to slang main `b34370a0` and ports the Slang frontend and the tester's `slang-ast` view to the push-based `CompilationUnit::create` API from NomicFoundation/slang#2068, which replaced `CompilationBuilder` and dropped the `MissingFile` diagnostic. Import resolution behaviour is unchanged; the frontend now borrows sources instead of cloning them, and the tester's resolver delegates the file-membership check to slang.


teofr (approved):
LGTM!

PR #689, opened 2026-09-03
ci: point trunk-branch references at main-solc

nebasuke:
* Rename any references in `main` to `main-solc`
* No point recording/running code coverage (and possibly conflicting with `main`), so workflow and config deleted
* Fixed up cache-warmup.yaml to allow for workflow_dispatch for `main-solc`. 

If you want caching for LLVM + solc, you would run `workflow_dispatch` with branch `main-solc` this would allow PRs on `main-solc` to use a cached build (unless bumping `solx-llvm` / `solx-solidity`).

# Claude summary - CI: rename `main` → `main-solc` in this branch's workflows

The rename left `main` references behind in `main-solc`'s workflows, in two shapes: hardcoded `ref: main` baselines, and triggers/cache guards keyed to `refs/heads/main`. Both break CI on this line, neither turns anything red.

Straight rename throughout — these files only ever run on `main-solc`, and the `main` arms are unreachable from here (a push to `main` runs `main`'s copy of the workflow; `github.ref` in these files can never be `refs/heads/main`).

## 1. The baseline built the wrong compiler

`integration-tests.yaml` and `compile-benchmark.yaml` checked out a literal `ref: main` as the comparison side. Since the rename that is the slang frontend, whose `solx-solidity` pin needs MLIR. On [#687's integration run](https://github.com/NomicFoundation/solx/actions/runs/33743716160):

```
CMake Error at CMakeLists.txt:134 (find_package):
  Could not find a package configuration file provided by "MLIR"
```

→ `Build solc (main)` exit 1 → `error: could not find native static library 'solc'` → `Build solx (main)` exit 101 → `"./target/release/solx" subprocess spawning: No such file or directory` in the baseline benchmark step.

Every baseline step is `continue-on-error: true` and the fail gate only inspects the PR-side steps, so the job went **green after 1h38m** with the summary comment reporting `⚪ not collected` / `⚠️ No baseline` for every suite. That degradation path is deliberate — a flaky baseline shouldn't lose healthy PR data — but it also means a permanently broken baseline is invisible.

Both now check out `main-solc`.

Step display names went `(main)` → `(main-solc)` so the logs read correctly.

## 2. Nothing had ever run *on* `main-solc`

`gh run list --branch main-solc` returns nothing. Every trunk trigger and cache guard was pinned to `main`:

| File | What | Why it's dead here |
|---|---|---|
| `test.yaml`, `cache-warmup.yaml` | `push.branches: [main]` | a push to `main` runs `main`'s copy |
| `test.yaml` ×2 | `save-if: github.ref == 'refs/heads/main'` | `github.ref` here is never `refs/heads/main` |

Every cache write in `build-llvm` and `build-solc` gates on `github.event_name != 'pull_request'`, so no PR can populate one either — leaving this line with no *automatic* way to fill a cache at all. The two lines also pin different submodules, so their keys never coincide:

```
main       solx-llvm 4907939b   solx-solidity 2cb35e4d
main-solc  solx-llvm 7d0702e1   solx-solidity 91fef221
```

Today's PRs are nonetheless *warm*, by an accident of history worth spelling out. The entries they hit were saved back when `main-solc` **was** `main`:

| key (abbreviated) | created | last accessed |
|---|---|---|
| `v1-llvm-Linux-X64-…-no-assertions-…-7d0702e1` | 2026-07-31 | 2026-09-04 |
| `v2-solc-Linux-X64-…-91fef221` | 2026-08-20 | 2026-09-04 |

Both sit in the `refs/heads/main` scope. The rename moved the branch, but a cache scope is just a ref string, so those pre-rename entries stay in the default-branch scope where every PR in the repo can read them — and main-solc PRs keep bumping `last_accessed_at`, holding them past the 7-day eviction window.

That is a wasting asset, not a working cache. Before this PR nothing could refresh it: the next `solx-llvm` or `solx-solidity` bump mints a new key, and with no push leg to populate it — and PR-scoped caches not shared between PRs — *every* PR after that bump would pay a full cold build. #687's `Tests` run building LLVM from scratch on Windows is that already happening on one platform.

The `push` leg added here closes that: a bump merged to `main-solc` matches the `paths` filter, so `cache-warmup` runs and saves into the `refs/heads/main-solc` scope, which main-solc pushes and PRs based on main-solc can both read. `workflow_dispatch --ref main-solc` remains as a manual escape hatch — it satisfies the same non-`pull_request` save condition, and with the concurrency group namespaced it can no longer be cancelled by `main`'s warm-up.

The residual risk is GitHub's 7-day inactivity eviction, which is independent of quota: a fortnight with no main-solc PR touching the borrowed entries drops them, and the first PR afterwards pays cold even with no bump involved. Quota-driven LRU is not a concern — the repo carries ~118 GiB of active caches, so its allowance is well above the 10 GB standard limit.

`cache-warmup`'s `concurrency.group` is the one place the two branches genuinely collide rather than merely going stale: groups are repo-wide, so the global constant `cache-warmup` plus `cancel-in-progress: true` means either line's warm-up kills the other's. Namespaced to `cache-warmup-main-solc`.

## 3. The report named the wrong branch

Every user-visible `main` in `solx-benchmark-converter` — the "PR vs `main`" heading, the failure and output verdicts, the no-baseline warnings — named a branch this line doesn't compare against. #689's own run said *"failures already present on `main`"* about `main-solc` runs. Renamed throughout, with the six golden fixtures and the inline assertion in `tests/summary.rs`.

**Toolchain identifiers deliberately stay put**: `00.solx-main` / `02.solx-main` in `toolchain_matrix.rs`, `[compilers.solx-main]` in the harness TOMLs, the `temp-solx-main` directory, `Role::Main`. Those are *role* names matched by exact string across three files, and the crate's own docs warn a rename must update every side or every summary renders a harness error. Renaming them changes report column labels for no behavioural gain — say the word if you want it and I'll do it as its own commit.

## 4. A broken baseline is no longer invisible

All the `main-solc`-side steps are `continue-on-error` by design — a flaky baseline must not redden an unrelated PR. But the only trace of a *permanently* broken one was `⚠️ No baseline` in the comment, indistinguishable from a suite that genuinely had nothing to compare, which is how #687 shipped a green check over a baseline that hadn't built for days.

The baseline steps now carry ids (`baseline_checkout`, `baseline_llvm`, `baseline_solc`, `baseline_solx`) and a new step turns any of them failing into a `::warning::` annotation plus a `$GITHUB_STEP_SUMMARY` callout — still without failing the job, so the documented best-effort behaviour is intact.

## 5. `docker-ci-image.yaml`

`branches: [main]` was unreachable here for the same reason as the others. Pointed at `main-solc`. `latest` and the date tag stay gated on `is_default_branch`, so runs from this line publish only the commit-sha tag.

## 6. Code coverage is deleted, not renamed

Once `main-solc` was in the picture the honest question was not which branch coverage should name, but whether it should run at all. It gated nothing:

- `codecov.yml` set `comment: false` and disabled both the `project` and `patch` statuses — no Codecov comment, no Codecov check.
- The workflow's own PR comment is computed from the PR's own lcov: absolute per-crate percentages, no delta, no baseline needed.
- It was never a required check — this branch's protection asks only for `PR Checks (Sanitizer)` and `PR Checks (Tests)`.
- Codecov holds no `main-solc` trunk data at all (it tracks 195 branches for this repo; `main-solc` isn't one), so even the report page the comment links to cannot render a diff.

For that it spent ~9 minutes of hosted runner per PR *warm* — and a cold `RelWithDebInfo+mlir+coverage` LLVM build is far worse. Not a trade worth making on the line whose frontend is being retired.

Deleting `coverage.yaml` takes `codecov.yml` with it (nothing else uploads) and `cache-warmup`'s `warm-llvm-coverage` job, which existed solely to prime the coverage config's LLVM artifact. The `coverage-v1` cargo cache was used by nothing else, so no other job goes colder for this.

The `enable-coverage` inputs on `build-llvm`, `build-rust` and `rust-unit-tests` stay. No workflow passes them now, but those composite actions are kept in step with `main`'s copies, and `solx-dev llvm build --enable-coverage` is still the documented local flow.

## Still needs a `main`-side change

Two items can't be fixed from this branch at all:

- **`cache-warmup`'s nightly cron** is dead here — scheduled runs only ever execute the default branch's copy of a workflow. `main-solc`'s caches are warmed by its own pushes alone. Giving the cron a ref matrix is a change to **`main`'s** copy. Noted in a comment.
- **`renovate.json`** has no `baseBranches`, and Renovate reads its config from the default branch only, so `main-solc` gets no dependency PRs. Needs `"baseBranches": ["main", "main-solc"]` on **`main`**.

## Also worth knowing

Retargeting a PR doesn't re-trigger anything. #687's timeline reads `head_ref_force_pushed` → `base_ref_changed` → `labeled`; only the workflows carrying `labeled` in their trigger types have a run on the current head. `Tests` and `Code coverage` last ran on the pre-force-push SHA, and `main-solc` requires `PR Checks (Tests)`, so retargeted PRs need a push or a manual re-run before they can merge.

## Verification

Everything CI lints that can run outside a full LLVM build, run at CI's own settings:

| Check | Result |
|---|---|
| `actionlint` + shellcheck, `SHELLCHECK_OPTS=--severity=warning` | clean (exit 0) |
| `zizmor 1.26.1 --format github .github` | `No findings to report` (exit 0) |
| `renovate-config-validator renovate.json` | `Config validated successfully` |
| `renovate --platform=local --dry-run=full` | 42 files / 364 deps, no errors; custom managers still 5 files |
| `cargo fmt --all -- --check` | clean (exit 0) |

Re-run after the coverage deletion: `actionlint` still exit 0, `zizmor` still `No findings to report` (suppressions dropped 47 → 44, matching the removed file).

The `renovate` dry-run matters because one custom manager reads `integration-tests.yaml`, the file I restructured most. Its matchString still extracts `{datasource: github-releases, depName: foundry-rs/foundry, currentValue: v1.7.1}` — the Foundry pin survived intact.

`cargo fmt` caught a genuine break: the lengthened `main-solc` string pushed the inline assertion in `tests/summary.rs:295` past `max_width`. Fixed by taking rustfmt's own wrapping. It was the only diff in the crate, so the newer rustfmt agrees with the pinned one everywhere else.

The baseline-checkout fix is confirmed end-to-end: this PR's integration run reported **bytecode size identical (54,924 comparisons), solx-tester gas identical (64,356), 0 new failures (685 pre-existing)** — against #687's `⚪ No output data` / `⚠️ No baseline`. That run predates the rewrite to a hardcoded `ref:`, but it checked out `${{ github.event.pull_request.base.ref }}`, which on this PR resolves to `main-solc` — the identical checkout, so the evidence carries over unchanged.

### What could not be verified locally

- `cargo clippy --frozen` and `cargo udeps` need the full workspace, i.e. a built LLVM and `libsolc`. Neither exists in my sandbox.
- **The converter's own unit tests never ran.** The pinned 1.97.1 toolchain there ships `cargo` and clippy but no `rustc`, so the checks above used the complete `stable`/1.98.0; and `cargo test -p solx-benchmark-converter` then can't resolve `slang_solidity_v2` — the cached git DB doesn't hold `main-solc`'s pinned rev `3b82850f`, `--offline` fails outright, and the online fetch stalls.

So the fixture assertions are verified **statically only**: every fixture line maps to a source string and both took the identical substitution, and the two strings without a fixture are asserted inline in `tests/summary.rs` (one prefix-only, one updated). CI's `cargo-checks` and `build-and-test` settle it — the fixtures are exact-match, so any missed occurrence fails loudly there rather than silently.


hedgar2017 (approved):
Thanks!

PR #690, opened 2026-09-04
fix(evm): describe the spill region with module flags

abinavpp:
Copy of https://github.com/NomicFoundation/solx/pull/653 for `main-solc`.


hedgar2017 (changes requested):
Please see #653.

PR #700, opened 2026-09-10
docs(slang): frontend design law and review skill

hedgar2017:
Records the Slang frontend's design law and adds a read-only review skill that reviews a change against it.

- **CLAUDE.md**: the Slang pipeline, its build and test commands, and the frontend's design law as numbered rules.
- **.claude/skills/review/SKILL.md**: four lenses, ambiguities as questions, findings as a report, posted on request.
- **README.md**: Slang named as the frontend, and the test suites split into what the tester runs and what is planned.

### DIFF
--- a/.claude/skills/review/SKILL.md
+++ b/.claude/skills/review/SKILL.md
@@ -1,0 +1,37 @@
+---
+name: review
+description: Read-only code review against CLAUDE.md. Ambiguities reach the user as questions, then every finding as a report. Nothing is changed. Use when the code builds and passes its tests.
+---
+
+# Review
+
+The skill changes nothing. It reads, checks, asks, and reports.
+
+## What is reviewed
+
+What the user names when invoking: a diff range, a branch, a path or a file. If nothing is named, the current branch against `main`. If the branch is `main`, the uncommitted changes. If there are none, ask the user with AskUserQuestion what to review.
+
+## Readers
+
+Launch a reader per lens, on Opus unless the invocation names another model. Each reader reads this repository's [`CLAUDE.md`](../../../CLAUDE.md) in full, then the whole change, and asks its question of every clause the change adds or alters.
+
+1. Removal: what fails if this clause is deleted? A clause nothing fails without is the finding.
+2. Trust: what does this clause check, derive or own that a lower layer already guarantees? The guarantee, read in that layer's code, is the evidence.
+3. Foreignness: where does the repository already do this another way? The sibling that does it is the evidence.
+4. Pinning: which claim of the change has no test that fails when the code is wrong? The mutation that leaves the test green is the evidence.
+
+The invocation may add lenses. Each added lens is its own question and its own reader.
+
+A reader runs nothing that writes, and reproduces a claim only on a scratch copy. It reports each finding with the file and line, what is wrong, the evidence, and the fix. A finding without evidence is not a finding.
+
+## Checking
+
+Every finding is re-read against the code before it goes further. One whose evidence does not hold is dropped. The same finding from several readers is reported once.
+
+## Decisions
+
+A decision is anything the review cannot settle from the code and `CLAUDE.md`: a finding with more than one defensible fix, a rule that reads two ways for this change, or a change whose intent the diff does not show. Each decision is put to the user with AskUserQuestion, one at a time, every option naming the evidence for it. The user's answer is final.
+
+## Report
+
+After the last decision is answered, every finding that survived checking, most consequential first. Each is a short paragraph: what is wrong, the evidence, what to do, and for a decided finding, the user's answer. Nothing is applied.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,137 +1,143 @@
-# CLAUDE.md
-
-This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
-
-## Project Overview
-
-**solx** is an optimizing Solidity compiler for EVM, developed by Matter Labs and Nomic Foundation. It uses a custom LLVM backend with an EVM target to generate optimized bytecode. The project is in beta.
+Human documentation starts at the [README](./README.md) and continues under [docs/src](./docs/src/). The [review skill](./.claude/skills/review/SKILL.md) reviews a change against this file.
 
 ## Architecture
 
-Three-repository structure via git submodules:
-- **solx** (this repo) — Rust workspace: CLI, compilation pipeline, LLVM IR generation
-- **solx-solidity** (submodule) — Fork of solc that emits Yul/EVM assembly
-- **solx-llvm** (submodule) — Fork of LLVM with an EVM target backend
+- **solx** (this repo) is the Rust workspace: the CLI, the Slang frontend, the dialect binding and codegen.
+- **Slang** is the parser and binder, a git dependency pinned by `rev` in [`Cargo.toml`](./Cargo.toml).
+- [**solx-llvm**](./solx-llvm/) (submodule) is a fork of LLVM with the Sol and Yul MLIR dialects and an EVM target backend.
+- [**solx-solidity**](./solx-solidity/) (submodule) is a fork of solc, kept for its [`test/libsolidity/semanticTests`](./solx-solidity/test/libsolidity/semanticTests/), which the tester runs.
 
-**Compilation pipeline:**
 ```
-Solidity → solc frontend (parse, analyze) → Yul/EVM assembly → LLVM IR → LLVM optimizer → EVM bytecode
-```
-
-Alternative pipeline via Slang (Rust-based frontend, WIP):
-```
-Solidity → Slang (parse, analyze) → MLIR → LLVM IR → EVM bytecode
+Solidity → Slang (parse, bind) → Sol-dialect MLIR → Sol→Yul→Standard passes → LLVM IR → LLVM optimizer → EVM bytecode
 ```
 
-### Workspace Crates (14 total)
+### Workspace Crates
 
 | Crate | Purpose |
 |---|---|
-| `solx` | CLI entry point, frontend dispatch |
-| `solx-core` | Pipeline orchestration (`standard_output_evm`, `standard_json_evm`, `yul_to_evm`, `llvm_ir_to_evm`) |
-| `solx-codegen-evm` | LLVM IR generation, optimization passes, bytecode assembly, linking |
-| `solx-slang` | Alternative Rust-based Slang frontend |
-| `solx-mlir` | MLIR-to-LLVM translation via melior |
-| `solx-yul` | Yul lexer/parser |
-| `solx-evm-assembly` | EVM assembly representation and ethereal IR |
-| `solx-standard-json` | solc-compatible JSON I/O protocol |
-| `solx-utils` | Shared types (contract names, EVM versions, hashing, CBOR metadata) |
-| `solx-dev` | Developer tool: builds LLVM, solc, runs project tests |
-| `solx-tester` | Integration test runner using REVM |
-| `solx-compiler-downloader` | Downloads/verifies compiler binaries |
-| `solx-benchmark-converter` | Benchmark analysis and comparison |
-| `solx-solc-test-adapter` | Adapter for upstream solc tests |
-
-**Important feature split:** `solx-slang` and `solx-mlir` use inkwell with `llvm21-1-no-llvm-linking`, while `solx-codegen-evm` uses `llvm21-1` (links LLVM). They are excluded from `default-members` to avoid Cargo feature unification conflicts. Use `cargo test-slang` to test them.
+| [`solx`](./solx/) | CLI entry point |
+| [`solx-core`](./solx-core/) | Pipeline orchestration |
+| [`solx-slang`](./solx-slang/) | Lowers the Slang AST to Sol-dialect MLIR |
+| [`solx-mlir`](./solx-mlir/) | The Sol and Yul dialect binding over melior: values, places, types, blocks |
+| [`solx-codegen-evm`](./solx-codegen-evm/) | LLVM IR to EVM bytecode: optimization, assembly, linking |
+| [`solx-standard-json`](./solx-standard-json/) | solc-compatible JSON I/O protocol |
+| [`solx-utils`](./solx-utils/) | Shared types: contract names, EVM versions, hashing, CBOR metadata |
+| [`solx-dev`](./solx-dev/) | Developer tool: builds LLVM, runs project tests |
+| [`solx-tester`](./solx-tester/) | Test runner over the REVM corpus |
+| [`solx-compiler-downloader`](./solx-compiler-downloader/) | Downloads and verifies compiler binaries |
+| [`solx-benchmark-converter`](./solx-benchmark-converter/) | Benchmark analysis and comparison |
 
 ## Build Commands
 
-### Prerequisites
-
-Build the `solx-dev` tool first, then use it to build LLVM and solc:
+Build `solx-dev`, then LLVM with MLIR, then solx:
 
 ```bash
 cargo build --release --bin solx-dev
-
-# Build LLVM (outputs to target-llvm/target-final/)
-./target/release/solx-dev llvm build --enable-mlir --enable-utils --build-type RelWithDebInfo
-
-# Build solc libraries (outputs to solx-solidity/build/)
-./target/release/solx-dev solc build
+./target/release/solx-dev llvm build --enable-mlir --enable-utils --build-type RelWithDebInfo   # target-llvm/target-final/
+cargo build-slang              # target/debug/solx
+cargo build-slang --release    # target/release/solx
 ```
-
-### Building solx
-
-```bash
-cargo build              # debug
-cargo build --release    # release (outputs to target/release/solx)
-```
-
-### Environment Variables (auto-set via .cargo/config.toml)
-
-- `LLVM_SYS_211_PREFIX` → `./target-llvm/target-final/`
-- `MLIR_SYS_210_PREFIX` → `./target-llvm/target-final/`
-- `TABLEGEN_210_PREFIX` → `./target-llvm/target-final/`
-- `SOLC_PREFIX` → `./solx-solidity/build/`
-- `BOOST_PREFIX` → `./solx-solidity/boost/lib/`
 
 ## Testing
 
-### Unit and CLI tests
-
 ```bash
-cargo test                              # all tests (unit + CLI)
-cargo test --lib                        # unit tests only
-cargo test --test mod                   # CLI integration tests only (target is `mod`)
-cargo test --test mod -- cli::bin::default  # specific test
+cargo test-slang                                        # unit and CLI tests of solx-slang, solx-mlir and solx
+cargo clippy-slang --all-targets
+cargo build-slang --release && cargo run-tester-slang   # the REVM corpus at -O M3B3 against target/release/solx
+cargo run-tester-slang --path tests/solidity/simple/default.sol   # one test
 ```
 
-### Slang/MLIR tests (cargo alias)
+LIT runs the fixtures under [`solx-mlir/tests/lit/`](./solx-mlir/tests/lit/) against `target/debug/solx`:
 
 ```bash
-cargo test-slang
-# expands to: cargo test -p solx-slang -p solx-mlir -p solx --no-default-features --features slang
+export PATH="$PWD/target-llvm/target-final/bin:$PWD/target/debug:$PATH"
+PYTHONPATH=solx-llvm/llvm/utils/lit python3 target-llvm/target-final/bin/llvm-lit solx-mlir/tests/lit/
 ```
 
-### Integration tests with solx-tester
+## Slang Frontend
 
-```bash
-cargo build --release
-./target/release/solx-tester --solidity-compiler ./target/release/solx
-./target/release/solx-tester --solidity-compiler ./target/release/solx --path tests/solidity/simple/default.sol
-./target/release/solx-tester --solidity-compiler ./target/release/solx --via-ir  # Yul pipeline only
-```
+`solx-slang` lowers the Slang AST to Sol-dialect MLIR through `solx-mlir`. The rules below are the design law of that frontend and the conventions a reviewer would otherwise repeat by hand.
 
-### Project tests (Foundry/Hardhat)
+### Ground truth
 
-```bash
-./target/release/solx-dev test foundry --test-config-path solx-dev/foundry-tests.toml
-./target/release/solx-dev test hardhat --test-config-path solx-dev/hardhat-tests.toml
-```
+1. Ground truth is legacy solc: `solc --asm`, `--bin` and `--storage-layout` define behavior. solx is never evidence about itself.
 
-## Code Conventions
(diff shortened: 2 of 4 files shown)


.claude/skills/review/SKILL.md
```diff
@@ -0,0 +1,30 @@
...
+
+Each reader reads this repository's `CLAUDE.md` in full, then the whole change. It runs nothing that writes, and reproduces a claim only on a scratch copy. It reports each finding with the file and line, what is wrong, the evidence, and the fix. A finding without evidence is not a finding.
+
+## Checking
+
+Every finding is re-read against the code before it goes further. One whose evidence does not hold is dropped. The same finding from two readers is one finding.
+
+## Decisions
+
+A finding with two defensible fixes is a decision. Each decision is put to the user with AskUserQuestion, one at a time, every option naming the evidence for it. The user's answer is final.
```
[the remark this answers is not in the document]
hedgar2017: Applied: a decision is now anything the review cannot settle from the code and CLAUDE.md, including a rule that reads two ways for the change and a change whose intent the diff does not show.

.claude/skills/review/SKILL.md
```diff
@@ -0,0 +1,37 @@
+---
```
ChristopherDedominici: General question:  do you want to have this skill Claude specific, or it should be generalized for all LLMs?

CLAUDE.md
```diff
@@ -1,137 +1,143 @@
...
-- **solx-solidity** (submodule) — Fork of solc that emits Yul/EVM assembly
-- **solx-llvm** (submodule) — Fork of LLVM with an EVM target backend
-
-**Compilation pipeline:**
-```
-Solidity → solc frontend (parse, analyze) → Yul/EVM assembly → LLVM IR → LLVM optimizer → EVM bytecode
-```
+- **solx** (this repo) is the Rust workspace: the CLI, the Slang frontend, the dialect binding and codegen.
+- **Slang** is the parser and binder, a git dependency pinned by `rev` in [`Cargo.toml`](./Cargo.toml).
+- [**solx-llvm**](./solx-llvm/) (submodule) is a fork of LLVM with the Sol and Yul MLIR dialects and an EVM target backend.
```
abinavpp: can we not hardcode the path? my llvm lives outside

CLAUDE.md
```diff
@@ -1,137 +1,143 @@
...
-- `TABLEGEN_210_PREFIX` → `./target-llvm/target-final/`
-- `SOLC_PREFIX` → `./solx-solidity/build/`
-- `BOOST_PREFIX` → `./solx-solidity/boost/lib/`
+`solx-slang` lowers the Slang AST to Sol-dialect MLIR through `solx-mlir`. The rules below are the design law of that frontend and the conventions a reviewer would otherwise repeat by hand.
 
-## Testing
+### Ground truth
 
-### Unit and CLI tests
+1. Ground truth is legacy solc: `solc --asm`, `--bin` and `--storage-layout` define behavior. solx is never evidence about itself.
```
abinavpp: but our asm tests should be inside llvm. so this section should go there. and also, our asm will always differ from solc, we mostly care about the final result

.claude/skills/review/SKILL.md
```diff
@@ -0,0 +1,37 @@
...
+Launch a reader per lens, on Opus unless the invocation names another model. Each reader reads this repository's [`CLAUDE.md`](../../../CLAUDE.md) in full, then the whole change, and asks its question of every clause the change adds or alters.
+
+1. Removal: what fails if this clause is deleted? A clause nothing fails without is the finding.
+2. Trust: what does this clause check, derive or own that a lower layer already guarantees? The guarantee, read in that layer's code, is the evidence.
+3. Foreignness: where does the repository already do this another way? The sibling that does it is the evidence.
+4. Pinning: which claim of the change has no test that fails when the code is wrong? The mutation that leaves the test green is the evidence.
+
+The invocation may add lenses. Each added lens is its own question and its own reader.
+
+A reader runs nothing that writes, and reproduces a claim only on a scratch copy. It reports each finding with the file and line, what is wrong, the evidence, and the fix. A finding without evidence is not a finding.
```
nebasuke: Add the rule a finding rests on. I ran the four lenses as written against #702 and asked each reader to cite one. 14 of 20 findings cited a rule cleanly. 3 could only borrow Fixtures 4 by analogy, because CLI expected-file tests have no rule of their own, which is a gap in CLAUDE.md this column exposed in one run. The 3 with no rule were code hygiene and a solx-standard-json change, outside what this skill reviews, and dropping them is right. So the column does two jobs: it filters out-of-scope findings and it shows where the law is missing a rule.

```suggestion
A reader runs nothing that writes. It reports each finding with the file and line, the rule of CLAUDE.md it rests on by section and number, what is wrong, the evidence, and the fix. A finding without evidence is not a finding. A finding without a rule is either a rule CLAUDE.md is missing or not a finding.
```

nebasuke: Pinning worked read-only. The reader produced nine findings by arguing mutations from the assertions and fixtures, including one that is not hypothetical: `matches_solc` compares parsed values, so it cannot see key order, and the frontend emits `type` first where solc emits it last. So drop the scratch-copy clause and say what Pinning does: argue the mutation from the test, run nothing.

.claude/skills/review/SKILL.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Checking
+
+Every finding is re-read against the code before it goes further. One whose evidence does not hold is dropped. The same finding from several readers is reported once.
```
nebasuke: Two things the run showed Checking has to do. First, dedupe: five findings came from three or more readers, one from all four. Second, check the fix against the repo's lints: two readers said to delete six field docs, and `.cargo/config.toml` enables `missing_docs_in_private_items`, so clippy would put them straight back. A finding whose fix a repo lint rejects does not hold. Either readers read the lint config on line 16, or Checking does.

.claude/skills/review/SKILL.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Decisions
+
+A decision is anything the review cannot settle from the code and `CLAUDE.md`: a finding with more than one defensible fix, a rule that reads two ways for this change, or a change whose intent the diff does not show. Each decision is put to the user with AskUserQuestion, one at a time, every option naming the evidence for it. The user's answer is final.
```
nebasuke: Decisions need the same dedupe as findings, and one more rule. All four readers independently raised "is ABI rendering part of the lowering layer?", which decides whether Scopes and modules 2 applies to abi.rs at all. A decision every reader raises is not a question for the user on each run, it is a gap in CLAUDE.md, and the report should say so once. The other shared decision, assemble in solx or wait for Slang, is real and is what #703 answers.

.claude/skills/review/SKILL.md
```diff
@@ -0,0 +1,37 @@
...
+description: Read-only code review against CLAUDE.md. Ambiguities reach the user as questions, then every finding as a report. Nothing is changed. Use when the code builds and passes its tests.
+---
+
+# Review
+
+The skill changes nothing. It reads, checks, asks, and reports.
+
+## What is reviewed
+
+What the user names when invoking: a diff range, a branch, a path or a file. If nothing is named, the current branch against `main`. If the branch is `main`, the uncommitted changes. If there are none, ask the user with AskUserQuestion what to review.
```
nebasuke: nit: `main...HEAD`, so a branch behind main does not review main's own commits as reverted.

CLAUDE.md
```diff
@@ -1,137 +1,143 @@
...
+
+### Domains
+
+1. A domain lands complete over the types that exist when it lands. Later domains extend the earlier mechanisms for the types they introduce, so every mechanism is designed once as an extension point.
+
+### Fixtures
+
+1. A LIT fixture pins op shape. A tester case under [`tests/solidity/`](./tests/solidity/) pins behavior.
+
+2. A fixture has one RUN line, `solx --emit-mlir=sol %s | FileCheck %s`, and no prose. The CHECKs are the whole statement.
```
nebasuke, line 90: This contradicts the suite: 92 of 122 fixtures under solx-mlir/tests/lit have two RUN lines and use check prefixes, for example control_flow.sol:
```
// RUN: solx --emit-mlir=sol %s | FileCheck --check-prefixes=CHECK,CHECK-SOLX %s
// RUN: solc --mlir-action=print-init %s 2>/dev/null | FileCheck --check-prefixes=CHECK,CHECK-SOLC %s
```
The second line is the solc comparison that Ground truth 1 asks for, so the rule as written outlaws the law's own oracle. Copilot raised this and the force-push kept the text. Describe the invariant, something like: a fixture has a RUN line per oracle, solx always and solc where print-init covers the construct, prefixes only for what one oracle emits and the other does not, and no prose.

CLAUDE.md
```diff
@@ -1,137 +1,143 @@
...
-- **Complex tests** (`tests/solidity/complex/`): directory with `test.json` + source files
-- Test metadata includes: `cases`, `contracts`, `libraries`, `ignore`, `modes` (`Y` for Yul, `E` for EVMLA)
+3. An imported type that clashes with a local one takes the prefix of the crate it came from.
 
-## CI Labels
+4. No contractions in names: `identifier`, not `id`; `message`, not `msg`; `context`, not `ctx`. Initialisms read as words stay short, such as `abi`, `mlir`, `ods`, `url`, `api`, and unit symbols such as `ms`.
 
-- `ci:sanitizer` — enable address sanitizer tests
-- `ci:integration` — enable integration tests
+5. A doc says why the item exists or what is non-obvious, never its name again.
```
nebasuke, line 110: `.cargo/config.toml` turns on `missing_docs_in_private_items` for clippy, and line 45 tells everyone to run `cargo clippy-slang --all-targets`. So every private field needs a doc, and for a field like `name: String` on an error entry there is nothing non-obvious to say. In the #702 dry run two readers told me to delete six such docs, and clippy would demand them straight back. The PR also removes the old line that mentioned `missing_docs`, so a reader has no way to know. Either exempt fields whose name is the whole fact, or drop the private-items lint. I would drop the lint. The public-items `-W missing_docs` in rustflags can stay.

CLAUDE.md
```diff
@@ -1,137 +1,143 @@
...
-### Unit and CLI tests
+1. Ground truth is legacy solc: `solc --asm`, `--bin` and `--storage-layout` define behavior. solx is never evidence about itself.
 
-```bash
-cargo test                              # all tests (unit + CLI)
-cargo test --lib                        # unit tests only
-cargo test --test mod                   # CLI integration tests only (target is `mod`)
-cargo test --test mod -- cli::bin::default  # specific test
-```
+2. The compiler shapes the tests, never the reverse. No emission code exists to keep a fixture passing. A fixture that stops matching a correct change is rewritten.
```
nebasuke: "The compiler shapes the tests, never the reverse" reads as tests following the compiler, which is the opposite of rule 1. The intent is the second sentence: no emission code exists to keep a fixture green. In the dry run a reader cited this rule to flag a fixture shaped around the construct the code gets wrong, that is, it read the rule against its own letter. Drop the first sentence or invert it: "Tests follow solc and the compiler follows the tests; no emission code exists to keep a fixture passing."

CLAUDE.md
```diff
@@ -1,137 +1,143 @@
...
+5. A doc says why the item exists or what is non-obvious, never its name again.
 
-## Renovate Config
+6. A `//` comment carries only a constraint the code cannot express.
 
-When editing `renovate.json`, validate locally before pushing (CI runs the same checks via the `renovate-config-check` job in `test.yaml`):
+### Pull requests
 
-- **Schema check:** `npx --yes --package renovate -- renovate-config-validator renovate.json`
-- **Full extraction dry-run:** `LOG_LEVEL=debug npx --yes renovate --platform=local --dry-run=full` — confirms each dep shows the expected `skipReason` / `updates`. Catches gotchas like `matchPackageNames` failing to match git-source cargo deps (where `packageName` is the git URL, not the `Cargo.toml` key — use `matchDepNames` for those, as the pinned-fork rule does for `inkwell`/`melior`/`slang_solidity`/`web3`).
```
nebasuke, line 137: Three things go with no new home: the inkwell `no-llvm-linking` feature split and why `default-members` excludes the Slang crates (old line 45), the CI labels, and the Renovate validator commands with the `matchDepNames` gotcha for git-source deps. None is derivable from the code and the Renovate one cost a real debugging session. Move them under docs/ or a CONTRIBUTING rather than delete.

CLAUDE.md
```diff
@@ -1,137 +1,143 @@
...
+1. A PR body is short and human-readable: one or two sentences saying what the PR delivers. Nothing else: no headings, lists, tables, file lists or narration.
```
nebasuke: Worth applying to this PR's own body before merge, it has headings and a bullet list at the moment. Nicer if the rule starts without an exception.

CLAUDE.md
```diff
@@ -1,137 +1,143 @@
...
+2. A fixture has one RUN line, `solx --emit-mlir=sol %s | FileCheck %s`, and no prose. The CHECKs are the whole statement.
+
+3. One fixture per construct. A new case joins the fixture that owns its construct.
+
+4. Every CHECK must fail when the input line it matches is deleted.
+
+5. Every fixture source line is observed by some CHECK.
 
-## Code Conventions
+6. A base the fixture does not observe is `abstract`.
```
nebasuke, line 98: Fixtures 1 names two kinds of test, LIT for op shape and the tester for behavior. The CLI tests under solx/tests are a third kind: they compare a compiler output against a file solc produced, and nothing here says what such a file must be. Without a rule the file drifts into being edited by hand, the command that made it is lost, and entries in it go unexercised because the source happens to declare things in the order the code sorts them. One rule covers it:
```suggestion
6. A base the fixture does not observe is `abstract`.

7. An expected file under `solx/tests/data/` is solc's output verbatim, with the solc version and command that produced it recorded beside it, and every entry in it is produced by a construct in the source that no other entry covers.
```
A gap the frontend does not fill yet is then filtered from both sides in the test, never edited out of the file, so the file stays regenerable.

README.md
```diff
@@ -22,10 +21,11 @@
 
 **solx** documentation is powered by [GitHub Pages](https://nomicfoundation.github.io/solx/latest/) and provided as an [mdBook](https://github.com/rust-lang/mdBook), while its Markdown sources can be found in [this directory](./docs/src/).
 To build the book, follow these [instructions](./docs/README.md).
+The design rules of the Slang frontend and the conventions for working on this repository with an agent are in [CLAUDE.md](./CLAUDE.md).
```
nebasuke, line 27: CLAUDE.md now opens with "Human documentation starts at the README and continues under docs/src", and its Architecture section says Slang lowers to MLIR. The architecture page this README links, docs/src/internals/01-architecture.md, still says solx-solidity emits Yul that solx translates to LLVM IR, has a "Why a fork of solc" section, and does not mention Slang or MLIR once. The testing page is the same, zero mentions. So a reader following the links gets two pipelines. Point the Architecture and Testing sections below at the CLAUDE.md sections for now, or land a note at the top of those two pages saying they describe the 0.1.x line.

README.md
```diff
@@ -4,11 +4,10 @@
...
 
-**solx** is a new optimizing compiler for EVM developed by [Matter Labs](https://matter-labs.io/) and [Nomic Foundation](https://nomic.foundation/).
+**solx** is an optimizing Solidity compiler for EVM by [Nomic Foundation](https://nomic.foundation/). [Slang](https://github.com/NomicFoundation/slang) parses and binds the Solidity, the frontend lowers it to MLIR, and an LLVM backend with an EVM target produces the bytecode.
 
 > [!WARNING]  
-> The project is in beta and must be used with caution. Please use it only for testing and experimentation.
-> If you want to use it in production, make sure to test your contracts thoroughly, or [contact us](#contact-us) first.
+> The project is in development and is not ready for production use.
 
 **solx** passes multiple test suites, including:
```
nebasuke: "solx passes multiple test suites" and the five bullets were written for the solc-frontend compiler. The intro above them now describes the Slang frontend, which does not pass these suites yet. Either drop the list until it is true of the compiler the intro describes, or turn it into a status line: which suites the tester runs and how they stand.

README.md
```diff
@@ -73,7 +73,6 @@ For reference, see [llvm-sys](https://crates.io/crates/llvm-sys) and [Local LLVM
 - All other crates are licensed under the terms of either
   - Apache License, Version 2.0 ([LICENSE-APACHE](./solx-standard-json/LICENSE-APACHE) or <http://www.apache.org/licenses/LICENSE-2.0>)
   - MIT license ([LICENSE-MIT](./solx-standard-json/LICENSE-MIT) or <http://opensource.org/licenses/MIT>)
-- [`solx-solidity`](https://github.com/NomicFoundation/solx-solidity/) is licensed under [GNU General Public License v3.0](https://github.com/NomicFoundation/solx-solidity/blob/0.8.34/LICENSE.txt)
```
nebasuke, line 76: The GPL line for solx-solidity goes, but the submodule stays: CLAUDE.md lists it as kept for the semantic tests the tester runs, and line 17 above still links into it. Keep the license line as long as the submodule is in the tree.

nebasuke (changes requested):
Tried the skill on #702. It was a bit expensive with >500K tokens due to all the lenses, but did match up with what I did myself in #703. 

I used the test review to let Claude iterate on the skill.md/claude.md and kept what I thought was possibly relevant.

CLAUDE.md
hedgar2017: Expected, we're going to remove solc after remaining 3 feature PRs are merged so I decided not to add it at all.

CLAUDE.md
hedgar2017: Sounds good, will remove the lint.

CLAUDE.md
hedgar2017: solc is getting removed, so I'd split LIT and semanticTests instead: the former follow, the latter are followed.

README.md
hedgar2017: Expected. Rewriting docs is a huge chunk so I'd like to do it in another PR, probably after removing solc.

README.md
hedgar2017: Good idea.

README.md
hedgar2017: This section originally meant to indicate what is linked into the shipped executable. IIRC, this is how the final license is decided.
Since we're merely importing tests from the repo, I'd remove all mentions of GPL in the repo as well. I'd say it's a part of the full solc removal.

.claude/skills/review/SKILL.md
hedgar2017: Applied verbatim. Report carries the rule as a field now too, since Readers requires it.

hedgar2017: Kept the scratch copy. The pinning reader on #684 ran the mutations on a copy and that's where the strongest evidence in that review came from. It now reads "A reader writes nothing in the repository. Pinning argues the mutation from the test, and may run it on a scratch copy."

.claude/skills/review/SKILL.md
hedgar2017: Applied as a general rule: "One whose fix a repository lint would undo does not hold either." Kept it general so it survives removing the private-items lint.

.claude/skills/review/SKILL.md
hedgar2017: Applied, with a different criterion. "Every reader raised it" is a count, not a reason to change the law. It's now "A design-related decision carries a proposal to add the rule to CLAUDE.md", and decisions are deduped like findings.

.claude/skills/review/SKILL.md
hedgar2017: Applied, but never main: it's `<base>...HEAD`, where base is the branch the PR targets.

CLAUDE.md
hedgar2017: All three are back in CLAUDE.md. Inkwell split under Build Commands, CI labels and the renovate dry run with the matchDepNames gotcha under Pull requests. I also moved the design law above build and test while I was in there.

CLAUDE.md
hedgar2017: Good point. The rule now allows short bullets and a numbered list, and I cut this PR's body to the first sentence plus one line per file.

CLAUDE.md
hedgar2017: Deferring this one. "solc's output verbatim" needs rewording once there's no solc to produce it, so I'd add the rule as part of the solc removal instead.

CLAUDE.md
```diff
@@ -109,12 +118,95 @@ cargo build --release
...
+
+1. A domain lands complete over the types that exist when it lands. Later domains extend the earlier mechanisms for the types they introduce, so every mechanism is designed once as an extension point.
+
+2. A refactor leaves the emitted MLIR of the whole corpus byte-identical. If the output moved, it was not a refactor.
+
+### Fixtures
+
+1. A LIT fixture pins op shape. A tester case under `tests/solidity/` pins behavior.
+
+2. A fixture has one RUN line, `solx --emit-mlir=sol %s | FileCheck %s`, and no prose. The CHECKs are the whole statement.
```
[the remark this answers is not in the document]
hedgar2017: Expected, same as the thread above: solc is going after the remaining 3 feature PRs, so I decided not to add the second RUN line to the rule at all.

CLAUDE.md
```diff
@@ -109,12 +118,95 @@ cargo build --release
...
+
+6. A `//` comment carries only a constraint the code cannot express.
+
+### Pull requests
 
-- Rust edition 2024, toolchain 1.97.1
-- `missing_docs` warning is enabled globally — all public items need doc comments
-- Release builds strip debug symbols (`strip = true`)
-- macOS minimum deployment target: 11.0
+1. A PR body is short and human-readable: one or two sentences saying what the PR delivers, then `` `cargo run-tester-slang`: X -> Y passed. `` Nothing else: no headings, lists, tables, file lists or narration.
```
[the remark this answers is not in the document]
hedgar2017: That rule is gone from CLAUDE.md. A pass count in a PR body goes stale on the next merge.

ChristopherDedominici (approved):
I reviewed only the skill since it's the only file where I have context. Looks good to me.

One point we might want to talk about is making this skill generic for all the Nomic teams. The skill itself is already generic. The "problem" we'll need to fix is having every team define all the rules needed for reviewing files in CLAUDE.md (or providing into it links to references with this info).

CLAUDE.md
```diff
@@ -1,137 +1,156 @@
...
+cargo build-slang --release && cargo run-tester-slang   # the REVM corpus at -O M3B3 against target/release/solx
+cargo run-tester-slang --path tests/solidity/simple/default.sol   # one test
+```
+
+LIT runs the fixtures under [`solx-mlir/tests/lit/`](./solx-mlir/tests/lit/) against `target/debug/solx`:
 
-Tests live in `tests/solidity/`, `tests/yul/`, `tests/llvm-ir/`.
+```bash
+export PATH="$PWD/target-llvm/target-final/bin:$PWD/target/debug:$PATH"
+PYTHONPATH=solx-llvm/llvm/utils/lit python3 target-llvm/target-final/bin/llvm-lit solx-mlir/tests/lit/
```
abinavpp: ditto: can we not hard code paths? we just need to say that these env vars should point to llvm bin dirs

abinavpp (approved):

nebasuke (approved):

PR #712, opened 2026-09-15
docs: writing rules in CLAUDE.md and review skill fixes

hedgar2017:
Gives everything written for this repository one owner in `CLAUDE.md`, so the PR-body rule keeps only what is specific to a pull request and the review skill points at it rather than restating it.


CLAUDE.md
```diff
@@ -1,5 +1,18 @@
...
+Everything written for this repository is short and human-readable.
+
+- No headings, bold labels or numbered scaffolding where a paragraph is enough.
+- No lead that summarizes what follows, and no closing that repeats it.
+- No restatement of what the reader can already see.
+- No account of how the work was done, and no approval of what is already correct.
+- No hedging: a claim that is not verified is not made.
+- No sycophancy, no emoji, and no sentence that only introduces the next one.
+- No tombstones: what was removed leaves no trace.
+- No negative claims: what a thing is not stays unwritten.
```
abinavpp: a lot of this already exist in my "prose" skill. and i think most of us already have similar preference/skill. i'm not sure if we should have it in this repo. but i don't see any problem with having it too.

abinavpp (approved):

CLAUDE.md
hedgar2017: I agree! Feel free to contribute so we can move it to a shared repository soon.
There is also https://github.com/NomicFoundation/singularity that I created long ago, but it has no company-wide traction.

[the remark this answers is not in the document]
hedgar2017: Thanks, adjusted.

PR #717, opened 2026-09-17
fix(slang): report diagnostics with their severity and code

### LINKED ISSUE #714: The Slang frontend reports every diagnostic as an error, including warnings
`solx-slang/src/slang/mod.rs:134-150` maps every entry of `unit.diagnostics()` through `OutputError::new_error_with_data` without consulting `diagnostic.severity()`, and `:162` then bails on `output.has_errors()`. So a Slang warning stops compilation and emits nothing.

The instance that surfaced it is `syntax/missing-version-pragma`, which Slang declares as `DiagnosticSeverity::Warning` and legacy solc also reports as a warning — `Source file does not specify required compiler version!`, compiling to bytecode with exit 0. Under Ground truth 1 solc defines behavior here, so solx should warn and carry on.

- `DiagnosticExtensions` is already imported at `mod.rs:13` for `.message()`; `.severity()` is on the same trait and is never called.
- Every LIT fixture is affected: none of the 158 under `solx-mlir/tests/lit/` declares a pragma, so 141 of them fail against a Slang revision that emits this diagnostic.
- The tester corpus is mostly unaffected — 1236 of 1265 sources already declare one.

hedgar2017:
Slang diagnostics reach the standard JSON output with their own severity and code, so a warning such as a missing version pragma compiles instead of failing. Closes #714.

- `solx-utils` owns `Warning`, its solc codes and `CodeSegment::size_limit()`; `Warning::code()` and `IGNORED_WARNING_CODES` read the same constants.
- Slang is pinned to `9a5a78c3`, where `DiagnosticSeverity::Warning` exists.
- Stacked on #671.


solx-standard-json/src/output/error/mod.rs
```diff
@@ -41,19 +41,22 @@ pub struct Error {
...
-    /// of the LLVM backend output. 6417 (`type(...).runtimeCode` with an assembly constructor)
-    /// applies to solx output as well and may deserve un-ignoring.
-    pub const IGNORED_WARNING_CODES: [&'static str; 3] = ["3860", "5574", "6417"];
+    /// The list of ignored `solc` warnings. The code size warnings are measured on the bytecode
+    /// `solc` itself would have emitted, which solx discards in favor of the LLVM backend output.
+    /// The assembly constructor warning applies to solx output as well and may deserve un-ignoring.
+    pub const IGNORED_WARNING_CODES: [&'static str; 3] = [
+        solx_utils::Warning::CODE_DEPLOY_CODE_SIZE,
+        solx_utils::Warning::CODE_RUNTIME_CODE_SIZE,
+        solx_utils::Warning::CODE_RUNTIME_CODE_ASSEMBLY_CONSTRUCTOR,
```
nebasuke, line 50: Nice!

nebasuke (approved):
Neat!

PR #719, opened 2026-09-18
fix: forward appendCBOR to solc so the reported Yul carries no metadata hash

### LINKED ISSUE #716: Standard JSON Yul output includes metadata despite `appendCBOR: false`
When compiling with solx 0.1.8 using these Standard JSON settings:

```json
"metadata": {
  "appendCBOR": false,
  "bytecodeHash": "none"
}
```

the returned `ir` still contains:

```yul
data ".metadata" hex"a26469706673582212208da001939a79df8e1c061b3236678e4fd3278b1bcca28ab914d6f039b624e4bb64736f6c63430008220033"
```

Solc omits this metadata with the same input. Solx’s final runtime bytecode correctly omits it, so this appears limited to the reported intermediate Yul.

hedgar2017:
The `appendCBOR` setting never reached solc, so the `ir` output reported solc's own metadata object with a real IPFS hash even though solx appended no CBOR to the bytecode. Fixes #716.

`bytecodeHash` stays unserialized: solc rejects `appendCBOR: false` together with any `bytecodeHash` but `none`, and solx accepts it with its `ipfs` default.


solx/tests/cli/no_cbor_metadata.rs
```diff
@@ -101,6 +101,26 @@ fn ipfs_llvm_ir() -> anyhow::Result<()> {
     Ok(())
 }
 
+#[cfg(feature = "solc")]
+#[test]
+fn ir() -> anyhow::Result<()> {
+    let _ = crate::common::setup();
+
+    let args = &[
```
nebasuke, line 109: optional nit: no_cbor_metadata.json already has appendCBOR: false but doesn't select ir, so the reporter's exact input isn't covered end to end. Adding "ir" to its output selection plus the same hex"" predicate in standard_json() would pin it as reported. Fine either way.

nebasuke (approved):

solx/tests/cli/no_cbor_metadata.rs
hedgar2017: Added!

